PR #1, opened 2017-11-11
Initial purl draft spec

pombredanne:
For reference this is the result of a discussion that started here
https://github.com/nexB/scancode-toolkit/issues/805


README.rst
```diff
@@ -0,0 +1,297 @@
...
+
+Known implementations
+~~~~~~~~~~~~~~~~~~~~~
+
+- in Python:
+- in Ruby:
+- in Go:
+- in JavaScript:
+- in Perl:
+- for the JVM:
```
kasper3: Please consider including .NET as well.

For implementation, these places are potential candidate for discussion:

https://github.com/nuget/home
https://github.com/dotnet/home
https://github.com/dotnet/standard

pombredanne: Very good point! I am adding it in the updated version that I am pushing shortly

pombredanne: @kasper3 this has been addressed in the latest push.

jayfk:
How are package managers handled that have no unique package name?

```
# resolves to the same package
django-allauth
django_allauth
Django-Allauth
dJAnGo_aLLauTH
```

pombredanne:
@jayfk your wrote:

> How are package managers handled that have no unique package name? PyPi treats - and _ as the same character and is case insensitive.

Thanks and this is an excellent point. I should know better! We should then specify this for each type. I added this for Pypi in https://github.com/package-url/purl-spec/pull/1/commits/042f1088bb17e0762e68b27daad30637ca44de4f

pombredanne:
@kasper3 the latest version covers NuGet and .NET

pombredanne:
@R2wenD2 @sschuberth @mnonnenmacher @jpopelka @jdaguil @JonoYang @majurg @mjherzog @chinyeungli @tdruez This draft ready for your review which is highly valued!

pombredanne:
@andrew you [wrote](https://github.com/nexB/scancode-toolkit/issues/805#issuecomment-343772803)

> Happy to implement on Libraries.io once the spec is finished 👌 

This is ready for your review now, I guess!

andrew:
@pombredanne it's gonna be a busy week for me, don't block on me!

ashcrow:
It would be good to have an example with the subpath as well. The way I read it this would work:
```
 github:package-url/purl-spec@244fd47e07d1004f0aed9c#/everybody/loves/dogs
```

ashcrow:
I'm happy to create a golang and/or python parser for this spec once it's finalized.

pombredanne:
@ashcrow you wrote:

> It would be good to have an example with the subpath as well. The way I read it this would work:
>> github:package-url/purl-spec@244fd47e07d1004f0aed9c#/everybody/loves/dogs

Yes! thanks. I added it in https://github.com/package-url/purl-spec/pull/1/commits/20f84b31d41a9c8aa082910ce46dc63fa6059f18 with a slight modification: the leading slash is not significant in a subpath. Note also FWIW that subpaths may be common for Go.

pombredanne:
@ashcrow you wrote:

> I'm happy to create a golang and/or python parser for this spec once it's finalized.

Let me make you a co-owner of the org together with @andrew 

I suggest that we use this convention for implementations repo names: `purl-language` as in `purl-python`, `purl-go`, `purl-ruby`, `purl-js`, etc.

pombredanne:
@ashcrow @andrew org owner invite sent. @jayfk do you want it too?

pombredanne:
@ashcrow I have a rough, poorly tested first draft toy python bit of python in scancode here 
https://github.com/nexB/scancode-toolkit/blob/275-streamline-package-manifests-models/src/packagedcode/purl.py

I will move this out to a bona-fide repo as a Python starter, I guess either public domain or MIT licensed with a "Copyright (c) the purl authors". public domain might be best? 

And in terms of contributions a simple DCO https://developercertificate.org/ should be plenty enough IMHO. Do you agree?

NB: if you are not comfy with being an org co-owner I can make you a plain contributor with rights to create repos instead.

ashcrow:
Cool! I'll take a look.

I'm fine with the MIT license and using the DCO.

jasinner:
Great idea, an opensource alternative to ISO 19770-2

pombredanne:
@jasinner scripsit:

> Great idea, an opensource alternative to ISO 19770-2

Interesting indeed, though it is not the intent. IMHO the whole SWID tags ISO spec and tagvault is a terrible ultra proprietary centralized wart to stay clear from ;)

pombredanne:
@jasinner I added a FAQ to the spec on SWID (and CPE while I was at it) 
@sschuberth  I added a FAQ  to the spec on Authority and using the user/password for version based on our discussion in https://github.com/nexB/scancode-toolkit/issues/805#issuecomment-343659705
@jayfk  I refined the type-specific namespace/name normalization for Pypi (lowercase and dash) and GitHub and Bitbucket (lowercase)

pombredanne:
For reference I posted a ticket to Grafeas https://github.com/Grafeas/Grafeas/issues/65 for `purl`

@R2wenD2 The latest push contains explanations why a `purl` should not use `://` after a `type`/scheme .... 

FWIW, I think the thing you call resource URLs here https://github.com/Grafeas/Grafeas/tree/b8ce70a4a88e18575e1d2d91912c156b078d5a8f#resource-urls may be likely invalid URLs per the RFC3986 that defines the URI syntax.

I explained why here: https://github.com/package-url/purl-spec/pull/1/files#diff-88b99bb28683bd5b7e3a204826ead112R169

The `purl` defined in this spec is trying hard to be a valid URL/URI.

ashcrow:
@pombredanne the Python code looks good to me. The only nit I have is the ©️ when pulling it into this repo. It would be ideal if both ©️ and license was consistent across implementations.

We should also name the libraries as close as possible. Most languages should be able to use purl but for the Go implementation we probably shouldn't re-use the package ``purl`` as it's used in [this](https://github.com/ian-kent/purl) project. Granted, different namespaces but could collide nonetheless.  We should have a backup when ``purl`` is not available. Something like ``pkgurl`` or ``packageurl``.

kasper3:
(mustache syntax guy here)
Is there a pure C (99 or 11) or C++ implementation of PURL? Also if someone has scribbled it down in JavaScript, that would be grand! :)

ashcrow:
@kasper3 not yet. The spec is still being refined. I have part of a Go implementation in the works but it wouldn't transfer 100% to JS. However, I expect there will be one 😄

pombredanne:
@ashcrow I pushed something better here as discussed: https://github.com/package-url/purl-python
and with an MIT and a simple `Copyright (c) the purl authors` 
I am not too proud of the code, but it somehow works for a first shot and passes the tests :P

pombredanne:
@ashcrow excellent points in the package names. I guess `purl` is best if available and `packageurl` could come second?

Unrelated, since a `purl` is a valid URL using existing libraries to do the bulk of parsing works nicely.
See https://github.com/package-url/purl-python/issues/1 for an example in Python.

pombredanne:
So since a purl is a valid URI/URL any URL parsing library can be used as a base. ( @sschuberth including Java's built-in parser) ... Therefore I suggest to simplify the parsing and building spec section to eschew most everything about parsing and building to defer instead to the corresponding specs and RFCs.

pombredanne:
While reviewing this IANA list of registered URI schemes https://www.iana.org/assignments/uri-schemes/uri-schemes.xhtml two thoughts came to crooked mind:

1. we should reference it in the spec here https://github.com/package-url/purl-spec/pull/1/files#diff-88b99bb28683bd5b7e3a204826ead112R138

2. to avoid any type/scheme naming conflict (there is already an official scheme for  `go://`) we could consider stating that:
  - we cannot use any official or known schemes unless registered for purl usage
  - OR we could prefix all the type with `purl+type` as in `purl+go:github.com/gorilla/context` .... 

I kinda like the `purl+` type prefix as this makes it always clear we deal with a purl BUT at the same time this makes the string a tad more heavier with more characters.
And we could also register with IANA `purl` and `purl+*` as official schemes

Thoughts?

pombredanne:
BTW, unrelated I am moving the FAQ to the wiki for simplicity.

pombredanne:
FYI I posted an updated purl-python library renaming from `purl` to `packageurl` to avoid any conflict with other Pypi packages, and added what starts to be a decent cross-language test suite as a JSON file at https://github.com/package-url/purl-python/blob/master/test-suite-data.json 

Its is also available on Pypi at https://pypi.python.org/pypi/purl-python/
I guess its purl would then be `pypi:purl-python@0.2.0` ? ;)

ashcrow:
@pombredanne the test suite fixtures may make sense to live in their own repo and pulled in by other repos as a way to have a common set of verification.

README.rst
```diff
@@ -0,0 +1,760 @@
...
+
+
+purl
+~~~~~
+
+`purl` stands for **package URL**.
+
+A `purl` is a URL composed of six components::
+
+    type:namespace/name@version?qualifiers#subpath
```
iamwillbar: Is host or repository ubiquitous enough that it should be part of the URL rather than a qualifier? Some possible examples:

```
npm://registry.npmjs.org/Microsoft/redie@0.1.0
github://github.com/Microsoft/redie
github://myghe.server.example/Foo/bar
```

Without this there's an assumption of what the default registry is for the PURL, that could be a good thing (because it's resilient to change) but the downsides are that:
* The consumer would need to know what the default repository is
* The consumer needs to check the qualifier to determine if the repository has been overridden and fallback if necessary

pombredanne:
@ashcrow you quite rightly wrote:

> the test suite fixtures may make sense to live in their own repo and pulled in by other repos as a way to have a common set of verification.

There you go: https://github.com/package-url/purl-test-suite

README.rst
pombredanne: @iamwillbar I think the default repositories for a `type` are ubiquitous enough IMHO.
The idea is to encapsulate in a type/scheme all the many details and specifics of the general "protocol" of a package: for instance when I say `nuget` or `maven` this at once refers to a public repository, conventions and rules to create and build packages, the API, network protocols and more to talk to the repository, the format of the package manifest and the package archives, the clients that can talk to the repository and query, fetch, install, a dependency resolution mechanism, and so on: all this is quite major.

In contrast I tend to think of the repository host as secondary: there is a main public one, but there can be eventually an infinity of alternative repositories beyond the main one: private, local, mirrors, etc. And assuming that content-wise the package files or artifacts are verified as the same, this is the same package: therefore alternate repositories are minor. Not unimportant but minor and less important than a type, name and version IMHO. And as a lesser data bit, they should therefore come later in the `purl` rather than at the beginning in the `Authority` components of a URL/URI. 

As for the defaults, I tried to list some for the some known package types. We could very make this a data file that would be usable by anyone (actually it likely make sense also as there are some type-specific purl normalization rules, such as for pypi as reported by @jayfk )

pombredanne: Think of it this way: package management CLI do not ask you to specify the default repository at all. They just use it by default unless you specific alternative hosts.

pombredanne:
> I kinda like the purl+ type prefix as this makes it always clear we deal with a purl BUT at the same time this makes the string a tad more heavier with more characters.
> And we could also register with IANA purl and purl+* as official schemes

After some thoughts, prefixing a purl with `purl+` is a neat ide but not needed at all. There is only one registered URL scheme that i'd like to hijack as a `type` and this is for `go:` 
The RFC3367 `go://` scheme is a somewhat arcane, mostly abandoned thing for an obscure CNRP protocol from 15 years ago at https://tools.ietf.org/html/rfc3367 : this was supposed to replace DNS in browser address bars so you could search by "name"; I never saw implemented anywhere: therefore I feel quite good about squatting it unless someone strongly objects to this.

pombredanne:
@sschuberth do you want to be on the GH org too, btw?

README.rst
pombredanne: @iamwillbar does my reply make sense to you?

ashcrow:
Here is the [go implementation](https://github.com/package-url/packageurl-go).

pombredanne:
@jayfk re:

> PyPi treats - and _ as the same character and is case insensitive.

Should this be normalized to `-` or `_`?

jayfk:
I've seen most packages using `-` so I'd go for that one.

README.rst
iamwillbar: It does, but I think my fundamental concern is that you can have the same package identity in two different repositories but be different packages (this is commonly used as an attack vector in fact). By being explicit about the authority you state the canonical trusted location for that package, yes, you might find the package in other places but you can't assume they're the same without some other validation (i.e. comparing purls without the identity is not a safe operation). Does that make sense?

README.rst
```diff
@@ -0,0 +1,760 @@
...
+
+    docker:cassandra@sha256:244fd47e07d1004f0aed9c
+    docker:customer/dockerimage@sha256:244fd47e07d1004f0aed9c?repository_url=gcr.io
+
+    gem:jruby-launcher@1.1.2?platform=java
+    gem:ruby-advisory-db-check@0.12.4
+
+    github:package-url/purl-spec@244fd47e07d1004f0aed9c
+
+    go:google.golang.org/genproto#googleapis/api/annotations
```
iamwillbar: Maybe use gopkg or golang? The term "go" is avoided in a lot of places because it's hard to search for (without excess noise) and this would avoid the conflict you mentioned with the go: prefix already being registered.

README.rst
```diff
@@ -0,0 +1,760 @@
...
+  - `purl` `subpath`: this is a URL `fragment`
+  - In a `purl` there is no support for a URL Authority (e.g. NO
+    `username`, `password`, `host` and `port` components).
+
+- Special URL schemes as defined in https://url.spec.whatwg.org/ such as
+  `file://`, `https://`, `http://` and `ftp://` are NOT valid `purl` types. They
+  may be used to reference URLs in separate attributes outside of a `purl` or in
+  a `purl` qualifier.
+
+- Version control system (VCS) URLs such `git://`, `svn://`, `hg://` or as
```
iamwillbar: I'm not sure why `github:` and `bitbucket:` would be valid but `git:` would not be (the same applies to the others)? github: is just a pointer to a source location on a named, proprietary, source control host, if that is valid I should also be able to have a point to a source location on *any* server that implements the Git (or alternate) protocol.

One argument could be that git://, svn://, hg:// do not refer to specific package type, but that's true of github: and bitbucket: as well, they're just a source location, at that source location could be anything.

README.rst
pombredanne: This makes a lot of sense. So should we use `golang:` or `gopkg:` as a package URLs `type` for Go?

@ashcrow and @jasinner for @victims ...
@R2wenD2 @vbatts and @elad165  for @Grafeas ... which way would you "Go"?

@griesemer @robpike @ken as the designated Golang designers... and @sdboyer as the lead contributor to `go dep` what would be your take and preference? Your input is highly valuable!

For reference, we are specifying here a simple package URLs to conveniently reference a software package in a mostly universal way across tools and techs as a compact URL/URI string. 

The question is: which URL scheme should we use to identity Go "packages"?
e.g. should we use `golang:` or `gopkg:` as a `purl` `type` for Go? Or something else?

And the neater `go://` scheme is already an officially IANA-registered URI/URL scheme (See https://github.com/package-url/purl-spec/pull/1#issuecomment-345274238 for details)

README.rst
pombredanne: You have a good point here again!. Here are some reasons that come to mind(and this ties in not squatting existing URL/URI schemes and the thread on `Go`):

`git://`, `svn://` and `bzr://` and heck... even `cvs:`/`pserver:` to some extent are well known URL schemes (even though they might not be officially registered) so reusing then for a `purl` would lead to a certain level confusion IMHO

While you are right that `github:` and `bitbucket:` are pointers to named and proprietary source control hosts, here are some considerations:

- I do not want to take a stance on open vs. proprietary here even though I am clearly on the open side personally :heart_eyes: 

- they (`github:`, `bitbucket:` and `gitlab:` and likely a few more) are a bit more than than just VCS pointers: they expose an API, naming and other conventions, ticket trackers, release mechanisms, metadata and more that goes well beyond a bare `git` or `hg` repo. This grants them a package-status...

- ... and even more so, as they are even natively supported by several package managers as a place to "provision" code deps (such as in Go or PHP Composer)

- they are kinda prominent as a place to host package source code. 

- bare VCS URLs have their own purpose but there is no set naming conventions on what would be a package name. I may for instance infer that git://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git is the URL to Linus's Linux git tree but not much more than this. I'd be at loss to make the same guess on `git://git.kernel.org/pub/scm/linux/kernel/git/pjt/test.git`

So based on all this, as much as I love `git://`-like URLs -- so much so that I even helped spec these for [SPDX](https://github.com/spdx/spdx-spec/blob/cfa1b9d08903befdf03e669da6472707b7b60cb9/chapters/3-package-information.md#37-package-download-location-) -- I think they might be best left out as valid `purl` `types` and tracked either as a separate data item outside of a `purl` OR  use them as one of the `qualifiers` as [proposed](https://github.com/package-url/purl-spec/pull/1/files#diff-88b99bb28683bd5b7e3a204826ead112R683) unless you strongly object to this.

Could there be use cases that would be made impossible by using this approach?

pombredanne:
@iamwillbar do you want an invite as a co-org admin BTW?

README.rst
iamwillbar: The scenario I was thinking originally is that in NPM's package.json you can depend on a package on any arbitrary git server, not just GitHub, and I'm not sure how I would represent that in a purl based on the current spec. The three pieces of information that would need to be encoded in the purl are that it is NPM, that it's accessible over the git protocol, and the location of it.

It's almost as if to support this scenario (and maybe it's a specialized case), you would need to do something like:
`npm+git:host.com/path/to/repo.git`

pombredanne: The same applies to Python pip requirements and Gemfiles AFAIK.
I think this would come out this way:

`npm:mypackage@12.4.5?vcs_url=git://host.com/path/to/repo.git`

pombredanne: or 
`npm:mypackage@12.4.5?vcs_url=git://host.com/path/to/repo.git@4345abcd34343`

pombredanne:
@andrew I think we should adopt a CoC. You are the master there. What should this be?

README.rst
pombredanne: Actually we could define a sub mini-spec for a `vurl` aka. `version control URL` which could be similar to the SPDX location spec, but abstract the strict requirement to set the transport because it does not matter too much whether you use https, git or or ssh transports from a repo reference standpoint: this is still the same repo to me....
May be something like `vurl+git://host.com/path/to/repo.git@commitish#subpath/in/repo` ? this is still a strongly hierarchical URL (e.g. host  comes before version/commitish, and a subpath may exist only in a certain commitish or branch )
This could be somewhat redundant with SPDX and possibly https://maven.apache.org/scm/scm-url-format.html too.... though may be simpler and more pragmatic?

Just a thought :) I am not convinced this is needed

joepie91: You'd definitely want to have some way of representing a generic Git or other VCS repository. There are still a lot of people who host their own VCS repositories (me included!).

I don't think it's viable to try and add every single VCS host on earth; there's no central catalog of these, and adding them as-they-are-encountered would create an ongoing workload and a lot of delays (as it will take time for individual hosts to be added), and disincentivize use of this specification for people who can't afford to wait for those host additions.

Simply having a `?vcs` qualifier doesn't quite cut it, either - sometimes the VCS repository is the *only* place in which a given piece of software is published, and there are no packages for it published elsewhere.

pombredanne: @joepie91 thank you  for chiming in, you wrote:

> Simply having a ?vcs qualifier doesn't quite cut it, either - sometimes the VCS repository is the only place in which a given piece of software is published, and there are no packages for it published elsewhere.

But even if these cases, these are still packages, right? e.g. this would still be an NPM with a proper package.sjon and a name and version, correct?

README.rst
pombredanne: @iamwillbar it makes perfect sense. 

You wrote:
> It does, but I think my fundamental concern is that you can have the same package identity in two different repositories but be different packages (this is commonly used as an attack vector in fact).

That's interesting.... would you have some pointers to examples of this? I would be quite interested to read more about it.

> By being explicit about the authority you state the canonical trusted location for that package, yes, you might find the package in other places but you can't assume they're the same without some other validation (i.e. comparing purls without the identity is not a safe operation). Does that make sense?

IMHO a `purl` should be dealing with with package files integrity and content verification: this is a complex topic that is `type`-specific and depends on many things already implemented in most if not all package managers with such thing as checksum or signature verification (including repositories public keys as is common for RPMs and Debian), ensuring the transport is TLS or SSH secured, etc. This is best managed outside of a `purl`as it is today. Yet there is provision to use `qualifiers` for this too. 

In the vast majority of the cases, comparing two`purl` strings should be fine. And if you need more, then again I think is best managed as an attribute outside of a purl ....

For instance, take Pypi: there can be several downloads available for a given name/version for various package archive formats and eventually native builds for diifferent os/arch/ as there https://pypi.python.org/pypi/lxml/

To get accurate down to the package archive level, you would want to get the download URL which looks this way:

https://pypi.python.org/packages/34/95/b3d5d1227561887eb691b7c6b0bc9aa7eaa752b13262df2d2f04c8633242/lxml-4.1.1-cp33-cp33m-macosx_10_6_x86_64.macosx_10_9_intel.macosx_10_9_x86_64.macosx_10_10_intel.macosx_10_10_x86_64.whl#md5=7a0fd79e4899eea1d32fe251e10105f3

And beside being a long and ugly name (here mostly to deal with some macOS ABI peculiarities) it contains two verifications: a trailing, weak md5 in the fragment and a string blake2b leading 512 bits checksum in the path

e.g. This and checksums are otherwise best managed outside IMHO

ashcrow:
I think ``golang`` is the best type name for ``Go``. It matches the website and how people refer to ``Go`` when avoiding misunderstanding.

README.rst
joepie91: @pombredanne No, it would not necessarily be. The only moment where a package name in package.json is checked for uniqueness is when you try to actually *publish* it - if I wanted to, I could create a package named `express` and simply never publish it to NPM, and that would work fine when used locally.

In that scenario, the primary (and only unique) identifier for that 'package' would be its Git repository - the package.json *does* have a `name` in it, but that name actually belongs to a *different* package from a registry perspective (and basing an identifier on that would therefore result in the identifier describing the wrong package).

This applies even moreso for eg. languages that don't *have* a package registry or even a package manager at all, where a VCS repository may well be the only place to obtain the package from.

pombredanne: You are making good points but the intent of this spec is to cater first and foremost to the common case, e.g. a well known published package: this should be clean, simple and concise. The simple case should be simple. It should be possible to cater to the complicated, goof and mistakes or unpublished things. That may not be simple then, just possible.

The goal is that when @andrew libraries.io and @jayfk pyup and @ashcrow and @jasinner victims and @Grafeas and my https://github.com/nexB/scancode-toolkit and more can all talk about the same package when we talk about about the same simple `purl` 

So in the scenario you list above, you have a goofy messed up case. A `purl` cannot make this any less goofy. In this case it might be something such as `npm:express@12.2?vcs_url=yourproprietaryrepo.git@commit` which may be incorrect name- and version-wise but that is still what is this package is known as when you install it with `npm` or `yarn`. 
The same would apply to an un-named private composer package. Weird and goofy, legit for composer but not something we should model a spec after IMHO and certainly not the common case. And in the vast majority of the cases, these weird and goofy things may not even need a purl at all.

> This applies even moreso for eg. languages that don't have a package registry or even a package manager at all, where a VCS repository may well be the only place to obtain the package from.

That would be the case of Golang with `go get` but even there is a name alright and an explicit origin in the namespace. 

Is there any public package repository that provides no name for its packages and only a version control system instead?    None that I can think of.

And even if this is so we can always find a warty way to accommodate warts and not make the wart the norm for the common case.

pombredanne:
ok, so I am updating the draft with `golang` instead of `go` as a type. I also chatted with Debian devs and internally dpkg always makes the package names lowercase (not the version, just the name) and will a note for this

README.rst
pombredanne: @joepie91 reading the package.json [spec](https://docs.npmjs.com/files/package.json#name):

> The most important things in your package.json are the name and version fields. Those are actually required, and your package won't install without them. The name and version together form an identifier that is assumed to be completely unique. Changes to the package should come along with changes to the version.

So yes, one could create goofy things, but this would break the central assumption of the `npm` "protocol" that `The most important things in your package.json`.

Again, fine by me, but I do neither have to care for this, nor cater to this, would you agree?

README.rst
guillemj: I've to agree that for at least .debs and .rpms, I think the assumption that there's some kind of flat namespace does not hold. The tuple name_version_arch is only ever unique (possibly) within a specific archive. Which is what stuff like apt would expect, but some vendors might not even honor.

For .deb packages, there's also the concept of the target distribution and the one built in. So the Debian vim source package targets the Debian unstable distribution, and in Debian it's built there. But when Ubuntu takes that source package (which targeted Debian unstable, as can be seen from its debian/changelog) it gets rebuilt on the current Ubuntu distribution at the time, so the resulting binaries are different.

I think there's a distinction between many of the other types, because they include a canonical archive. While deb and rpm are just package formats.

For example in the dpkg world we've got the concept of a vendor (https://manpages.debian.org/dpkg-vendor), and actively encourage derivatives, forks and alternative implementations to declare their own vendor so that it can be referred to (https://wiki.debian.org/Derivatives/CensusFull). My plan is still to start embedding that vendor information into each .deb built so that we can distinguish them, and uniquely identify these different tuples (https://www.hadrons.org/~guillem/debian/docs/origin.proposal).

Also, even using something like "jessie" is not unique enough, and it's a moving attribute, as packages move from distribution to distribution (say uploaded to unstable, and migrated to testing, which points to stretch). And  while declaring the complete repository URL is certainly very cumbersome, and I'd say even an implementation detail that should not be leaking, I'd recommend instead including the vendor in the purl. So I think I'd qualify it by the vendor as part of the archive location, and then add possibly an attribute for the distribution currently in, which would be certainly optional. So the binary dpkg packages from Debian currently in stretch and Ubuntu could look like this?

  deb:debian/dpkg@1.19.0.4?arch=amd64&distro=stretch
  deb:ubuntu/dpkg@1.19.0.4?arch=amd64

pombredanne: @guillemj Thank you ++ to chime in. As the lead maintainer for `dpkg`,  this is quite an honor! I owe you so much and I am so grateful for your code. 
Now, this makes 100% sense and I just pushed an update with this and you examples.

pombredanne: @guillemj Actually, would both of these be built from the same exact source code? e.g.  would Ubuntu ever rebuild a package changing the sources and not change the version? because if this the exact same source, then may be the vendor could be a `qualifier` instead since then the "provenance" of the package would resolve to the exact same copyright and license, same bugs, same vulnerabilities, same everything source wise and may not be much different from an `arch`? 
(and hopefully when https://reproducible-builds.org is everywhere this may even be the same binaries?) 

So, may be this instead?

 
    deb:dpkg@1.19.0.4?arch=amd64&distro=stretch&vendor=debian
    deb:dpkg@1.19.0.4?arch=amd64&&vendor=ubuntu

README.rst
```diff
@@ -0,0 +1,781 @@
...
+How to parse a `purl` string in its components
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Parsing a `purl` ASCII string into its components works from right to left,
+from `subpath` to `type`.
+
+Note: some extra type-specific normalizations are required.
+See the "Known types section" for details.
+
+To parse a `purl` string in its components:
```
pombredanne: Since a `purl` is a valid URL, I will simplify this vastly to reference the URL syntax RFCs and only focus on the extra specifics instead. This will make it simpler and one should be able to build a `purl` parser using a standard existing URL/URI parser alright with some minimal additions. Let me play with this in Python first.

README.rst
guillemj: re:Ubuntu, right, I'd expect them not to reuse the name_version tuple for the source package for content that is distinct, and I'm not sure how much mechanism they have that would prevent that, even if due to accidents or omissions. But the reverse could happen. Someone packages something first in Ubuntu, and then someone else in Debian packages the same and does not notice, and reuses the same tuple. This actually applies to any Debian derivative, for which there might be even less tight relationships. So, I'd not even assume the uniqueness for source packages, even though between Debian and Ubuntu that should be mostly ok in principle. Also nothing says that the upstream tarball has not been modified when preparing the source package, to say, strip non-free stuff, which obviously depends on what every derivative considers non-free. :)

re:Reproducible-builds. That still cannot give guarantees beyond the same archive. Reproducible builds are defined as the same output for the same specified input, not for any random input. So, that's why we record a subset of the build environment. As long as the packages are not exactly the same, because for example Ubuntu has not synced all of them from Debian unstable, or even the builds happened on different days (so had different build-dependency versions installed) then the resulting binaries will most probably not be the same.

I guess in the end whether the vendor might need to be a fundamental part of the purl or just a qualifier, depends on the purpose of the purl. If it's to refer to an abstract reference of a package with certain attributes, then that might be fine. If it's to reference a specific instance of that package, that needs to be installed on a system, then probably the vendor is fundamental. I think there's widespread agreement within Debian that mixing and matching packages from different repos is unwise, more so when the different repos contain rebuilds of the same packages.

README.rst
jonasob: I fear you're moving the goalpost here. Your original premise is that a "`purl` is a URL string used to identify and locate a software package in a mostly universal and uniform way". This makes sense, for that which are clearly software packages: npm, gem, rpm, and so on. But by including `github`, you make the implicit claim that a repository on Github can be seen as a software package.

And if you make this claim, then it's my belief that you need to extend that to cover any Git vcs. From what can be seen as defining a "software package", I see no discerning difference between one hosted on Github and one hosted elsewhere.

Else, this all looks very good to me :-)

ashcrow:
```
deb:debian/dpkg@1.19.0.4?arch=amd64&distro=stretch
deb:ubuntu/dpkg@1.19.0.4?arch=amd64
```

This got me thinking a bit more around the ``rpm`` case as well. The examples currently show:

```
     rpm:curl@7.50.3-1.fc25?arch=i386&distro=fedora-25
     rpm:curl@7.50.3-1.fc25?arch=src
```

I think ``rpm`` should follow a similar pattern to deb. Something like:

```
     rpm:fedora/curl@7.50.3-1.fc25?arch=i386&distro=fedora-25
     rpm:opensuse/curl@7.56.1-1.1.?arch=i386&distro=opensuse-tumbleweed
```

We may want to have the release as part of the namespace as well, though that may be a bit too far 😄: 

```
     rpm:fedora/25/curl@7.50.3-1.fc25?arch=i386
     rpm:opensuse/tumbleweed/curl@7.56.1-1.1.?arch=i386
```

README.rst
pombredanne: @jonasob Thank you ++ for your feedback. Much honored!

You wrote:

> And if you make this claim, then it's my belief that you need to extend that to cover any Git vcs. From what can be seen as defining a "software package", I see no discerning difference between one hosted on Github and one hosted elsewhere.

I am not making an implicit claim about a repo on Github. Instead this is more about acknowledging that the practice is mostly to treat these as such in @andrew and @benjam [Libraries.io](https://github.com/librariesio/libraries.io/tree/5213d3b03354aed98bbb36c2db92475d093f6e76/app/models/concerns), in @reiz [versioneye](https://github.com/versioneye/crawl_r/tree/master/lib/versioneye/crawlers), in @jpopelka and @fridex [fabric8](https://github.com/fabric8-analytics/fabric8-analytics-worker/blob/aea81992d419e2c984c3962cbd775590d52a5a08/f8a_worker/utils.py#L34) and a few more, including my own code.  

This is just recognizing that a Kalithea, GitHub, Savanah, Sourceforge, Gitlab, Bitbucket, etc "repo" is more than just a VCS repo but comes with a package-like set of data e.g. a namespace and naming conventions, licensing & provenance data, releasing, APIs, tickets, docs, etc. similar to the data accumulated in package manifests. 

In contrast a bare VCS repo does not have much of these. It is however an important (I'd say essential) locator in many cases. So here is how I would handle a bare Git repo:

`generic:<somename>?vcs_url=git://myhost/mypath/myrepo.git@commitish` 

<somename> could be at worst the name of the repo.
(adding if needed a subpath to that vc_url) using essentially the same URL convention that I contributed to SPDX. We could even relax this URL syntax a bit and define it clearly in this spec (for instance the fact we use the git, ssh, or https protocol is not of much interest compared to the fact this is a git or hg repo.). 

See also this related [discussion](https://github.com/package-url/purl-spec/pull/1#discussion_r151830997)
 
> Else, this all looks very good to me :-)

Thanks!

README.rst
pombredanne: @guillemj Thank you for the details. .... I am still weighing the pros and cons of using the Debian vendor as a namespace vs a qualifier key/value. Right now it is using the `namespace` based on your feedback, but I feel we would be better served by a qualifier: this still keeps distinct `purl`s for each vendor but also makes two `deb` packages with the same name and version sort close by which I kinda like as in the vast majority of the cases they would be distinct but eventually minor variants of the same code. In all cases, Debian would always sort before Ubuntu because of the alphabet (and the historical precedence as it should ;) )

pombredanne:
@ashcrow Thanks! ... and this nails it: we will use the vendors as namespace  as @guillemj proposed for `deb:` and the same for `rpm:` as you proposed. This nails it nicely :)
 Let me update the spec

pombredanne:
@ashcrow do you know if an RPM name is case sensitive BTW?
And as we are adding more `type`-specific rules, I wonder if we could come with a simple JSON file that would list in a language neutral way the normalization to apply to each `purl` component.... but that may be over doing it for now.

README.rst
pombredanne: @jonasob So for instance, I would consider [gitea](https://git.fsfe.org/jonas)  that you use at the FSFE as a `type` as it has all the same attributes as a Bitbucket and similar.

ashcrow:
> @ashcrow do you know if an RPM name is case sensitive BTW?

``rpm`` is case sensitive. For the most part it seems most packages are lowercase but there are some (EX: ``python3-PyYAML``) which mix case. Naming obviously is specific to distro based on their naming conventions ... for example ``PyYAML`` could be:

- ``PyYAML``
- ``pyyaml``
- ``python-yaml``
- `` python-pyyaml``
- ``python3-PyYAML``
- etc...

README.rst
pombredanne: @guillemj thx. We are using vendor as namespace for deb and rpm now, thanks to your feedback!

pombredanne:
@joepie91 @jonasob @iamwillbar based on our discussions above, I will add a section to clarify VCS URLs (whih are not `purl`s but quite valuable. As suggested to avoid any confusion with existing schemes and in particular the widely used `git://` scheme and will be based on [pip](https://pip.pypa.io/en/latest/reference/pip_install/#vcs-support) , [SPDX Downlaod location](https://github.com/spdx/spdx-spec/blob/cfa1b9d08903befdf03e669da6472707b7b60cb9/chapters/3-package-information.md#37-package-download-location-) , [package.json](https://docs.npmjs.com/files/package.json#repository) , [npm install] (https://docs.npmjs.com/cli/install) 

This will be something like: 
```
vcs+[git|hg|bzr|cvs|svn,p4|etc]://authority/path@<version/ or commitish'>#<subpath inside repo>
```
And usable either outside of a `purl` or as a `vcs_url` qualifier.

Would this work?

pombredanne:
**Version ranges.**...

There is a use case to specify a range of versions as is commonly done for dependencies. Another domain would be when you have a bug or vulnerability when it can be handy to say: this bug or security vulnerability exists in this range of versions and is fixed starting with this version, etc.

It could be tempting to hack the `version` component for this. Yet that would make thing very quickly confusing. Furthermore, each package `type` has its own weird way to define these version constraints in a mostly similar but byzantinely different way.

So what I could do is add an optional `qualifiers` to define this with these caveats: the version requirements string will be opaque and completely `type` specific, using the syntax defined for each `type`. As for all `qualifiers` it will be optional and in most case should NOT be in the `purl` but stored as an external attribute.

Alternatively we could specify a simple common syntax for such version range and requirements, but it feels to me as a rabbit :rabbit2: hole I'd better no dig in. :rabbit:

README.rst
```diff
@@ -0,0 +1,793 @@
...
+=======
+
+When tools, APIs and databases process or store multiple package types, it is
+difficult to reference the same software package across tools in a uniform way.
+
+For example, these tools, specifications and API use relatively similar
+approaches to identify and locate software packages, each with subtle
+differences in syntax, naming and conventions:
+
+- Grafeas uses a scheme, namespace, name and version in a URL-like string
```
R2wenD2: Probably helpful to add links to anything relevant here so folks can easily look at references?

README.rst
```diff
@@ -0,0 +1,793 @@
...
+~~~~~~~~~~~~~~~~~~~~
+
+::
+
+    bitbucket:birkenfeld/pygments-main@244fd47e07d1014f0aed9c
+
+    deb:debian/curl@7.50.3-1?arch=i386&distro=jessie
+
+    docker:cassandra@sha256:244fd47e07d1004f0aed9c
+    docker:customer/dockerimage@sha256:244fd47e07d1004f0aed9c?repository_url=gcr.io
```
R2wenD2: When I last reviewed, I think this was docker:gcr.io/customer/dockerimage@sha256:hash - what lead to the change here?
Why not include the repo in the namesapce?

README.rst
```diff
@@ -0,0 +1,793 @@
...
+  - Since a `purl` never contains a URL Authority, its `type` must not be
+    suffixed with double slash as in 'docker://' and should use instead
+    'docker:'. Otherwise this would be an invalid URI per rfc3986 at
+    https://tools.ietf.org/html/rfc3986#section-3.3::
+
+        If a URI does not contain an authority component, then the path
+        cannot begin with two slash characters ("//").
+
+    While it is acceptable to use such '://' suffix, its is not significant and
+    not needed for unambiguous parsing even if it make a `purl` look likea a
```
R2wenD2: s/make/makes

R2wenD2: s/likea/like

README.rst
```diff
@@ -0,0 +1,793 @@
...
+  - Each `namespace` segment must be a percent-encoded string
+  - When percent-decoded, a segment:
+
+    - must not contain a '/'
+    - must not be empty
+
+  - A URL host or Authority must NOT be used as a `namespace`. Use instead a
+    `repository_url` qualifier. Note however that for some types, the
+    `namespace` may look like a host.
+
```
R2wenD2: Again, I'd like to understand why a namespace cannot use URL host?

README.rst
```diff
@@ -0,0 +1,793 @@
...
+- `swift` for Swift packages:
+- `vim` for Vim scripts packages:
+- `wordpress` for Wordpress packages:
+- `yocto` for Yocto recipe packages
+
+
+Known `qualifiers` key/value pairs
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Note: Do not abuse of `qualifiers`: it can be tempting to use many qualifier
```
R2wenD2: s/of//

README.rst
```diff
@@ -0,0 +1,793 @@
...
+
+Additional, separate external attributes stored outside of a `purl` are the
+preferred mechanism to convey extra long and optional information such as a
+download URL, vcs URL or checksums in an API, database or web form.
+
+
+With this warning, the known `key` and `value` defined here are valid for use in
+all package types:
+
+- `repository_url` is an extra URL for an alternative, non-default package
```
R2wenD2: "non-default" leaves an open question about what default actually is.

iamwillbar:
@pombredanne my issue with this syntax is it doesn't tell you what to expect at that path:

```
vcs+[git|hg|bzr|cvs|svn,p4|etc]://authority/path@<version/ or commitish'>#<subpath inside repo>
```

The specific scenario I was hoping to address was an NPM that is in source control and not published to an NPM registry. This is a relatively common scenario. That purl tells me that there's something in a source control directory but doesn't tell me what the something is.

Something like this would be more useful for the scenarios I'm thinking about (npm could be any of the purl package types that support version control addressing):
```
npm+[git|hg|bzr|cvs|svn,p4|etc]://authority/path@<version/ or commitish'>#<subpath inside repo>
```

iamwillbar:
RE: version ranges

One option is to state that purl is only used to point at a single package version and not support version ranges at all. It depends on the scenarios that purl is looking to address. For our scenarios we're nearly always wanting to address packages post-resolution so we're dealing with single versions.

iamwillbar:
@pombredanne I just saw your comment about being co-owner. Yes, I'd be happy to help co-own this effort. Also, not sure if you saw the issue I opened #3, this thread is getting a little unwieldy.

README.rst
pombredanne: Good point. I have added these!

README.rst
pombredanne: @R2wenD2 I wanted to avoid having for being uniform, but to you point, in hindsight, the docker command would be `docker pull gcr.io/customer/dockerimage` for instance e.g. a non-default registry would be best in the namespace. Let me revert this back as in anycase the namespace would be optional

README.rst
pombredanne: It makes sense for docker, so that would be type specific. But in the general case, we are avoiding  the reuse of the authority part of a URL. I updated things to support docker namespace for this

README.rst
pombredanne: Fixed

pombredanne: Fixed

README.rst
pombredanne: Fixed

README.rst
pombredanne: I added an explanation. This default is listed for a given type. such as packagist.org for `composer`

pombredanne:
@iamwillbar I sent you an owner invite. Thank you!

you wrote:

> RE: version ranges
> One option is to state that purl is only used to point at a single package version and not support version ranges at all. It depends on the scenarios that purl is looking to address. For our scenarios we're nearly always wanting to address packages post-resolution so we're dealing with single versions.
> If we did tackle ranges I'd suggest having a separate "semver range" spec and then have purl adopt the "semver range" spec.

All excellent points that are closing the case: I totally agree that this is best left out for now.

pombredanne:
@iamwillbar re: vcs URLs the inetnet would not be to support something like:

`npm+git://somereg.com/path@abcded4545#some/path`

Instead I would rather use the `vcs_url` as an optional qualifier that would take a well specified URL, the principle being to avoid adding extra schemes and reusing a URL host/authority in the `purl` and just pushing this less common case to the qualifiers.

e.g. get something like this:


`npm:myname@1.2.3?vcs_url=vcs+git://somereg.com/path@abcded4545#some/path`

pombredanne:
Following @iamwillbar suggestion in #3 , I will be merging this, in master tomorrow and starting a new PR to have a cleaner slate for a final round of comment. 
This is starting to looks rather good to me... 

**Any volunteer for a .Net, JVM and JavaScript implementation?**

Alternatively I could try to see how the Python one could be bundled as a NuGet or a Maven Jar.... and I could fire a transpiler to get a JS version as @all3fox worked out a clean path to transpile Python to JS this summer in his GSoC project here: https://github.com/nexB/license-expression/tree/transpile

sdboyer:
I have many, many thoughts on all this (primarily good 🎉😁), but I've been very busy these past few days so haven't had a chance to put them down. so, two quick confirmations:

1. `golang` is fine.
2. ranges absolutely should not be included; those are constraints, not versions. related (all versions can be treated as constraints; not all constraints are versions) but fundamentally different, and my understanding of purl's goals suggests that versions are what's wanted, here.

pombredanne:
@sdboyer Thank you ++ and this is an honour to have `Mr godep` chiming in: your feedback is spot on. So:
 
1. `golang` is the type for Go language `type`
2. no version ranges should be in a `purl` `version` alright as this would be confusing and semantically incorrect

pombredanne:
BTW, re: implementations @andrew may come up with a Ruby one in the future, but he and @BenJam are likely super busy with holiday community services at https://github.com/24pullrequests/24pullrequests/

pombredanne:
@indirect I wonder if you could chime in for Rubygems? In particular on this https://github.com/package-url/purl-spec/pull/1/files#diff-88b99bb28683bd5b7e3a204826ead112R514 ?

pombredanne:
@zkat any input on NPMs by chance? In particular on this https://github.com/package-url/purl-spec/pull/1/files#diff-88b99bb28683bd5b7e3a204826ead112R584 and the threads related to NPMs available through Git only as brought by @iamwillbar

pombredanne:
@segiddins would you care to check what would the type-specific elements for cocoapods? this would mean a lot!

pombredanne:
@tgamblin this `purl` Package URL thing might be of some tangential interest to you and @spack

segiddins:
@pombredanne I don't know what this repo is, are you sure you meant to tag me?

pombredanne:
@segiddins yes I meant to ask you about your feedback on this package URL spec and if it makes sense to you when it comes to the planned treatment of Rubygems (and/or Cocoapods) such as here: https://github.com/package-url/purl-spec/pull/1/files#diff-88b99bb28683bd5b7e3a204826ead112R514

pombredanne:
But for now, let me merge this so we can use a simpler review process of tickets!

pombredanne:
So next step, please review the main doc and submits tickets as needed, this will be less messy than a crowded PR. Thanks you all for chiming in!

README.rst
```diff
@@ -0,0 +1,806 @@
...
+    npm:foobar@12.3.1
+
+    nuget:EnterpriseLibrary.Common@6.0.1304
+
+    pypi:django@1.11.1
+
+    rpm:fedora/curl@7.50.3-1.fc25?arch=i386&distro=fedora-25
+    rpm:opensuse/curl@7.56.1-1.1.?arch=i386&distro=opensuse-tumbleweed
+
+(NB: some checksums are truncated for brevity)
```
tgamblin: I see a potential problem here in that this is going to result in a lot of new URL schemes, but URL schemes require a lot of review to get approved.  [See here](https://www.iana.org/assignments/uri-schemes/uri-schemes.xhtml).  I suspect they're going to reject this spec on that basis alone.  The obvious way around it would be to change these to look something like:

```
purl:rpm/fedora/curl@7.50.3-1.fc25?arch=i386&distro=fedora-25
purl:github/package-url/purl-spec@244fd47e07d1004f0aed9c
...
etc.
```

Although honestly if this is going to be a thing, and it's always written in a URL, the `url` part of `purl` seems obvious.  Why not put the focus on the package?  e.g.:

```
pkg:github/package-url/purl-spec@244fd47e07d1004f0aed9c
...
etc.
```

You'll note that `pkg:` isn't yet claimed 😄 .

README.rst
```diff
@@ -0,0 +1,806 @@
...
+  - The `subpath` must be interpreted as relative to the root of the package
+
+
+Character encoding
+~~~~~~~~~~~~~~~~~~
+
+For clarity and simplicity a `purl` is always an ASCII string. To ensure that
+there is no ambiguity when parsing a `purl`, separator characters and non-ASCII
+characters must be UTF-encoded and then percent-encoded as defined at::
+
```
tgamblin: I'm not sure this is the way to go.  URLs can have UTF-8 characters, and they've benefitted countries with different character sets, e.g. Japan and China.  I think you should allow UTF-8 to be inclusive.  What if there are packages whose users don't ever write their names in roman characters?  ASCII seems very limiting there.  [See here](https://www.w3.org/International/articles/idn-and-iri/)

tgamblin:
@pombredanne: looks cool!  Two questions and kind of one for @andrew:

1. The original referenced issue says the goal is to have a "unique" identifier for each package, although the spec doesn't seem to dwell on that too much, which is probably a good idea.  Do you have ideas on how to reconcile the same package fetched from multiple sources?  e.g., the same Python package might exist in `pypi`, `conda`, `spack`, and system package managers.  @andrew: does `libraries.io` do anything to reconcile the different names?
2. Is the idea that these URLs will one day be fetchable by `curl`?  How do you see the translation being implemented?  I guess `libraries.io` could provide that as a service?

README.rst
```diff
@@ -0,0 +1,806 @@
...
+Each package manager, platform, type or ecosystem has its own conventions and
+protocols to identify, locate and provision software packages.
+
+
+Problem
+=======
+
+When tools, APIs and databases process or store multiple package types, it is
+difficult to reference the same software package across tools in a uniform way.
```
iarna: Ok… so is the goal to provide a URI that:

1) That tells you which tool consumes this
AND
2) Provides enough information for that tool to consume it

?

Or is the goal only to provide a way to uniquely reference a specific package, not to be able to reconstruct how its installed?

README.rst
iarna: I saw `npm://registry.npmjs.org/Microsoft/redie@0.1.0` and immediately went: well, that's a bad idea.

As has been suggested elsewhere in the thread, eliding the registry when referring to the global registry solves the proxy problem. (One other thing, package owner should never be in an npm purl.)

So, something like:  `npm:redie@0.1.0`

Or on a private registry: `npm:redie@0.1.0?registry=private.corp.com`

The namespace thing conveniently is compatible w/ npm scopes. They do canonically start with an `@` which would be important to keep in mind when writing parsers. Thatd be like: `npm:@iarna/cli@1.0.0`

README.rst
```diff
@@ -0,0 +1,806 @@
...
+    type:namespace/name@version?qualifiers#subpath
+
+Components are separated by a specific character for unambiguous parsing.
+
+The defintion for each components is:
+
+- **type**: the package "type" or package "protocol" such as maven, npm, nuget,
+  gem, pypi, etc. Required.
+- **namespace**: some name prefix such as a Maven groupid, a Docker image owner,
+  a GitHub user or organization. Optional and type-specific.
```
iarna: Just gonna add it here too: the npm version of this starts with `@`. It would be super nice to not have to escape that. 😆

README.rst
```diff
@@ -0,0 +1,806 @@
...
+`username`, `password`, `host` and `port` components. A `namespace` segment may
+sometimes look like a `host` but its interpretation is specific to a `type`.
+
+
+Some `purl` examples
+~~~~~~~~~~~~~~~~~~~~
+
+::
+
+    bitbucket:birkenfeld/pygments-main@244fd47e07d1014f0aed9c
```
iarna: `bitbucket` isn't really a consumer though, is it? Like, what kind of package does that specifier refer to?

This is sticky because npm currently consumes specifiers that look almost like this. Specifically, for npm that would be: `bitbucket:birkenfeld/pygments-main#244fd47e07d1014f0aed9c`. 

The full npm specifier, assuming it was an npm package, would be: `pygments-main@bitbucket:birkenfeld/pygments-main#244fd47e07d1014f0aed9c`

I'm wondering how that would encode as a purl, particularly seeing as `#` is already used for a subpath.

README.rst
iarna: What does `npm:mypackage@12.4.5?vcs_url=git://host.com/path/to/repo.git` refer to exactly? Because npm specifiers can be a version OR a git URL, and git URLs can have versions, eg, the following are valid and each refer to a different thing:
```
mypackage@12.4.5
mypackage@git://host.com/path/to/repo.git
mypackage@git://host.com/path/to/repo.git#semver:12.4.5
```
Further, if you have `mypackage@12.4.5` then you can get the vcs URL from the package metadata.

npm supports `bitbucket:`, `github:`, `gitlab:` and `gist:` prefixes. (Implementation is [here](https://github.com/npm/hosted-git-info#readme))

README.rst
pombredanne: @tgamblin I moved this URL and scheme discussion to this ticket: https://github.com/package-url/purl-spec/issues/4

README.rst
pombredanne: I am with you there: the whole point is to avoid any weird thing and ensure that things are properly UTF-8 and then percent encoded. This means that eventually any character may be used, but using non-ASCII would require a bit more encoding and decoding.

pombredanne:
@tgamblin  you wrote:

> looks cool!  Two questions and kind of one for @andrew:
Thanks!


> 1. The original referenced issue says the goal is to have a "unique" identifier for each package, although the spec doesn't seem to dwell on that too much, which is probably good.  Do you have ideas on how to reconcile the same package fetched from multiple sources?  e.g., the same Python package might exist in `pypi`, `conda`, `spack`, and system package managers.  @andrew: does `libraries.io` do anything to reconcile the different names?

I kinda like to think of these as "mostly" unique, at least unique if a package manager/type provides some unicity within its standard package manager and within a repo/registry of these. Most provide such a guarantee. 

As for thing being the same, I would think this is something that a DB of `purl`s can help with. There is a an amazing graph of relations among the packages: one upstream package may be repackaged in Linux distro, has its source on GH and BB, be bundled or packaged on Conda, spack as RubyGems, etc.

For me, I intend to maintain such relationships in https://github.com/nexB/vulnerablecode (e.g. relate a CPE and several `purl`s together and relate this cluster to a vulnerability; and I capture some relationships in https://github.com/nexB/scancode-toolkit/blob/275-streamline-package-manifests-models/src/packagedcode/models.py#L237 (e.g. this srpm is the sources or this rpm)

Finding that two packages are the same is not trivial matter though. 
I know of two efforts in that domain, focused on Linux mostly:
- oswatershed by @tannewt which is now dormant https://web.archive.org/web/20140531044841/http://oswatershed.org/ ... the most up to date fork is at: https://github.com/pombredanne/open-source-watershed 
- @repology https://repology.org/ by @AMDmi3 which is reasonably new and actively developed.

In all cases, this is hard and @AMDmi3 does a rather superb job in this domain with his concept of "meta package"





> 2. Is the idea that these URLs will one day be fetchable by `curl`?  How do you see the translation being implemented?  I guess `libraries.io` could provide that as a service?

I would not make this part of the goals .... though there is a discussion in https://github.com/package-url/purl-spec/issues/5 started by @jackfirth where I toyed with such an hypothetical tool that could fetch a `purl` https://github.com/package-url/purl-spec/issues/5#issuecomment-347132121

README.rst
pombredanne: @iarna bitbucket, github, gitlab are not strictly-speaking providing packages but are de-factor large reservoirs of packagish-things. They provide more than just a a VCS repo and have tickets, release, some API-fetchable metadata, etc which makes these "packagish" enough to join the fray IMHO.
In the case you mentioned you are encoding the whole VCS address as an NPM version and this works beautifully IMHO:
The purl would be:
`npm:pygments-main@bitbucket:birkenfeld/pygments-main%23244fd47e07d1014f0aed9c`

Of note is the percent encoding of the version `#` to avoid parsing an incorrect `subpath` that would otherwise come out ass this ugly mess:
type='npm', name='pygments-main', version='bitbucket:birkenfeld/pygments-main',  subpath='244fd47e07d1014f0aed9c'

README.rst
pombredanne: @iarna I am with you there, but FWIW there are some percent-encoding alright in NPMs registry API URLs:
For instance https://registry.yarnpkg.com/@invisionag%2feslint-config-ivx/ works but https://registry.yarnpkg.com/@invisionag/feslint-config-ivx/  comes out as `{"error":"Not found"}` just so you known.... the slash between the "namespace" and name must be percent encoded....and yes, I know you might think that I have researched too many of these tiny details and quirks. 

Here in a `purl` I guess this is not strictly needed: a name is always required. Therefore a leading `@` in a scoped NPM `namespace` is never ambiguous. What would be ambiguous is if we allow unescaped `@` anywhere in the name or namespace and a `purl` comes with no version.

For instance, say a package name is `super@package` for the hypothetical `weirdo` package type:

With the `purl` `weirdo:super@package` I would parse `@``package` as a version even though no version was provided: `weirdo:super@package@1.2.3` would not be ambiguous though of course, but my hypothetical weirdo packages rarely have a version attached...

So for simplicity I specified that the whole name and each namespace segment should be percent-encoded: we could relax this for the leading `@` in namespace/name (to make scoped NPMs look beautiful) alright. I will update this and both can be acceptable in any case.

README.rst
pombredanne: @iarna you wrote:

> Ok… so is the goal to provide a URI that:

>    That tells you which tool consumes this
>    AND
>    Provides enough information for that tool to consume it ?

Not exactly `which tools consumes it`: for instance a `maven` `purl` does not tell you to use Maven, or Gradle, SBT or Ivy,  an `npm` `purl` does not tell you you should use the `npm` tool over `yarn` or anything else. `pypi` does not tell you to use `pip` , `easy_install` or `buildout` .... what `npm` mor `pypi` mean here is the whole spet of packaging format, manifest, registry, protocols and APIs that a tool uses to deal with such package `type`.

And provide just enough to consume it: well that's a possibility though it may be just a side goal.

In #5 I toyed with the hypothetical idea of a "meta" package manager.... Though frankly this would be a just a franken manager IMHO.


> Or is the goal only to provide a way to uniquely reference a specific package, not to be able to reconstruct how its installed?

> Who would consume these? Humans? Software? Which software?

I think that the distinguished [racketist](https://racket-lang.org/) @jackfirth provides a better summary in #6 that I could have ever written. I intend to steal his words and add them to the spec:

> [...] here are the proposed use cases as I understand them:

> - Cross-system metadata indexing to search and monitor packages by metadata like available versions, dependencies, contributors, etc. across multiple package managers (libraries.io)

> - Vulnerability tracking to determine whether a package's set of possible transitive dependencies includes a known vulnerability and whether the version constraints of that dependency graph allow or prevent patching

> - Other kinds of package-content-agnostic analysis tools, especially tools that look at the dependency graphs of package ecosystems


So IMHO the primary beneficiaries are folks or DBs or APIs or tools that deal with several package formats and languages. Very selfishly,that might include me, @andrew @jayfk @jasinner @ashcrow @R2wenD2 @tgamblin @jpopelka and hopefully many more. This may be used in these context for UI and DB and APIs. 

It may also --if this gets enough traction-- be something that influences positively some standardization in the domain in the future as a side benefit: say tomorrow you decide to include an `npm` package size as a key identifying package attribute: may you will think about whether this makes sense purl-wise? and either reconsider this or contribute to `purl` to adopt such a fine new standard?

pombredanne:
So this PR is closed but the discussion can go on there alright. Tickets are best going forward though!

In particular an important one would be #9 : `Should the purl scheme/type be prefixed with `purl+`?`
Please chime in on this as this would be a reasonably important change.

sschuberth:
> @sschuberth do you want to be on the GH org too, btw?

Only if it gets a nice icon 😆

pombredanne:
@sschuberth then you won the right to design a logo! Invitation sent!

But what's wrong with this fine logo? :D
https://avatars2.githubusercontent.com/u/33497028?s=60&v=4

sschuberth:
@pombredanne Thanks!

> But what's wrong with this fine logo?

Nothing... if you like Space Invaders :wink:

PR #18, opened 2017-11-29
Fixed Maven examples

brianf:
I noticed that sources is actually a classifier and added another complex example.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -578,7 +578,8 @@ candidate list further down.
   - Examples::
 
         maven:org.apache.xmlgraphics/batik-anim@1.9.1
-        maven:org.apache.xmlgraphics/batik-anim@1.9.1?packaging=sources
+        maven:org.apache.xmlgraphics/batik-anim@1.9.1?classifier=sources
+        maven:org.apache.xmlgraphics/batik-anim@1.9.1?package=zip&classifier=dist
 
 
 - `npm` for Node NPM packages:
--- a/README.rst
+++ b/README.rst
@@ -575,13 +575,18 @@ candidate list further down.
 
   - The default repository is `repo1.maven.org`
   - The group id is the `namespace` and the artifact id is the `name`
-  - Known `qualifiers` keys are: `classifier` and `packaging` as defined in the
-    POM documentation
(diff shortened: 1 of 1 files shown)


README.rst
```diff
@@ -578,7 +580,8 @@ candidate list further down.
   - Examples::
 
         maven:org.apache.xmlgraphics/batik-anim@1.9.1
-        maven:org.apache.xmlgraphics/batik-anim@1.9.1?packaging=sources
+        maven:org.apache.xmlgraphics/batik-anim@1.9.1?classifier=sources
+        maven:org.apache.xmlgraphics/batik-anim@1.9.1?package=zip&classifier=dist
```
pombredanne: Thanks!
Is this really `package` and not `packaging` there? I used that based on this: https://maven.apache.org/pom.html but you should know better than me!

Also do you mind to add a [DCO](https://developercertificate.org/) signed-off-by in the PR text body at least and in the future in commit messages?

ashcrow:
@pombredanne If you're still keen on it I think now would be a good time to move the test-suite file into the spec. That way when folks update the spec they can update the test files without doing extra work.

README.rst
brianf: Let me fix that. Technically I think it should be extension here since packaging is more a term used in the pom which brings in a lifecycle and plugins. The resulting files may not map directly to a package. Example: package maven-plugin still generates a jar. If you're trying to programmatically fetch from the repo and you only have the package, you end up having to embed knowledge of various mappings. Here we are referring to the resulting coordinates to construct a file, and extension is much more obvious.

pombredanne:
@ashcrow test suite all moved to this repo now.
@brianf if you care having a look at it, that would be very nice!

See https://github.com/package-url/purl-spec/blob/c1be96d8c880d278651d4bb0ee90f392ad2e2043/test-suite-data.json

brianf:
Reconciled the spec in a separate pull

brianf:
Gah. I'm a little rusty on the specifics here. It would be more maven proper to use type not extension. Fixing...

pombredanne:
LGTM. Thank you ++ . merging!

PR #22, opened 2017-11-30
fixed up maven cases to match examples in the spec

brianf:


pombredanne:
Thanks again! merging.

Issue #9, opened 2017-11-27
Should the purl scheme/type be prefixed with `purl+`?


pombredanne:
From https://github.com/package-url/purl-spec/pull/1#issuecomment-344537386 :

> While reviewing this IANA list of registered URI schemes https://www.iana.org/assignments/uri-schemes/uri-schemes.xhtml two thoughts came to my crooked mind:

> 1. we should reference it in the spec here https://github.com/package-url/purl-spec/pull/1/files#diff-88b99bb28683bd5b7e3a204826ead112R138

> 2. to avoid any type/scheme naming conflict (there is already an official scheme for  `go://`) we could consider stating that:
>  - we cannot use any official or known schemes unless registered for purl usage
>  - OR we could prefix all the type with `purl+type` as in `purl+go:github.com/gorilla/context` .... 

> I kinda like the `purl+` type prefix as this makes it always clear we deal with a purl BUT at the same time this makes the string a tad more heavier with more characters.
> And we could also register with IANA `purl` and `purl+*` as official schemes

> Thoughts?

ashcrow:
Doing ``purl+`` does make sense. It is a little ugly IMHO but it also does specify strongly what it is and lets us use names that we should probably not use because they are valid in different forms elsewhere.

pombredanne:
@ashcrow agreed. And it may defuse some arguments from the URL/URI experts too

pombredanne:
@R2wenD2 what do you think about this `purl+` URL scheme/type prefix?

jackfirth:
This has some problems would be very difficult to get registered. But in the interest of productive discussion, I'll suggest an alternative: why not just define a single `purl` scheme and specify that the first component of the authority must be some registered "package namespace"? PURLs would then look like this:

- `purl://pypi/django@1.11.1`
- `purl//docker/ubuntu@jessie`

This doesn't solve the problem of specifying how each of those namespaces work, but it would let you divide up the spec like this:

- The main PURL spec, which registers the `purl` scheme, defines structure and semantics common to all PURLs including what stuff is optional and required in a PURL, specifies how PURLs will maintain forward compatibility, defines what "package namespaces" are, and defines a package namespace IANA registry.
- A spec for each package manager that defines its package namespace, registering it in the package namespace registry defined in the PURL spec.

[...]

PR #31, opened 2018-02-09
Use a "pkg:" scheme throughout for purl

pombredanne:
* use a "pkg:" scheme throughout for purl #9
 * also update spec license to MIT per #21 
 * update test suite accordingly

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -460,11 +460,18 @@ To parse a `purl` string in its components:
   - The left side lowercased is the `type`
   - The right side is the `remainder`
 
+- Split the `remainder` once from right on '@'
+
+  - The left side is the `remainder`
+  - Percent-decode the right side. This is the `version`.
+  - UTF-8-decode the `version` if needed in your programming language
+  - This is the `version`
+
 - Split the `remainder` once from right on '/'
 
   - The left side is the `remainder`
-  - Percent-decode the right side
-  - UTF-8-decode the `name` if needed in your programming language
+  - Percent-decode the right side. This is the `name`
+  - UTF-8-decode this `name` if needed in your programming language
   - Apply type-specific normalization to the `name` if needed
   - This is the `name`
 
--- a/README.rst
+++ b/README.rst
@@ -199,11 +199,9 @@ The rules for each component are:
         If a URI does not contain an authority component, then the path
         cannot begin with two slash characters ("//").
 
-    While it is acceptable to use such '://' scheme suffix, its is not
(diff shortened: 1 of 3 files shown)


ashcrow:
I'm good with this. I'd like to see our spec be versioned though so we can denote in libraries what version(s) are supported.

pombredanne:
@ashcrow yes, we need it versioned and this would be towards a first version

pombredanne:
Your feedback is welcomed!
@R2wenD2 @sschuberth ping :)
@jeffmcaffer ping too !

pombredanne:
@goneall @kestewart @tsteenbe @zvr ping too. Your feedback is welcomed since this is considered for use in SPDX too!

README.rst
```diff
@@ -166,43 +180,49 @@ A `purl` is a URL
...
 
         If a URI does not contain an authority component, then the path
         cannot begin with two slash characters ("//").
 
-    While it is acceptable to use such '://' suffix, its is not significant and
-    not needed for unambiguous parsing even if it makes a `purl` look like a
-    familiar web URL. In its canonical form, a `purl` must NOT use such '://'
-    `type` suffix. 
-  - `purl` parsers must accept URLs with such '://' and must ignore the '//'.
+    While it is acceptable to use such '://' scheme suffix, its is not
```
goneall: In what situations would it be acceptable to use '://' - seems to contradict other statements where it is not allowed

pombredanne: Good point and this could be cleaned. The point there was to say that we can make parser cope with undue `://` but using them would make the purl an invalid URL. So to you point It is best to move this part only to the parsing procedure description.
Thanks!

sschuberth:
@pombredanne, could you please also add a rationale to the commit message *why* we would want to use a "pkg:" scheme for purl? Like, please document the benefit.

pombredanne:
@sschuberth good point. I will do so shortly!

pombredanne:
@sschuberth sorry for getting back so late to you on this but I think that all the discussions  and rationale comment on why using a "pkg:" scheme are found in the ticket #9 
Do you think some would  still need to be carried over here?

sschuberth:
Well, I prefer to have those things in Git, and not (only) in GitHub, for portability and consistency. But that's probably only my personal preference.

pombredanne:
@sschuberth what I will do instead is compile a doc that will explain the rationale and move all that in the repo as git tracked file.
In the meantime, I am merging this alright

PR #37, opened 2018-04-06
Adjust maven test-suite-data

jdillon:
Fixes some inconsistency introduced by: https://github.com/package-url/purl-spec/commit/d141f9cf14e5b268a5baf06ee54fbaff9ae9657e

It appears that the above change added some inconsistency in `purl` and `canonical_purl` wrt to changes to the qualifier names.  And in the case of the "maven pom reference" test-data, had an erroneous `packaging=sources` qualifier that doesn't make any sense (probably copy-pasta?)

### DIFF
--- a/test-suite-data.json
+++ b/test-suite-data.json
@@ -110,7 +110,7 @@
   {
     "description": "maven uses qualifiers",
     "purl": "Maven:org.apache.xmlgraphics/batik-anim@1.9.1?repositorY_url=repo.spring.io/release&classifier=sources",
-    "canonical_purl": "maven:org.apache.xmlgraphics/batik-anim@1.9.1?packaging=sources&repository_url=repo.spring.io/release",
+    "canonical_purl": "maven:org.apache.xmlgraphics/batik-anim@1.9.1?repository_url=repo.spring.io/release&classifier=sources",
     "type": "maven",
     "namespace": "org.apache.xmlgraphics",
     "name": "batik-anim",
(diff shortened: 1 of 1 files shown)


brianf:
+1, looks like I had a few cut paste issues in the original.

pombredanne:
@jdillon Thanks! let me merge this on top of the PR #31  this week.

jdillon:
looks like this needs to be updated wrt to the pkg url scheme changes, but still needs to be fixed...

sschuberth:
@jdillon, mind rebasing this onto current master to resolve the conflicts?

NAHDEEZAI:
compile 'com.google.dagger:dagger:2.17'

stevespringett:
@jdillon do you mind rebasing so this can be merged in?

jdillon:
looks like there is only one major problem with "maven pom reference" test-data, "extension" should be "type".  the other entry was removed or refined and fixed in the process.

pombredanne:
@jdillon thank you!. 
you wrote:
> looks like there is only one major problem with "maven pom reference" test-data, "extension" should be "type". the other entry was removed or refined and fixed in the process.

FWIW the extension vs. type is something that always puzzled me to be honest especially this doc is a tad of an head scratcher http://maven.apache.org/ref/3.6.0/maven-core/artifact-handlers.html

With that said @brianf you suggested and contributed the use of extension rather than type .
So what should it be?

stevespringett:
@jdillon I've looked at your changes and agree with the modifications. They reflect correct Maven terminology. In fact, the CycloneDX Maven plugin, which automatically generates BOMs with PackageURL as one of the fields, uses `type` instead of `extension`, which is in-line with the changes you've made.

jdillon:
@stevespringett did you still want a pr just to change that one test?  I think you understand the change and have been working on it already; so this can probably be closed?

PR #46, opened 2018-08-16
Added space in version & qualifier + invalid qualifier key name

stevespringett:
In this example, the purl and canonicalized version should be the same but it a good test of the implementations use of encoding and decoding on this field.

Signed off by: Steve Springett


sschuberth:
I've integrated this manually as GitHub for some reason did not accept the merge conflict resolution I had done in its web editor.

PR #49, opened 2018-08-30
Add `packageurl-rs` to the known implementations

althonos:
Hi there ! 

I've developed an [implementation for Rust](https://github.com/althonos/packageurl-rs) (actually, more like a parser/serializer since it does not do validation right now) so I thought it was the right place to put it although not complete.


sschuberth:
Thanks!

Issue #39, opened 2018-04-06
Percent encoding spec and : and /


jdillon:
The notes in the specification about percent encoding of ":" are a bit confusing:

```
the '#', '?', '@' and ':' characters must NOT be encoded when used as separators. They may need to be encoded elsewhere
the ':' type separator does not need to and must NOT be encoded. It is unambiguous unencoded everywhere
```

It seems like these 2 contradict either other.  The former indicates that ":" may need to be encoded elsewhere.  The later indicates that ":" is "unambiguous unencoded everywhere".

Similarly the qualifier component documentation says:

```
A value must be must be a percent-encoded string
```
... and does not mention anything about "/".  But the test-suite-data.json references canonical_purl representations like `repository_url=repo.spring.io/release`.

And the percent-encoding docs state:
```
the '/' used as namespace/name and subpath segments separator does not need to and must NOT be percent-encoded. It is unambiguous unencoded everywhere
```

My interpretation of this boils down to... "/" is never encoded, but the language in the parts of the specification are unclear.  The name, namespace and subpath parts are clear wrt to "/", which leaves the qualifier and version bits as vague as to if "/" is supposed to be percent-encoded or not.

stevespringett:
With reference to:

`
pkg:docker/customer/dockerimage@sha256:244fd47e07d1004f0aed9c?repository_url=gcr.io
`

It is my opinion that the testsuite in the .NET, Python, and Java implementations as well as in the spec repo are incorrect and violate the spec. Whereas the testsuite in the Rust implementation adheres to the spec (as written).

Spec states:
* The version is prefixed by a '@' separator when not empty
* A version must be a percent-encoded string
* the '#', '?', '@' and ':' characters must NOT be encoded when used as separators. They may need to be encoded elsewhere

Therefore, the canonicalized purl should be changed to:

`
pkg:docker/customer/dockerimage@sha256%3A244fd47e07d1004f0aed9c?repository_url=gcr.io
`

We need clarification on this.

stevespringett:
With reference to:

```
pkg:Maven/org.apache.xmlgraphics/batik-anim@1.9.1?repositorY_url=repo.spring.io/release&classifier=sources
```

It is my opinion that the testsuite in the .NET, Python, and Java implementations as well as in the spec repo are incorrect and violate the spec. Whereas the testsuite in the Rust implementation adheres to the spec (as written).

Spec states:
* the '=' qualifiers key/value separator must NOT be encoded
[...]

PR #52, opened 2019-01-12
Update Test Suite

jeremylong:
1. Correct versions that included a purl with an un-encoded colon in the version (test-suite-data.json:88).
    
    Per [purl-spec#character-encoding](https://github.com/package-url/purl-spec#character-encoding)
    
    > the '#', '?', '@' and ':' characters must NOT be encoded when used as separators. They may need to be encoded elsewhere
2. Corrected copy paste issue in description of check for /// (test-suite-data.json:255).
3. Added test for un-necessary trailing slash (test-suite-data.json:267).
4. Added a bintray example to test namespaces with multiple segments (test-suite-data.json:279).
5. Added test case to check for invalid '..' segments in the subpath (test-suite-data.json:327).
5. Added test case to check for invalid '.' segments in the subpath (test-suite-data.json:339).

### DIFF
--- a/test-suite-data.json
+++ b/test-suite-data.json
@@ -85,8 +85,8 @@
   },
   {
     "description": "docker uses qualifiers and hash image id as versions",
-    "purl": "pkg:docker/customer/dockerimage@sha256:244fd47e07d1004f0aed9c?repository_url=gcr.io",
-    "canonical_purl": "pkg:docker/customer/dockerimage@sha256:244fd47e07d1004f0aed9c?repository_url=gcr.io",
+    "purl": "pkg:docker/customer/dockerimage@sha256%3A244fd47e07d1004f0aed9c?repository_url=gcr.io",
+    "canonical_purl": "pkg:docker/customer/dockerimage@sha256%3A244fd47e07d1004f0aed9c?repository_url=gcr.io",
     "type": "docker",
     "namespace": "customer",
     "name": "dockerimage",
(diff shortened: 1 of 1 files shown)


test-suite-data.json
```diff
@@ -85,8 +85,8 @@
   },
   {
     "description": "docker uses qualifiers and hash image id as versions",
-    "purl": "pkg:docker/customer/dockerimage@sha256:244fd47e07d1004f0aed9c?repository_url=gcr.io",
-    "canonical_purl": "pkg:docker/customer/dockerimage@sha256:244fd47e07d1004f0aed9c?repository_url=gcr.io",
+    "purl": "pkg:docker/customer/dockerimage@sha256%3A244fd47e07d1004f0aed9c?repository_url=gcr.io",
+    "canonical_purl": "pkg:docker/customer/dockerimage@sha256%3A244fd47e07d1004f0aed9c?repository_url=gcr.io",
```
stevespringett: This hasn't been clarified yet and is a subject of discussion in #39 which needs to be clarified before merge.

test-suite-data.json
```diff
@@ -252,7 +252,7 @@
     "is_invalid": false
   },
   {
-    "description": "slash /// after type  is not significant",
+    "description": "slash /// after scheme is not significant",
```
stevespringett: Thanks for the correction. This is a carryover from a time before PackageURLs began with pkg:/

test-suite-data.json
```diff
@@ -263,6 +263,30 @@
     "subpath": null,
     "is_invalid": false
   },
+  {
+    "description": "slash / between version and qualifiers separator is not significant",
+    "purl": "pkg:maven/org.apache.commons/io@2.6/?scope=test",
```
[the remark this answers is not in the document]
jeremylong: This was added as part of code coverage in the java implementation. If this is invalid - then the [How to parse](https://github.com/package-url/purl-spec#how-to-parse-a-purl-string-in-its-components) may contain unnecessary steps. The step after `Split the remainder once from left on ':'`  is `Strip the remainder from leading and trailing '/'`. Trying to figure out what other type of URL would end up with a trailing '/' at this point in parsing.

JimFuller-RedHat:
this is an old PR -  everything seems addressed except for the maven/java fun  - can this PR be closed ?

pombredanne:
@JimFuller-RedHat re:

> this is an old PR - everything seems addressed except for the maven/java fun - can this PR be closed ?

Likely so, but there is still some encoding issue that @johnmhoran is tracking in https://github.com/package-url/purl-spec/issues/344 and I created at least https://github.com/package-url/purl-spec/pull/361

The other topics addressed in the PR would need to be carefully reviewed and adopted or pushed back. @jeremylong IMHO there are too many important but separate concerns in the PR, and I think we should address both the test suite and the spec at once, each with small PRs.

johnmhoran:
Thanks for your comment @JimFuller-RedHat.  IMHO this is not ready to be closed -- it raises a version issue, a subpath issue, and an issue re that odd '/' just before the '?' separator.  I am trying to address the various topics raised here -- see, e.g., the worksheet I'm drafting to track all encoding issues at once, grouped by PURL component so we can make a better informed decision.  See https://docs.google.com/spreadsheets/d/1biOCUY4eCqQaYmfGDHVrASV9igYEzct6

johnmhoran:
That would be great @jeremylong -- thank you.  👍

jeremylong:
Closing this PR as it has been broken down into separate PRs: #364, #365, #366, #367, and #368.

johnmhoran:
My belated thanks @jeremylong !  👍

PR #62, opened 2019-06-06
Add extra package types

grv87:
Terraform: https://registry.terraform.io/
Vagrant: https://app.vagrantup.com/boxes/search
Chocolatey: https://chocolatey.org
Helm: https://helm.sh
Gradle Plugins: https://plugins.gradle.org
Ivy: https://ant.apache.org/ivy/

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -690,6 +690,7 @@ Other candidate types to define:
 - `cargo` for Rust packages:
 - `carthage` for Cocoapods Cocoa packages:
 - `chef` for Chef packages:
+- `chocolatey` for Chocolatey packages
 - `clojars` for Clojure packages:
 - `cocoapods` for Cocoapods iOS packages:
 - `conan` for Conan C/C++ packages:
@@ -705,9 +706,11 @@ Other candidate types to define:
 - `eclipse` for Eclipse projects packages:
 - `gitea` for Gitea-based packages:
 - `gitlab` for Gitlab-based packages:
+- `gradle` for Gradle plugins
 - `guix` for Guix packages:
 - `hackage` for Haskell packages:
 - `haxe` for Haxe packages:
+- `helm` for Kubernetes packages
 - `hex` for Erlang and Elixir packages
 - `julia` for Julia packages:
 - `lua` for LuaRocks packages:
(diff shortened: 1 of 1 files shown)


stevespringett:
Is there such a thing as Ivy packages? I didn't think so. I though Ivy resolved GAVs and used Maven conventions and repos. If that's the case, then having Ivy as a type is not valid, as the package itself is maven. Clarifications? as my experience with Ivy is short-lived and a really long time ago.

grv87:
Yes, Ivy is able to read from Maven repositories.

But it also has its own separate packaging system, with different layout conventions and different metatada.
See e.g. comparison of layouts in [Artifactory documentation](https://www.jfrog.com/confluence/display/RTF/Repository+Layouts#RepositoryLayouts-ArtifactPathPatterns).

Maybe you never heard of this because it is primary intended for enterprise use. See the last paragraph [here](https://ant.apache.org/ivy/m2comparison.html).

jdillon:
Ivy isn't a package system, its just a dependency management/resolve system.  And its use is highly specific to the configuration of the resolver.

I am also not sure it makes sense to have a purl for it, as w/o the resolve configuration context, as well as the format type context its kinda useless to consume.

You could easily take a maven purl and use ivy to resolve the artifact, and you likely could configure other similar-to-maven-ish layouts to resolve non-maven artifacts.

grv87:
If by *package system* we mean just the format of archives, then Maven is not package system too. Package system is jar/war/ear.

If we consider metadata and repository layouts too, then, as I pointed out, genuine Ivy has different layout and metadata from Maven.
For non-believers I've set up example Ivy repo in my Artifactory: [link](https://fidata.jfrog.io/fidata/webapp/#/artifacts/browse/tree/General/ivy-test/org.fidata.gradle/gradle-keygen-plugin/1.2.1-develop-SNAPSHOT/ivy-1.2.1-develop-SNAPSHOT.xml).
User `reader`, password `Pa$$w0rd`
See that there is no Maven's `pom`. Ivy has its own metadata, xml with `ivy-module` root tag.
So, it is package system.
I've uploaded package in that repo using [Gradle](https://docs.gradle.org/current/userguide/publishing_ivy.html).

But to avoid long holywar I just removed Ivy from the PR, for now.

grv87:
Reference for metadata: https://ant.apache.org/ivy/history/latest-milestone/ivyfile.html

README.rst
```diff
@@ -705,9 +706,11 @@ Other candidate types to define:
 - `eclipse` for Eclipse projects packages:
 - `gitea` for Gitea-based packages:
 - `gitlab` for Gitlab-based packages:
+- `gradle` for Gradle plugins
 - `guix` for Guix packages:
 - `hackage` for Haskell packages:
 - `haxe` for Haxe packages:
+- `helm` for Kubernetes packages
```
iamwillbar: We should be more precise about what `helm` represents.

```suggestion
- `helm` for Helm charts
```

PR #70, opened 2019-11-07
clear up ambiguity with RPM epochs

### LINKED ISSUE #69: ambiguity: RPM versions with an epoch
The README says:
> the version is the combined epoch (if not 0), version and release of an RPM.
but the list of RPM examples doesn't include any with an epoch:
```
pkg:rpm/opensuse/curl@7.56.1-1.1.?arch=i386&distro=opensuse-tumbleweed
```
**Question:** how should we combine the epoch with the version?

bradcupit:
Closes #69

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -667,14 +667,17 @@ candidate list further down.
   - the `namespace` is the vendor such as fedora or opensuse
     It is not case sensitive and must be lowercased.
   - the `name` is the RPM name and is case sensitive.
-  - the `version` is the combined epoch (if not 0), version and release of an
-    RPM.
+  - the `version` is the combined version and release of an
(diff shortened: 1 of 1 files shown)


bradcupit:
@iamwillbar I added the requested changes. Let me know if this works for you!

iamwillbar:
@wgwoods do you have any thoughts on this before I go ahead and merge it?

wgwoods:
I want to agree with the notion here but - ugly though it is - I'm not sure whether you can consider epoch as optional separate from version? (More details in [my comment in #69](https://github.com/package-url/purl-spec/issues/69#issuecomment-555180371))

PR #71, opened 2019-11-27
Fix docker example with repository_url

dol:
The docker example where not correct according to the spec. E.g. https://github.com/package-url/purl-spec/blob/18fa4e7b3eeb87d529318592f92a40e964f5f2e5/test-suite-data.json#L89

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -114,7 +114,7 @@ Some `purl` examples
     pkg:deb/debian/curl@7.50.3-1?arch=i386&distro=jessie
 
     pkg:docker/cassandra@sha256:244fd47e07d1004f0aed9c
-    pkg:docker/gcr.io/customer/dockerimage@sha256:244fd47e07d1004f0aed9c
+    pkg:docker/customer/dockerimage@sha256:244fd47e07d1004f0aed9c?repository_url=gcr.io
 
     pkg:gem/jruby-launcher@1.1.2?platform=java
     pkg:gem/ruby-advisory-db-check@0.12.4
@@ -543,7 +543,7 @@ candidate list further down.
 
         pkg:docker/cassandra@latest
         pkg:docker/smartentry/debian@dc437cc87d10
-        pkg:docker/gcr.io/customer/dockerimage@sha256%3A244fd47e07d10
+        pkg:docker/customer/dockerimage@sha256%3A244fd47e07d10?repository_url=gcr.io
 
 
 - `gem` for Rubygems:


README.rst
```diff
@@ -114,7 +114,7 @@ Some `purl` examples
     pkg:deb/debian/curl@7.50.3-1?arch=i386&distro=jessie
 
     pkg:docker/cassandra@sha256:244fd47e07d1004f0aed9c
-    pkg:docker/gcr.io/customer/dockerimage@sha256:244fd47e07d1004f0aed9c
+    pkg:docker/customer/dockerimage@sha256:244fd47e07d1004f0aed9c?repository_url=gcr.io
```
e-backmark-ericsson: As I read the spec it should be perfectly fine to let the 'namespace' part of the purl be 'gcr.io/customer' as in the example already given. Do you read it differently? https://github.com/package-url/purl-spec#rules-for-each-purl-component

dol: According to the '[Docker Registry HTTP API V2](https://docs.docker.com/registry/spec/api/#overview)' definition of a name can be composed from one or more component paths.
When parsing a docker package definition it's not clear if there is an repository present or not.

Does the first part after `pkg:docker/<optional_repository>` overwrite the default repository name? If so, what is the rule? If 3 or more path component parts are present, then consider the first element as repository. This only works if every implementation of the 'Docker Registry HTTP API V2' limits the name to two path components.

E.g:
This is a possible namespace without repository declaration
pkg:docker/firstpart/secondpart/thirdpart/dockerimage@sha256:244fd47e07d1004f0aed9c

I hope that clears up the pull request.

stevespringett:
This 'fix' breaks the spec and is likely due to naming confusion.

PURL has standardized on using `repository_url` to refer to non-default repo locations. The default repo for Docker is `https://hub.docker.com`. If an org has their own Docker registry, then they would specify `repository_url` with the URL to their registry. The word 'repository` in PURL refers to a registry in Docker.

In the test suite:
```
pkg:docker/customer/dockerimage@sha256:244fd47e07d1004f0aed9c?repository_url=gcr.io
```
gcr.io is specified as the registry since it's image is not located in the default of hub.docker.com. The existing test suite is correct.

dol:
@stevespringett Isn't what you describing exactly what this PR/fix is about.
Do you think `pkg:docker/gcr.io/customer/dockerimage@sha256:244fd47e07d1004f0aed9c` is correct.

stevespringett:
@dol no, it is not correct. Doh. I thought this PR was fixing the actual test case, which is correct. Didn't realize the README was incorrect. Need more coffee.

Yes, this PR should be merged!

dol:
@stevespringett No problem. I found out about the problem in the doc by looking at the test cases. ;-)

dol:
### Note for the future
Naming of PURL and 'Docker Registry HTTP API V2' are a bit confusing
**[PURL naming](https://github.com/package-url/purl-spec#purl) <-> [Docker Registry HTTP API V2](https://docs.docker.com/registry/spec/api/#overview#overview)**
repository <-> registry
namespace <-> repository name

PR #75, opened 2020-04-01
Add cargo in purl-spec

TG1999:
Signed-off-by: TG1999 <tushar.goel.dav@gmail.com>

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -506,6 +506,18 @@ candidate list further down.
         pkg:bitbucket/birkenfeld/pygments-main@244fd47e07d1014f0aed9c
 
 
+- `cargo` for Rust:
+
+  - The default repository is `https://crates.io/`
+  - The `name` is the repository name.
+  - The `version` is the package version.
+  - Examples::
+
+        pkg:cargo/rand@0.7.2
+        pkg:cargo/clap@2.33.0
+        pkg:cargo/structopt@0.3.11
+
+
 - `composer` for Composer PHP packages:
 
   - The default repository is `https://packagist.org`
@@ -525,7 +537,7 @@ candidate list further down.
   - The `namespace` is the "vendor" name such as "debian" or "ubuntu".
     It is not case sensitive and must be lowercased.
   - The `name` is not case sensitive and must be lowercased.
-  - The `version` is is the package version.
+  - The `version` is the package version.
   - `arch` is the `qualifiers` `key` for a package architecture
   - Examples::
 
@@ -704,7 +716,6 @@ Other candidate types to define:
 - `bower` for Bower JavaScript packages:
 - `brew` for Homebrew packages:
 - `buildroot` for Buildroot packages
-- `cargo` for Rust packages:
 - `carthage` for Cocoapods Cocoa packages:
 - `chef` for Chef packages:
 - `chocolatey` for Chocolatey packages


stevespringett:
Is cargo case-sensitive? If so, this should be mentioned. Are there any special character concerns?

TG1999:
No special character concerns, crates does not contain special characters, they contain a '-' & '_' only, and upper case letters are not allowed in crate packages

sschuberth:
I just stumbled upon this after a discussion with @pombredanne as I'm wondering whether the *package type* for Cargo should be "crate" instead of "cargo". Using "crate" would seem more consistent to me compared to existing package types. @TG1999, any opinion here?

TG1999:
Hi @sschuberth I think crate makes more sense here, @pombredanne thoughts on this :) ?

ja98200wilcox:
>"crate" instead of "cargo"
This seems like a breaking change in nature.  What would happen to all the users who have implemented support around cargo?

DarthHater:
I'm with @ja98200wilcox here, and also for another reason. There are other ecosystems such as maven, where technically maven would be the incorrect term as well, given maven has many different repositories, and central is only one of them. It gets even funnier when you think about other tools like gradle, that use a mostly maven compliant (but not 100%) format. I think of the type as more representative of "packaging format", which cargo seems to represent fine, and crate could too. Given this has already been adopted (cargo), I think that it could cause issues, especially in a backward compatible fashion.

sschuberth:
> This seems like a breaking change. What would happen to all the users who have implemented support around cargo?

Frankly, to me that's a separate question. First, I'd like to clarify what's correct according to the intention of the PURL spec. Then, we can discuss if and how to address it.

> There are other ecosystems such as maven, where technically maven would be the incorrect term as well, given maven has many different repositories, and central is only one of them.

I beg to differ here. Of couse, "Maven" != "Maven Central", but what all *Maven* repositories have in common is a fixed directory layout, and that layout is commonly called "Maven layout". Also, while the package type implies a default package registry, the package type does not necessarily need to be named after the host name of the default package registry.

> It gets even funnier when you think about other tools like gradle or sbt, that use a mostly maven compliant (but not 100%) format.

To the best of my knowledge both Gradle and sbt are 100% Maven compliant, regarding both consuming and publishing Maven artifacts.

> I think of the type as more representative of "packaging format", which cargo seems to represent fine, and crate could too.

I initially had that thinking, too, which is why I had reached out to @pombredanne by chat to clarify. In that conversation he told me to think of the "package type" rather in terms of the "ecosystem", which includes protocols, formats and conventions.

It also helped me personally to further distinguish the "package type" from the "package format" (which is not directly covered in PURL, but could be via qualifiers). For Maven, the package type would be "Maven", but the package format would be "jar", "war", "aar" etc.

The general problem with the term "Maven" is that it can actually refer to many things: The build system, the kind of artifact, or the server layout. This all forms an ecosystem, so it's IMO fair to refer to that using "Maven" as the PURL package type.

I regard Cargo as just one arbitrary build system for Rust that can consume Crates, similar to like Gradle can consume Maven artifacts. So why would you in the Gradle / Maven case name the package type *not* after the build system (Gradle in this case), but for the Cargo / Crate case you *do* name it after the build system (Cargo in this case)?

DarthHater:
Diminishing cargo down to just a build system is a bit odd, it's also a package manager, and a build system, and it has it's own protocol for talking to crates.io, due to being a package manager. It's very synonymous with maven, and it's hard to call it arbitrary given it's pretty much THEE package manager for Rust (the docs for Rust sure suggest so). Obviously Maven isn't java, but it can be used to compile your java programs, much in the same way that Cargo can be used to call rustc, etc...

sschuberth:
> Diminishing cargo down to just a build system is a bit odd, it's also a package manager

I was actually using the term "build system" synonymously to "package manager" here, just to avoid to type both.

> it's hard to call it arbitrary given it's pretty much THEE package manager for Rust

Just like Maven was THE package manager for Java before there was Gradle. Thus, I believe we should decouple the PURL package type from the package manager name whenever possible in a meaningful way, and to me using "crate" is more meaningful than using "cargo". But I'm not a Rust guy after all.

I'm perfectly fine if the consensus here will be something else than what I personally believe to be correct, but I'd like us to come up with a *clear* definition of the package type in this regard, and to come up with a convincing rationale why currently the package type sometimes is named after the package manager (like Cargo), but sometimes after the package registry (PyPI), and other times after the ecosystem (Maven).

Issue #74, opened 2020-03-09
Add Sonatype OSS as adopter 


sbs2001:
https://ossindex.sonatype.org/doc/coordinates here they clearly mention they use purl . By the way this implies the whole Nexus platform is using purl is some way or another. 

Definitely add this in our README

stevespringett:
Correct. OSS Index supports it as does Nexus IQ (part of Nexus Lifecycle)

brianf:
I'll make the addition

pombredanne:
:+1:

PR #78, opened 2020-04-01
Issue #74 - add OSSindex / Lifecycle as known users

brianf:
https://github.com/package-url/purl-spec/issues/74


stevespringett:
Thanks Brian

PR #82, opened 2020-04-15
Maven artifacts are defined by classifier and extension

hboutemy:
type is an attribute of dependencies, that defines extension and default classifier,
see http://maven.apache.org/ref/current/maven-model/maven.html#class_dependency

perhaps this updates arrives a little bit late in time

for compatibility with previous inexact definition, old `type` can be considered as extensions: I suppose nobody used default types where extension does not match type http://maven.apache.org/ref/3.6.3/maven-core/artifact-handlers.html

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -634,19 +634,20 @@ candidate list further down.
 
   - The default repository is `https://repo.maven.apache.org/maven2`
   - The group id is the `namespace` and the artifact id is the `name`
-  - Known `qualifiers` keys are: `classifier` and `type` as defined in the
-    POM documentation. Note that Maven uses a concept / coordinate called packaging
-    which does not map directly 1:1 to a file extension. In this use case, we need
-    to construct a link to one of many possible artifacts. Maven itself uses type 
-    in a dependency declaration when needed to disambiguate between them.
+  - Known `qualifiers` keys are: `classifier` and `extension` (defaults to `jar`).
+    Note that Maven uses a concept of dependency `type`, as defined in the
+    POM documentation, which does not map directly 1:1 to a file extension.
+    Maven itself uses type in a dependency declaration to define a default extension
+    and classifier and a few attributes to know how to manage the file (see default artifact
+    handlers).
   - Examples::
 
         pkg:maven/org.apache.xmlgraphics/batik-anim@1.9.1
-        pkg:maven/org.apache.xmlgraphics/batik-anim@1.9.1?type=pom
+        pkg:maven/org.apache.xmlgraphics/batik-anim@1.9.1?extension=pom
         pkg:maven/org.apache.xmlgraphics/batik-anim@1.9.1?classifier=sources
-        pkg:maven/org.apache.xmlgraphics/batik-anim@1.9.1?type=zip&classifier=dist
-        pkg:maven/net.sf.jacob-projec/jacob@1.14.3?classifier=x86&type=dll
-        pkg:maven/net.sf.jacob-projec/jacob@1.14.3?classifier=x64&type=dll
+        pkg:maven/org.apache.xmlgraphics/batik-anim@1.9.1?extension=zip&classifier=dist
+        pkg:maven/net.sf.jacob-projec/jacob@1.14.3?classifier=x86&extension=dll
+        pkg:maven/net.sf.jacob-projec/jacob@1.14.3?classifier=x64&extension=dll
 
 
 - `npm` for Node NPM packages:


brianf:
When I used Type, it was intentional because I felt like it was more common for people to refer to coordinates using type than by extension and allow the maven system to figure out what extension was implied. It's unclear which is actually more correct, it depends on your POV.

hboutemy:
yes, you're right: people use "type" because it's written in pom.xml without really knowing how type is different from extension (and that's our Maven project fault of having taken time to explain things clearly, because that's not so easy to find concise explanation...)

now that I improved [Maven core documentation on artifact-handlers](http://maven.apache.org/ref/3.6.3/maven-core/artifact-handlers.html) (1,5 years ago...), it's clear to me that "type" refers to **dependency** type, and that **file/artifact** extension and classifier are deducted from dependency type, with ability to override classifier when transforming a dependency into an artifact

there are multiple consequences that influence choice for purl:
1. there are multiple combinations of Maven type + classifier that lead to the same target file, where one file has only one extension + classifier (for example `artifact-tests.jar` can be referenced as a dependency with `test-jar` type only (no classifier), `test-jar` type and `tests` classifier, or `jar` type and `tests` classifier or many other combinations given how many Maven default types are configured with jar extension)
2. the list of types is open, if someone defines custom types in Maven, with non-empty default classifier, purl can't guess: purl's type can only be strictly equivalent to file extension, not really Maven's type

I'm now convinced that:
1. for the purpose of purl, file/artifact extension+classifier is more adapted, then should be a target
2. it will require explanation to users, who will ask why it differs from <dependency><type> written in pom.xml
3. purl should support type for transition, with limitation that it's not exactly type as in Maven dependencies: type in purl is equivalent to extension, where in Maven, if type=`test-jar` and no classifier is provided, file extension will be `jar` and file classifier will be `tests`, unlike in purl where there will be no classifier and file extension will be `test-jar`

WDYT?

jdillon:
I do not think the semantics of maven2 artifact handling need to or should surface to package-url specification.

For me personally it boils down to if you look at this from a reference of the pom.xml and the <dependency> or if you look at it from generating of a URL to fetch the content from a maven2 repository.  These are the main points where the difference between "type" and "extension" are relevant.

FTR if you are considering maven1 and maven2 which are mostly compatible but not wholly compatible as independent coordinates "type" and "extension" have more complexities as m1 has "type" and m2 has "extension".

I believe that the current expression of package-url "type" is reasonable as in *most* cases the "type" is the same as "extension".  For m1 there is no extension so that is meaningless and pkg:maven doesn't impl a version of the impl but instead the basic coordinates of the datatype for the tool.

I do not think that package-url should be concerned with the details of the maven-specific implementation which for m2 and m1 is significantly  different and for m2 has different and implications for how type and packaging can be considered based on the runtime path of the maven reactor invoking the build.

If you consider m1 and m2 (and m3+ whatever might come) and consider the complexity of type/extension/classifier mapping which shouldn't be anything a specification should be done.
And you consider the primary use-cases of package-url wrt to maven artifacts....

My impression is that the value of package-url here is to define something much closer to a <dependency> reference and not a <url> to download the thing from. and such as that I think that ?type is correct and that ?extension should be left to platform specific interpretation.

I totally get that @hboutemy you worked on the internals of this but as the best representation of the "coordinates" of an maven artifact I think the <dependency> representation is a lot closer to what is needed than the details of how you take coordinates and translated that into a URL.

stevespringett:
When CycloneDX creates a BOM which includes a PURL, it specifically does this:

```java
qualifiers.put("type", artifact.getType());
```

There is a one-to-one relationship between a Maven Artifacts type and the `type` qualifier in PURL.

Agree with @jdillon.

jdillon:
If we look at the spec of how you use the coordinate information, its via pom.xml <dependency> and I believe that packaged-url for maven type should follow this, it is essentially the public contract for coordinate information; and how it actually translates that is very different and highly version specific.

jdillon:
-1 ... leave "type" asis

hboutemy:
ok to stay with type in purl spec: explanation is too complex, let's stay simple

but just know that if you try to use CycloneDX generated BOM with a dependency on an artifact with pom.xml type = `test-jar`, `ejb`, `ejb-client`, you'll see a few cases where `qualifiers.put("type", artifact.getType());` should be replaced using Maven AtifactHandler, because I doubt you want to generate `pkg:maven/.../...@...?type=test-jar` but instead `pkg:maven/.../...@...?type=jar&classifier=tests`

but at least, that discussion happens in CycloneDX BOM generator, not purl spec

PR #92, opened 2020-10-01
Proposal: Add homepage_url

david-a-wheeler:
Many software packages do not have a publicly-accessible
version control system (VCS) nor a public download page.
If package-URL is to be able to identify arbitrary packages it
must be able to handle these cases. In addition, just having the VCS
information is sometimes not enough to learn more about the software.

This proposal adds `homepage_url` to `generic`. This makes it
possible to refer to these (many) programs, along with their version numbers.
The homepage URL is also really helpful to users, who may want to
know more about the program and the other information may not be enough
to help.

The CII Best Practices badge has had great success using
homepage URL and VCS URL as a unique project identifier, even for projects
which aren't under the control of a language-level packaging system.

This commit also adds information on how to handle multiple qualifier values
(separate with "&" and sort keys using the "C" locale).

Signed-off-by: David A. Wheeler <dwheeler@dwheeler.com>

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -91,6 +91,8 @@ The defintion for each components is:
 - **version**: the version of the package. Optional.
 - **qualifiers**: extra qualifying data for a package such as an OS,
   architecture, a distro, etc. Optional and type-specific.
+  Multiple qualifiers must be separated with "&" and the keys should be
+  sorted by key in the "C" locale.
 - **subpath**: extra subpath within a package, relative to the package root.
   Optional.
 
@@ -580,6 +582,11 @@ candidate list further down.
     the `generic` type and eventually contributed back to this specification
   - as for other `type`, the `name` component is mandatory. In the worst case
     it can be a file or directory name.
+  - A `homepage_url` may be provided in `qualifiers`; this identifies the
+    main "home page" for users to learn more about the software.
+    This is especially useful when no public `download_url` and `vcs_url`
+    are available.  It's also useful to help users find the correct
+    package home page.
   - Examples (truncated for brevity)::
 
        pkg:generic/openssl@1.1.10g


README.rst
```diff
@@ -91,6 +91,8 @@ The defintion for each components is:
 - **version**: the version of the package. Optional.
 - **qualifiers**: extra qualifying data for a package such as an OS,
   architecture, a distro, etc. Optional and type-specific.
+  Multiple qualifiers must be separated with "&" and the keys should be
```
pombredanne, line 94: Do you mind to put that in a separate PR?
I agree that some improvement would be welcomed there, but you definition may conflict with the spec:

In https://github.com/package-url/purl-spec/tree/ef3fa021b54e3e6014bc6d435cc497db030e234b#how-to-build-purl-string-from-its-components we have:
> sort this list of qualifier strings lexicographically 


And in https://github.com/package-url/purl-spec/tree/ef3fa021b54e3e6014bc6d435cc497db030e234b#a-purl-is-a-url 
> purl qualifiers: this maps to a URL query

So all in all IMHO we could in this area avoid reference to sorting and just introduce that qualifiers are `key=value` pairs separated by `&` using the same syntax as a URL query.

And add a mention about sorting being the "canonical" form in the qualifiers definition in https://github.com/package-url/purl-spec/tree/ef3fa021b54e3e6014bc6d435cc497db030e234b#rules-for-each-purl-component ?

README.rst
```diff
@@ -580,6 +582,11 @@ candidate list further down.
     the `generic` type and eventually contributed back to this specification
   - as for other `type`, the `name` component is mandatory. In the worst case
     it can be a file or directory name.
+  - A `homepage_url` may be provided in `qualifiers`; this identifies the
```
pombredanne, line 585: LGTM! I can merge alright but without the qualifiers part if that's OK with you.

pombredanne (changes requested):
Thank you ++
See my nitpicking inline for your consideration

README.rst
david-a-wheeler: @pombredanne - a separate PR would be great! I probably should have done that anyway.

README.rst
david-a-wheeler: I don't understand what you mean by "without the qualifiers part". It needs to go somewhere, and presumably that would be in the qualifiers. If you just think it's clear from context, I guess that's fine as long as it's clear. We all want a clear spec :-).

pombredanne: Sorry if this came out as not really clear: by "without the qualifiers part" I meant without the part about qualifiers above https://github.com/package-url/purl-spec/pull/92/files#r504010637 that would be best in a separate PR.

david-a-wheeler: Oh I see! Sure, that'd be great. Improvements are improvements.

README.rst
```diff
@@ -91,6 +91,8 @@ The defintion for each components is:
 - **version**: the version of the package. Optional.
 - **qualifiers**: extra qualifying data for a package such as an OS,
   architecture, a distro, etc. Optional and type-specific.
+  Multiple qualifiers must be separated with "&" and the keys should be
+  sorted by key in the "C" locale.
```
pombredanne, line 95: BTW, just a side note for the wording for this new PR: `sorted by key in the "C" locale.` should be clarified and/or simplified. Also I am not sure this needs to be made so prominent as we do not provide so much details at that level for other components. 
Note that the purl is already specified as being ASCII-only and the qualifier sorting is stated as `sort this list of qualifier strings lexicographically` which would be clear with ASCII and unambiguous as the keys are specified as being lowercase only, and there cannot be two qualifiers with the same key.   This boils down to sorting digits first in ascending numeric order then letters in alphabetic order which is alright the C locale, but is also quite simply the default sort order in Python, Ruby, JavaScript. If we introduce the "C" locale there is a bit more complexity to process. 

But to your point making the syntax and sorting should be made more prominent, likely under the "Rules for each purl component/qualifiers"

david-a-wheeler: "Sorted by key lexicographically" is better than being unstated. I was trying for maximum precision. As long as there are only ASCII digits and lowercase letters I *think* locale doesn't matter, though there may be an exception.

README.rst
pombredanne: At this stage the latest spec states:

> This is a query string composed of zero or more key=value pairs each separated by a '&' ampersand. A key and value are separated by the equal '=' character

Therefore I think we are good wrt. the first part of this PR which is about `Multiple qualifiers must be separated with "&"`

README.rst
pombredanne: I think we are good with lexicographic sorting (being a fancy word for alphabetical/alphanumeric) as we have only ASCII as specified. This was not raised as an issue since so we are good this way. I found that invoking a "C" locale has been raising more questions and head scratching that help

pombredanne (commented):
Based on a chat with @johnmhoran,  this PR is now stale. We could still add a homepage_url qualifier to the generic type, but this cannot merged as-is 

See https://github.com/package-url/purl-spec/issues/339 for an issue to track this. 

@david-a-wheeler I am closing this now and you are welcome to submit a new PR on this.

Issue #84, opened 2020-05-07
Wildcards in purl?


copernico:
In certain applications it would make sense to specify certain parts of purl's as intervals or wildcards.

Examples:
  * `pkg:npm/foobar@12.3.*`
  * `pkg:npm/foobar@12.3.[0..11]`

Is this covered in the current specification? Will it be? Or is something that is out of scope (e.g., because consumers of purl can still decide to parse strings containing such wildcard expressions according to application-specific semantics?).

Thanks, keep up the good work!

stevespringett:
The spec does not provide this capability. There are currently no plans to support this as there have been no proposals that work across all ecosystems. Related to #66

copernico:
@stevespringett Thanks for your reply. In our project, we're settling on using a super-set of purl, where version intervals and wildcards are allowed, but those "wildcard" expressions are ultimately expanded to "constant" purls.

A more general question is: how uncommon are scenarios where one needs to do something similar to what we are doing...because if that is a common need, perhaps, before everyone diverges and adopts a different solution, some alignment might desirable....

stevespringett:
There is certainly a need to have ranges and wildcards. Lots of orgs and various projects have a need for it. However, I have not seen any proposal set forth that would:

1) Provide the ability to specify wildcards.
2) Address the impact it would have on location.

For example, if a wildcard or range is specified, how does that impact WHERE the package is located? Case in point, some packages start out being published in one package repo, and move to being published to a different package repo. A good example of this is https://github.com/everit-org/json-schema/ where the project started publishing artifacts to Maven Central, but moved to Jitpack for increased security and transparency in the resulting artifact.

Since PURL has a default package repository for most PURL types, a wildcard or range would assume the location remains constant, when it fact it may not.

Proposals welcome.

pombredanne:
@copernico @stevespringett here are some thoughts on the topic
(re-posted from https://github.com/ossf/wg-vulnerability-disclosures/issues/28#issuecomment-687833562 )

Version ranges (or more generally version constraints and that would include any wildcards or globs) are not an easy thing as there cannot be a universal definition since there is no universal way to represent and compare versions across all package types. 

To add them to purl, we could either have a partly universal/partly type-specific approach or be entirely package type-specific.

#### a partly universal/partly type-specific approach

- Define a base `universal notation` (universal as used for any package type / ecosystem)
[...]

PR #79, opened 2020-04-02
Separate the core syntax definition from purl type definitions. 

stevespringett:
Signed-off-by: stevespringett

Issue #66, opened 2019-10-16
Version range


iamwillbar:
Is there desire for PURL to support version ranges or is that out of scope? For example, to describe vulnerable versions of a package.

brianf:
If nothing else, capturing the various ecosystem version sorting is really key. Then you could at least use purl ids as upper and lower bounds and still evaluate what's in between.

pombredanne:
@iamwillbar there is a need a alright to have a common way to express version ranges... but I wonder if this is possible, because there is no universal way to express versions. Semver comes close (but cannot handle some epochs or Debian "or" AFAIK)


@brianf 
There is no universal way to compare versions too (which is why RPM- and Debian-based distros had to adopt a concept of epoch) so documenting the way things are compared for each package `type` would be great as a start.

@iamwillbar if you were to provide some unified specification for version ranges what would it look like?
(leaving aside for now if this could be stuffed in a PURL or not)

mprpic:
@pombredanne I suppose nothing is preventing purl users from specifying the versioning scheme in a qualifier, e.g. `pkg:pypi/django@1.11.1?version_scheme=semver`. Given a set of these, you could order them and determine vulnerable versions by comparing against a known fixed version. This however falls apart as soon as you have multiple supported version streams that are updated independently (as is the case for Django). There are a ton of "standards" on version range specifications:

- https://maven.apache.org/enforcer/enforcer-rules/versionRanges.html
- https://semver.npmjs.com/
- https://www.python.org/dev/peps/pep-0440/#compatible-release

If we were to use the Python one, and use [this security release](https://www.djangoproject.com/weblog/2020/sep/01/security-releases/) as an example, the purls for the released versions could be:

- `pkg:pypi/django@3.1.1?version_scheme=semver&vulnerable_versions=3.1.*`
- `pkg:pypi/django@3.0.10?version_scheme=semver&vulnerable_versions=3.0.*`
- `pkg:pypi/django@2.2.16?version_scheme=semver&vulnerable_versions=2.2.*`

Mandatory qualifiers are not a thing in the spec however, so this would solely depend on maintainers of said projects to use them.

pombredanne:
@mprpic 

> I suppose nothing is preventing purl users from specifying the versioning scheme in a qualifier, e.g. pkg:pypi/django@1.11.1?version_scheme=semver. 

Sure thing nothing prevents this, and we could even make it part of the spec too.
[...]

PR #93, opened 2020-10-01
Proposal: Add version ranges

david-a-wheeler:
Package-URL by itself isn't very good for reporting about
vulnerabilities because it cannot report version ranges.
CPE, by contrast, *can* report version ranges.

This commit proposes adding version range support to package_URL.
I think node-semver is pretty common, so I referenced that; other
options are possible too. Note that even if you don't use semver,
the version range still works, since hierarchical numbers are very common.

Signed-off-by: David A. Wheeler <dwheeler@dwheeler.com>

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -88,7 +88,7 @@ The defintion for each components is:
 - **namespace**: some name prefix such as a Maven groupid, a Docker image owner,
   a GitHub user or organization. Optional and type-specific.
 - **name**: the name of the package. Required.
-- **version**: the version of the package. Optional.
+- **version**: the version or version specifier of the package. Optional.
 - **qualifiers**: extra qualifying data for a package such as an OS,
   architecture, a distro, etc. Optional and type-specific.
 - **subpath**: extra subpath within a package, relative to the package root.
@@ -257,6 +257,14 @@ The rules for each component are:
     conventions such as semver for NPMs or nevra conventions for RPMS. A `type`
     may define a procedure to compare and sort versions, but there is no
     reliable and uniform way to do such comparison consistently.
+  - Version ranges may also be provided for search capabilities.
+    For example, in a vulnerability report it
+    may be important to say "versions 1.1.3 through 4.2.6 are vulnerable".
+    Ranges may be specified following
+    [node-semver](https://github.com/npm/node-semver)i syntax; note that
+    spaces must be encoded as `%20%`.
+    Note that version ranges are only intended for use in searches, and
+    should NOT be used when specifying the content of an actual package.
 
 
 - **qualifiers**:


stevespringett:
This PR does not address issues identified in #66 and and #84.

david-a-wheeler:
Thanks for the cross-reference!

I think this *does* handle Debian OR. Doesn't "||" provide the same functionality? If not, please enlighten me!

Handling epochs is a bigger issue. But epochs work very much like a "hidden 0. prefix" on a typical version number, and people who follow SemVer won't have epochs. So I think this could be easily modified to support that case. Something like "Versions that begin with the form NUMBER: are presumed to have an epoch of that number; if unstated the epoch is 0. Epochs act as the topmost number in a version."

I'm absolutely *not* married to node-semvar range syntax; others would be fine. However, I think it's important to have *some* syntax for it, and I thought it'd be easier to start with a specific concrete proposal.

I did pick node-semvar intentionally. One reason is that it can express some fairly complex situations. For example, it can relatively easily handle the "multiple streams of versions" case.

stevespringett:
In addition to the above, one additional aspect that will need to be addressed is the default repository url for each PURL type. Introducing a range may alter where a component is located, so we'll need to specify what that means.

E.g. 
* Does a range eliminate the default repository url entirely? 
* Does a range continue to have a default repository url, but a list of alternatives may be specified? 
* Does the range only work if the location does not change and a new range has to be used for those where the repo url is different? 
* Something else?

david-a-wheeler:
@stevespringett - I think that's backwards. Instead, the version numbers necessarily only apply to the given repo URL (default URL if there's a default and no other URL is given, or a specified URL if one is specified).

For example, there are many different programs all called "zlib"; some of them have the same underlying origin. If I specify version X.Y.Z, that version number would only apply to the repo URL given. If there's a different repo URL, there's no reason to believe that version X.Y.Z refers to the same thing.

johnmod3:
With NIST moving away from CPE to SWID, hows does this impact PURL?

david-a-wheeler:
I believe SWID has two big problems: (1) it's a huge XML file, making processing it very inconvenient, and (2) it can't refer to version ranges.  This makes moving from CPEs hard. A big advantage of SWID over package-URL is that SWID can refer to closed source software where no VCS or download source is publicly visible.

If pURL supports version ranges and has a mechanism for referring to software home pages (or something else that supposed closed source software), then package-URL suddenly has big advantages over SWID for many purposes.

johnmod3:
Hi David, totally agree - but NIST seems to be hell bound to go with SWID (even though Microsoft doesn't seem to use it any longer) - JC has way more context

stevespringett:
David. I think the zlib example is incomplete. To my knowledge, the zlib project does not have a default distribution repository. However, distributions that package zlib and distribute it to Conan, Buckaroo, Anaconda, Debian, Redhat, etc, will have a default repository.

I think a cleaner example (and more representative of the majority of software packages) are those that distribute the package to a repository when released. Typically, each ecosystem will have a default package repository. npm.js for Node.js, Maven Central for Java/Maven projects, etc.

My prior point was that if you introduce version ranges, we need to also account for how those package versions can be resolved.

For example, the https://github.com/everit-org/json-schema/  project previously distributed all their compiled packages to Maven Central. However, Maven Central lacks certain security controls and transparency, so the json-schema project moved the default repository for the project to use Jitpack instead. Newer versions of json-schema are not available on Maven Central. Therefore, version ranges need to account for these scenarios.

The first sentence of the PURL spec is:
> A purl or package URL is an attempt to standardize existing approaches to reliably identify and locate software packages.

So location information needs to be addressed in version ranges when the default repository moves between versions.

stevespringett:
Also FYI, I chatted with @pombredanne about trying to represent SWID in PURL, and there seems to be interest in that. 

Also regarding SWID, the attributes of the SoftwareIdentify field can be represented in PURL and can be used for vulnerability use cases - whenever the NVD gets around to supporting it. I have a few different ideas how that can work and will be submitting a PR - however, I've been holding off until #79 is merged.

david-a-wheeler:
I agree that this scenario needs to be addressed. However, I think it would be cleaner for people who need to support ranges to also support multiple package urls separated by whitespace. Then you could have a package URL with a version range and refer to a particular location. This also cleanly deals with the case where version numbers are different for different locations, a common challenge. It also deals with the case where there are many different packages that share a underlying problem, for example, a vulnerability in a specification can cause multiple packages with many versions with them to all apply.

I think that would be much cleaner than trying to link version numbers and repository locations within a single package URL.

kerberosmansour:
I do agree that it should have ranges I added some reasoning behind it [here on one of the OSSF Issues](https://github.com/ossf/wg-vulnerability-disclosures/issues/28#issuecomment-702665084) to @pombredanne

david-a-wheeler:
As I noted earlier, I'm not married to this specific version syntax. There are several to choose from.

Here's another:
https://raw.githubusercontent.com/CVEProject/automation-working-group/master/cve_json_schema/v5.x_discuss/cve513.schema

I think the key is to be able to support a range of version numbers. The "special" case is epoch numbers, which many systems can't handle, but as I noted earlier I think those are relatively easy to add.

pombredanne:
@david-a-wheeler first thank you++ for this!

I am not too much in favor of overloading the `version` with a ranges syntax as I feel it will be a source of confusion. I would rather craft a new qualifier for that instead. What do you think?
There are also considerations wrt. the many versioning schemes and how versions cane be compared that I collected in https://github.com/package-url/purl-spec/issues/84#issuecomment-707766899

pombredanne:
@stevespringett re:

> My prior point was that if you introduce version ranges, we need to also account for how those package versions can be resolved.

It makes sense, and in particular how they are compared matters quite a bit.

> For example, the https://github.com/everit-org/json-schema/ project previously distributed all their compiled packages to Maven Central. However, Maven Central lacks certain security controls and transparency, so the json-schema project moved the default repository for the project to use Jitpack instead. Newer versions of json-schema are not available on Maven Central. Therefore, version ranges need to account for these scenarios.
> 
> The first sentence of the PURL spec is:
> 
> > A purl or package URL is an attempt to standardize existing approaches to reliably identify and locate software packages.
> 
> So location information needs to be addressed in version ranges when the default repository moves between versions.

IMHO in this case location information would come from specifying a `repository_url=https://jitpack.io` qualifier for these package versions that are there and dealing with this specific package and its move more one place to another would have to be handle specially. It looks like https://github.com/everit-org/json-schema is exceptional enough that it would not deserve a generic treatment.
And actually based on their home page we really have two Package URLs so this is an even better case:
- newer `pkg:maven/com.github.everit-org.json-schema/org.everit.json.schema@1.12.1?repository_url=https://jitpack.io`
- older `pkg:maven/com.github.erosb/everit-json-schema-jdk6@1.9.2` published at Maven Central
(and BTW kudos to @erosb for doing this to avoid confusion)

erosb:
Hello, maven coordinates of the everit-org/json-schema are quite baroque, mostly because I left everit-org (the company) soon after the first release of the json-schema project. So the history is:
 * it is released as `org.everit.json:org.everit.json.schema` in range 1.0.0 - 1.5.1 on maven central
 * the newer versions (1.6.0 - 1.12.1) are primarily on jitpack
 * versions 1.9.2 - 1.12.1 are available both on jitpack and maven central, but on maven central the groupId:artifactId is `com.github.erosb:everit-json-schema` (while on jitpack the same releases are available as `org.everit.json:org.everit.json.schema`)
 * a JDK6-comatible backport of version 1.9.2 was created by a contributor that I released with the coordinates `com.github.erosb:everit-json-schema-jdk6:1.9.2`

I don't expect any tooling to support it :)

kerberosmansour (commented):
I don't see any issues there.

pombredanne:
@david-a-wheeler Please see #93 for an extended take on the same topic, but drafted as a separate mini spec.

README.rst
```diff
@@ -257,6 +257,14 @@ The rules for each component are:
     conventions such as semver for NPMs or nevra conventions for RPMS. A `type`
     may define a procedure to compare and sort versions, but there is no
     reliable and uniform way to do such comparison consistently.
+  - Version ranges may also be provided for search capabilities.
+    For example, in a vulnerability report it
+    may be important to say "versions 1.1.3 through 4.2.6 are vulnerable".
+    Ranges may be specified following
+    [node-semver](https://github.com/npm/node-semver)i syntax; note that
```
isaacs, line 264: I'm in favor of this choice. Makes my life much easier 😂

pombredanne: @isaacs I eventually merged https://github.com/package-url/purl-spec/blob/master/VERSION-RANGE-SPEC.rst which is already in use in CycloneDX and CSAF. We needed to design something that works beyond npm, node-semver ranges.

pombredanne:
@david-a-wheeler I am closing this.
https://github.com/package-url/purl-spec/blob/master/VERSION-RANGE-SPEC.rst has been merged with the #139 and is already in use in CycloneDX and OASIS Common Security Advisory Framework (CSAF).

There is value to spec an optional qualifier for a version range, in the case we have no version, but I do not want to mix a version range with a plain version in the same PURL component, this is going to be messy. 

The follow up for the qualifier is in https://github.com/package-url/purl-spec/issues/340

PR #97, opened 2021-01-30
WiP: Massive refactoring.

KOLANICH:
1. Separated package source types into groups, and groups - into kinds. And defined relations between kinds.
2. Renamed sources related to langs.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -83,8 +83,9 @@ The defintion for each components is:
 - **scheme**: this is the URL scheme with the constant value of "pkg". One of
   the primary reason for this single scheme is to facilitate the future official
   registration of the "pkg" scheme for package URLs. Required.
-- **type**: the package "type" or package "protocol" such as maven, npm, nuget,
-  gem, pypi, etc. Required.
+- **group**: defines the way which types are allowed and how they should be processed. Required.
+- **type**: the package "type" or package "protocol" such as git, svn, dotnet/nuget,
+  ruby/gem, py/pypi, etc. Required.
 - **namespace**: some name prefix such as a Maven groupid, a Docker image owner,
   a GitHub user or organization. Optional and type-specific.
 - **name**: the name of the package. Required.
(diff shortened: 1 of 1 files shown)


pombredanne:
@KOLANICH Thank you for this, but I am not sure I see the value of this proposed changed
1. this makes all existing purl incompatible
2. this creates an additional level in the naming which requires more specification and more thinking when using a url 

Can you elaborate on the benefit?

If anything, my preference would be towards some future simplification of the naming by merging the namespace and name in one rather than splitting in more components.

Also there is only so much information that we can pack into a purl before making it unusable. In particular a purl is not a version control URL by design and mixing them is quickly looking overloaded and hard to read IMHO. Instead I would prefer to reuse https://github.com/spdx/spdx-spec/blob/cfa1b9d08903/chapters/3-package-information.md#37-package-download-location- as mentioned in the spec and if need be a package can be defined by a purl and a VCS URL and more urls.

KOLANICH:
>this makes all existing purl incompatible

Certainly. But any big refactoring of an existing spec does so.

>this creates an additional level in the naming which requires more specification and more thinking when using a url

It is by design. The proposed changes were driven by the needs of `prebuilder` software. Its purpose is to build packages from source code and generate the resulting artifacts and metainformation for them in a form of an installable repo. Currently the info is specified as a hierarchy of python classes, precisely specifying how to retrieve the source and what I want in the end. I could have implemented auto-discovery of the VCS and build system, but I decided that instead a user must specify it explicitly. It is beneficial because

* no network requests is needed to get known this info
* the source can support multiple build systems and VCSes on the same URI, and it is impossible to decide what to use automatically.
* if the source changes the build system, the maintainer must go and manually check that nothing has broken, instead of not failing and silently producing potentially broken packages

I wanna implement support for a declarative config somewhen. In order to do that all the info has to be available within a declarative config.

Since there exists thr standard, it should probably use it. But currently it is not very suitable, it doesn't encode enough information.

So I have added some stuff I need that I feel belongs here, and refactored the spec. While refactoring,  I was trying to make it so, that for some components of the URI a routine can be looked up in a hashtable that would convert the components next to it into classes, interfaces of which are tied to the hierarchy of their nodes in the URI AST. Also some changes are not so technical (i.e. there is usually no technical issue to have managers for `pl`s and `distro`s within the same hashtable) but just organing the info making it easier to read (and write, if autocompletion is implemented) and also reducing probability of name collisions by introducing namespaces.

>Instead I would prefer to reuse https://github.com/spdx/spdx-spec/blob/cfa1b9d08903/chapters/3-package-information.md#37-package-download-location- as mentioned in the spec and if need be a package can be defined by a purl and a VCS URL and more urls.

May make sense, if `/vcs/ ` is not dropped, though it would certainly complicate the parser. Dropping `/vcs/` would ruin the hierarchy, so it contradicts the goal of keeping everything structured enough.

pombredanne:
It seems that what you call a `group` is entirely derivable from a `type`. Can a `type` be in more than one group? I would rather this is as an attribute of a type or a qualifier of a Package URL that you could track on your side, without a need to change anything in the spec; and handling of version control is best handled as a VCS URL which could be separate or used as a qualifier.

> The proposed changes were driven by the needs of prebuilder software.

Fair enough, but I am not sure a general purpose specification that's been adopted would need to be massively refactored by the needs of one tool? 


Note also that is has never been a goal of this spec to document everything about a package; in particular I feel that anything that would tell how to build a package would be best left out

KOLANICH:
>It seems that what you call a group is entirely derivable from a type.

Yes.

>Can a type be in more than one group?

No and yes. No in the sense that items of different groups are mapped to classes with different interfaces (so if one wanted to collapse this, he would have to provide unique names for types names, like `github-releases` and `github-actions`). Yes in the sense that some types are designed to wrap other types, i.e. a Debian package can be built from source code on GitHub.

>that you could track on your side, without a need to change anything in the spec

I certainly can do it on my side. And doing it certainly would require no changes in the spec. But I think the spec needs some improvement. At least the spec text should be structured better and similar things should be grouped together, and duplicates and ambiguites should be eliminated. If someone refers to a GitHub repo using `github`, then what does he mean? Does he mean fetching artifacts from releases, or fetching artifacts from Actions or fetching source code using git and building the package? This PR proposes to leave fetching code using VCS to full URIs and leave `github` (which becomes `releases/gh`) for using Releases.

pombredanne:
@KOLANICH re:

>  At least the spec text should be structured better and similar things should be grouped together, and duplicates and ambiguites should be eliminated.

Then I would prefer that you submit smaller focused changes if this works for you. I am not ready to accept the big change with groups being part of the syntax, but some grouping of  types documentation by related types can make sense. 
I am closing this PR for now then. Feel free to reopen if you feel strongly about it.

KOLANICH:
Feel free to choose and take the changes you want from this PR. I don't know exactly which changes you would like and which you don't, and any hypothesizing will likely be a waste of time, so feel free to do it yourself.

PR #102, opened 2021-05-03
Separate purl specification from type definitions

pombredanne:
This is essentially the same as #79 with a twist to keep the git history.


pombredanne:
@stevespringett I redid your split of the spec vs. definition keep the history attach to the split files with a little bit of git fu such that we have the line-level blame for each file. I also made minor changes and formatting. If this is OK with you this would be merged in favor of #79

stevespringett (approved):
The changes look good to me. Thanks for the git fu @pombredanne.

pombredanne:
Merging this now then! at last!

PR #105, opened 2021-05-06
Fix vcs_url example and encode "@" where necessary

### LINKED ISSUE #104: Examples of vcs_url usage should follow the spec
The definition of the `vcs_url` qualifier says that the value should follow the [SPDX download specification](https://github.com/spdx/spdx-spec/blob/cfa1b9d08903/chapters/3-package-information.md#37-package-download-location), but one of the examples of this qualifier doesn't follow that specification. Specifically, the generic type gives

    pkg:generic/bitwarderl?vcs_url=https://git.fsfe.org/dxtr/bitwarderl@cc55108da32

as an example but the URL would have to be git+https://git.fsfe.org/dxtr/bitwarderl@cc55108da32 to be compliant.

(I can supply a PR once #102 is merged.)

magnusbaeck:
The URL in the generic type's example of how the vcs_url qualifier can be used has been prefixed with "git+" to make it compliant with the SPDX specification. This fixes #104.

Also making sure "@" is percent-encoded in a couple of places where it needs to be encoded according to the [_Character encoding_ section](https://github.com/package-url/purl-spec/blob/master/PURL-SPECIFICATION.rst#character-encoding).

### DIFF
--- a/PURL-TYPES.rst
+++ b/PURL-TYPES.rst
@@ -114,7 +114,7 @@ Known `purl` types
 
        pkg:generic/openssl@1.1.10g
        pkg:generic/openssl@1.1.10g?download_url=https://openssl.org/source/openssl-1.1.0g.tar.gz&checksum=sha256:de4d501267da
-       pkg:generic/bitwarderl?vcs_url=git%2Bhttps://git.fsfe.org/dxtr/bitwarderl@cc55108da32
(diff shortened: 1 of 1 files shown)


iamwillbar (dismissed):

magnusbaeck:
@iamwillbar I've rebased these patches on top of your reformatting commit.

PR #110, opened 2021-06-30
Add type definition for Swift packages

mattt:
This PR adds a type definition for Swift packages, according to the recently-adopted [standards](https://github.com/apple/swift-package-manager/blob/main/Documentation/Registry.md) for registry identifiers.

I also have a Swift implementation of the PURL specification available here: https://github.com/mattt/packageurl-swift.
Please let me know if you'd be interested in moving that into the @package-url organization.

### DIFF
--- a/PURL-TYPES.rst
+++ b/PURL-TYPES.rst
@@ -255,6 +255,37 @@ rpm
       pkg:rpm/fedora/curl@7.50.3-1.fc25?arch=i386&distro=fedora-25
       pkg:rpm/centerim@4.22.10-1.el6?arch=i686&epoch=1&distro=fedora-25
 
+swift
+----
+``swift`` for Swift packages:
+
+- There is currently no default repository for Swift packages.
+- The ``namespace`` is used for the scope of a Swift package.
+  It is not case sensitive and must be lowercased.
+
+  - A package scope consists of alphanumeric characters and hyphens.
+    Hyphens may not occur at the beginning or end,
+    nor consecutively within a scope.
+  - The maximum length of a package scope is 39 characters.
+  - A valid package scope matches the following regular expression pattern
+    ``\A[a-zA-Z0-9](?:[a-zA-Z0-9]|-(?=[a-zA-Z0-9])){0,38}\z``
+
+- The ``name`` is the name declared in the Swift package manifest.
+  It is case and normalization insensitive,
+  and must be normalized using Unicode Normalization Form KC (NFKC)
+  and then transformed with locale-independent case folding.
+
+  - The maximum length of a package name is 128 characters.
+  - A valid package name matches the following regular expression pattern
+    ``\A\p{XID_Start}\p{XID_Continue}{0,127}\z``
+
+- The ``version`` is the version of the package release,
+  and must be a valid SemVer version number.
+- Examples::
+
+      pkg:swift/alamofire/Alamofire@5.4.3
+      pkg:swift/flight-school/Money@1.3.0
+      pkg:swift/realm/SwiftLint@0.43.1
 
 Other candidate types to define:
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
@@ -309,7 +340,6 @@ Other candidate types to define:
 - ``puppet`` for Puppet Forge packages:
 - ``sourceforge`` for Sourceforge-based packages:
 - ``sublime`` for Sublime packages:
-- ``swift`` for Swift packages:
 - ``terraform`` for Terraform modules
 - ``vagrant`` for Vagrant boxes
 - ``vim`` for Vim scripts packages:
--- a/test-suite-data.json
+++ b/test-suite-data.json
@@ -298,5 +298,65 @@
     "qualifiers": {"in production": "true"},
     "subpath": null,
     "is_invalid": true
+  },
+  {
+    "description": "valid Swift purl",
+    "purl": "pkg:swift/mona/LinkedList@1.0.0",
+    "canonical_purl": "pkg:swift/mona/LinkedList@1.0.0",
+    "type": "swift",
+    "namespace": "mona",
+    "name": "LinkedList",
+    "version": "1.0.0",
+    "qualifiers": null,
+    "subpath": null,
+    "is_invalid": false
+  },
+  {
+    "description": "valid Swift purl with canonicalization",
+    "purl": "pkg:swift/MONA/𝒜①@1.0.0",
+    "canonical_purl": "pkg:swift/mona/a1@1.0.0",
+    "type": "swift",
+    "namespace": "mona",
+    "name": "a1",
+    "version": "1.0.0",
+    "qualifiers": null,
+    "subpath": null,
+    "is_invalid": false
+  },
+  {
+    "description": "Swift purl with invalid namespace",
+    "purl": "pkg:swift/--mona--/LinkedList@1.0.0",
+    "canonical_purl": null,
+    "type": "swift",
+    "namespace": "--mona--",
+    "name": "LinkedList",
+    "version": "1.0.0",
+    "qualifiers": null,
+    "subpath": null,
+    "is_invalid": true
+  },
+  {
+    "description": "Swift purl with invalid name",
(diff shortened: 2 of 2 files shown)


iamwillbar (approved):
LGTM

PURL-TYPES.rst
```diff
@@ -255,6 +255,37 @@ rpm
       pkg:rpm/fedora/curl@7.50.3-1.fc25?arch=i386&distro=fedora-25
       pkg:rpm/centerim@4.22.10-1.el6?arch=i686&epoch=1&distro=fedora-25
 
+swift
+----
+``swift`` for Swift packages:
+
+- There is currently no default repository for Swift packages.
+- The ``namespace`` is used for the scope of a Swift package.
+  It is not case sensitive and must be lowercased.
```
stevespringett, line 264: Is namespace really not case sensitive? Take https://swiftpackageregistry.com/SDWebImage/SDWebImage as an example. To my knowledge SDWebImage is both the namespace and name and both have upper and lower case letters. Also to my knowledge, most Swift packages use GitHub which is not case sensitive. But Swift is decentralized and could leverage alternative services which could introduce case sensitivity, correct? Or is my understanding wrong?
 
 Also a question... GitHub uses organizations (or users) and these essentially become the namespace. However, given that Swift packages could be hosted elsewhere, is namespace a requirement? I haven't seen any indication that namespace is required. If this is true, then the spec should state that namespace is optional.

PURL-TYPES.rst
```diff
@@ -255,6 +255,37 @@ rpm
...
+
+  - A package scope consists of alphanumeric characters and hyphens.
+    Hyphens may not occur at the beginning or end,
+    nor consecutively within a scope.
+  - The maximum length of a package scope is 39 characters.
+  - A valid package scope matches the following regular expression pattern
+    ``\A[a-zA-Z0-9](?:[a-zA-Z0-9]|-(?=[a-zA-Z0-9])){0,38}\z``
+
+- The ``name`` is the name declared in the Swift package manifest.
+  It is case and normalization insensitive,
```
stevespringett, line 274: Same question with case sensitivity as with namespace.

PURL-TYPES.rst
```diff
@@ -255,6 +255,37 @@ rpm
...
+  and must be normalized using Unicode Normalization Form KC (NFKC)
+  and then transformed with locale-independent case folding.
+
+  - The maximum length of a package name is 128 characters.
+  - A valid package name matches the following regular expression pattern
+    ``\A\p{XID_Start}\p{XID_Continue}{0,127}\z``
+
+- The ``version`` is the version of the package release,
+  and must be a valid SemVer version number.
```
stevespringett, line 283: Is it possible to specify a specific commit or tag instead of a version? If so, then we need to account for that scenario in the spec.

PURL-TYPES.rst
```diff
@@ -255,6 +255,37 @@ rpm
...
+    ``\A[a-zA-Z0-9](?:[a-zA-Z0-9]|-(?=[a-zA-Z0-9])){0,38}\z``
+
+- The ``name`` is the name declared in the Swift package manifest.
+  It is case and normalization insensitive,
+  and must be normalized using Unicode Normalization Form KC (NFKC)
+  and then transformed with locale-independent case folding.
+
+  - The maximum length of a package name is 128 characters.
+  - A valid package name matches the following regular expression pattern
+    ``\A\p{XID_Start}\p{XID_Continue}{0,127}\z``
```
stevespringett, line 280: IMO if we want to specify a regex pattern for a valid package name, we need to account for the lowest common denominator. XID_Start and XID_Continue are not recognized as valid patterns in the first two regular expression libraries I used. Many regex parsers only support a small subset of capabilities and others will lean towards one dialect of regex syntax. For simplicity, I would recommend using english language to specify what is valid and what is not.

PURL-TYPES.rst
mattt: A package's namespace is case-insensitive and case-preserving. In the case of your example, the canonical namespace is "SDWebImage", and that's what would be returned in JSON responses from the server. However, the namespace is case-insensitive, meaning that the server doesn't distinguish among "sdwebimage", "SDWebImage", or any other strings that differ only in case. This is similar to how GitHub handles URLs to repositories. github.com/sdwebimage/sdwebimage resolves the same as github.com/SDWebImage/SDWebImage

PURL-TYPES.rst
mattt: A package name, like a namespace, is case-insensitive. It's also normalization-insensitive to accommodate the greater range of characters allowed for names, relative to namespaces.

PURL-TYPES.rst
mattt: No, a Swift package release is identified by a registry with only a SemVer version number.

PURL-TYPES.rst
mattt: The regular expression here uses the `\p{UNICODE PROPERTY NAME}` metacharacter described by [ICU pattern syntax](https://unicode-org.github.io/icu/userguide/strings/regexp.html). Unfortunately, it's not feasible to express this using an alternative syntax (see [UAX #31](https://unicode.org/reports/tr31/)).

What do you think about something like this?

> - A valid package name consists of a character with the Unicode property `XID_Start` followed by up to 127 characters with the Unicode property `XID_Continue`, or expressed by ICU regular expression pattern:
     `\A\p{XID_Start}\p{XID_Continue}{0,127}\z`

mattt:
@stevespringett Thanks for your feedback! To respond more directly to https://github.com/package-url/purl-spec/pull/110#discussion_r662328200: 

Swift recently adopted a package registry system, that defines a new scheme for identifying package dependencies. Previously, package dependencies could be declared using a URL to a Git repository (often on GitHub). Swift Package Manager will continue to support this, but the preferred way going forward will be to resolve dependencies with a registry using this new identity scheme.

Most package dependencies will be available both by a package identifier and through a URL. For the purposes of PURL and supply chain modeling, we can understand the former as using the `swift` package type and the latter as using either a `generic` or host-specific (`github` / `bitbucket` / etc.) type.

mattt:
Marking this PR as draft, because we're discussing changes to how packages are named. I'll mark this ready for review once we reach a final consensus. Thanks for your patience as we work through all of this.

PURL-TYPES.rst
```diff
@@ -255,6 +255,37 @@ rpm
...
+
+- The ``version`` is the version of the package release,
+  and must be a valid SemVer version number.
+- Examples::
+
+      pkg:swift/alamofire/Alamofire@5.4.3
```
bradcupit, line 286: Since there's no Swift default package registry just using namespace, name, & version seems like it could be ambiguous. If we had this URL:
```
github.com/alamofire/Alamofire
```
Then Alamofire decided to migrated to gitlab:
```
gitlab.com/alamofire/Alamofire    (not real, I made it up)
```

Maybe they keep their old (now stale) repo on github, but newer versions only go to gitlab.
Which repo does `pkg:swift/alamofire/Alamofire@5.4.3` refer to?

I'm on at Sonatype uses purl internally, and, to avoid ambiguity, we put the hostname in there. Tweaking the above example:
```
pkg:swift/github.com/alamofire/Alamofire@5.4.3
```

That would resolve the ambiguity.

Alternatively we could tack on the repository_url qualifier, but that'd require us defining a default repository for Swift. Which we could do but what would it be? github?

bradcupit: I also found out [OSS Index does the same thing](https://ossindex.sonatype.org/ecosystem/swift) with their Swift purl strings. Their example:

```
pkg:swift/github.com/Alamofire/Alamofire
```

OSS Index is hosted and maintained by a different Sonatype team than the one I represent, but if two teams came to the same conclusion perhaps including the host name is a good idea.

bradcupit: [This competing PR](https://github.com/package-url/purl-spec/pull/116) fits with the comments in this thread, and with what OSS Index does.

full disclosure: the author of that PR and I are on the same team at Sonatype.

Issue #58, opened 2019-04-09
clarify handling of "+" (plus) sign and blanks (spaces)


gernot-h:
As "+" is frequently used within Debian package versions, I'd like to see a clarification (and probably examples/tests for it) whether this needs to be percent-encoded or not. 

This also makes me wonder how blanks (" ") shall be encoded within the different parts of a purl - theoretically, I assume it should be "%20" in most cases? While I would consider "+" way more readable (as in usual HTML form encoding), it probably is just wrong according to RFC3986 if used anywhere else than in the qualifiers, right?

jdillon:
I am not 100% sure on the specifics, but I think that "+" needs to be URL-encoded where used.  and that " " also needs to be URL-encoded.  So that the unencoded component values are the proper values the package system needs to function.

I do not believe that package-url is meant to make fully _readable_ strings, but is more meant to make fully interoperable strings.  The notes about NPM "@scope" being encoded as "%40scope" is good example reference of this.

While I understand that "+" is generally nicer to look at than "%20", the purpose and value of package-url isn't really about making package coordinates easy for humans to read.  It certainly has some provisions for this but its limited, and the spec can no cope with any future unknowns so beyond the given limited set of rules, anything else has to be url-encoded.

gernot-h:
Well, in my use case (and I assume in many others), having readable strings will make it way more easy to introduce package-urls and convince colleagues to use them instead of inventing their own representations for all package managers around.

However, I fully understand and agree that unambiguousness is more important - and given the possible ambiguity of "+" (people or wrong implementations might decode this to " "), I also think that it shall be percent-encoded. 

But I hoped that someone with a clearer understanding of RFC3986 & friends could provide an authoritative answer - but perhaps I should just prepare a pull request with some clarifications and see what happens. :)

gernot-h:
@bufferoverflow, any thoughts on this from your side or any recommendation whom with strong background in URL RFCs we could ask to review this?

tommyknows:
I've just spent some time thinking about this as well, and want to bring in a couple of points to this discussion, as I think the spec in it's current state might not be completely clear about this.

1) The spec does not define how plus-signs or spaces should be encoded.
2) In a URL, a `+` sign is allowed assuming it is *not* in the query-part. In a Query, a plus-sign would need to be encoded to `%2B`. 
3) In a URL, a ` ` (space) is encoded to `%20`. In the Query-part of a URL however, ` ` is encoded to a `+`.

[...]

PR #121, opened 2021-08-28
Correct the url encoded symbol

vBarbaros:
The old value contained the hex equivalent of the symbol, making it confusing to understand the actual purl structure to be respected.


iamwillbar:
There's discussion about this in #58 but the current stance is that any special characters in the name need to be encoded, otherwise there could be confusion about where the version portion (delimited by an `@`) begins.

PR #122, opened 2021-09-08
Provide mode detail on namespace segments

AidanDelaney:
The specification provides examples of namespace segments
but does not provide a grammar to describe the.  Adopt a
restriction on the segments presented in rfc3986 to describe
namespaces.


johnmhoran:
@AidanDelaney Thank you for this PR!  In light of @pombredanne 's comments about adding significant complexity and your reply, I am going to close the PR.  If your view changes, please feel free to open this as a new issue.

Issue #103, opened 2021-05-04
Define cocoapods as a purl type


stevespringett:
Currently, `cocoapods` is a reserved purl type but is currently undefined. This ticket is to track the definition of the purl type.

Initial thoughts:

cocoapods for Cocoapods:

* The default repository is https://cdn.cocoapods.org/

* The name is the pod name and is case sensitive.

* `arch` is the qualifiers key for a package architecture

Examples:
```
pkg:cocoapods/AFNetworking@4.0.1
pkg:cocoapods/MapsIndoors@3.24.0?arch=arm
pkg:cocoapods/macOSThemeKit@1.2.3?arch=x64
```

stevespringett:
From my understanding, there are two main repos:
* https://github.com/CocoaPods/Specs.git
* https://cdn.cocoapods.org/

Whichever one is the predominate repo with the most packages, and with more adoption, should likely be the default repo.

macblazer:
The two repo URLs listed above, https://github.com/CocoaPods/Specs.git and https://cdn.cocoapods.org/, are actually the same single package repo.  The CDN version is the preferred one; the direct GitHub link contains the same info.  They should essentially be treated as mirrors.

CocoaPods can define custom package repositories.  The purl `repository_url` qualifier should likely be used to indicate a custom package repository.

CocoaPods can also define local on-disk dependencies as well as direct-to-git sources that do not go through any package repo.  Perhaps the `download_url` qualifier should be used to indicate packages that are directly linked.  Local pods would use a `file://`type URL while direct pods would use a `https:` URL.  Examples:

```
# Example of a custom package repository
pkg:cocoapods/CustomPod@1.2.3?repository_url=https://custom_repo.example.com/path/to/repo/specs.git

# Example of a local on disk pod (could use relative or absolute pathing in the file:// URL)
pkg:cocoapods/LocalPod@1.2.3?download_url=file://relative/path/to/pod

# Example of direct pod access (no package repository used)
pkg:cocoapods/DirectPod@1.2.3?download_url=https://git_server.example.com/path/to/directpod.git
```

bradcupit:
using what @macblazer posted we'd have this example of a custom package repository (a.k.a. private specs repo -- though this one's public).

```
pkg:cocoapods/Aerodramus@2.0.0?repository_url=https://github.com/artsy/Specs.git
```

or that same package if referenced as a direct-to-git source:

```
[...]

PR #125, opened 2021-09-23
Implemented cocoapods type as defined in #103

stevespringett:


stevespringett:
Refer to #103 for what the community has defined to be accurate about the cocoapods type.

iamwillbar (approved):

Issue #124, opened 2021-09-16
Register purl pkg: scheme @ IANA


pombredanne:
This came up as a topic in a recent discussion.... I think that with the actual adoption, now is a good time!
I will draft the submission here and start the process of registering at IANA.

elear:
Yes please.  The registration template can be found in [RFC7595](https://www.rfc-editor.org/rfc/rfc7595.html#section-7.4) Section 7.4, but do read the rest of the document.  Today, anyone could conceivably request "pkg:" for other reasons.  If you need help, contact me.

pombredanne:
@elear I have been working on the registration a while back... I just pushed a PR for review as #141 
I will socialize this today with the IETF URI review list too.

@coderpatros ^

elear:
Fantastic!  I've just added purl support for my apt2sbom cyclonedx generation tool.  I'll do the same with spdx when I understand where they want that info.  Let me know if you want an extra pair of eyes.

pombredanne:
I just sent this email to the uri-review list
https://mailarchive.ietf.org/arch/msg/uri-review/6m26SxW5e1eVzzQvQ-04oIbnqUY/

You guys may want to chime in there too.

whiskeysierra:
(The IANA URI review highlights that it's not a URL, but a URI. But I guess it's to late for this project's name to be changed accordingly.)

pombredanne:
@whiskeysierra re:

> (The IANA URI review highlights that it's not a URL, but a URI. But I guess it's to late for this project's name to be changed accordingly.)

Thank you for stopping by!

I am not sure I read the same from the current IANA discussion, unless I am missing something... do you have a specific quote and pointer you could share here?

FWIW, there have been several debates on this topic in the past. Do you mind to have a look at https://github.com/package-url/purl-spec/issues/5 and https://github.com/package-url/purl-spec/issues/9 for some background?

whiskeysierra:
(Sorry for derailing this issue. Thanks a bunch for the links, those are insightful!)

elear:
Hmm.  Purl reminds us of treasures, while puri reminds us of tasty Indian cuisine ;-)  I'd suggest not fixating on the name...

Hritik14:
Just a note, after this is done we would want to have it on https://en.wikipedia.org/wiki/List_of_URI_schemes as well <3

iamwillbar:
@pombredanne is there anything I can do to help get this registration submitted?

[...]

PR #141, opened 2021-12-13
Add IANA pkg: registration draft

pombredanne:
This is a useful step for:
- #124


pombredanne:
@elear I definitely want an extra pair of eyes! @coderpatros ditto!

coderpatros:
Not significant for the registration, but Package URL support was added to Tern too https://github.com/tern-tools/tern

pombredanne:
From the feedback of the IANA review, I need to add a section on security consideration. I am a bit at loss there to explain what could be the security issues with a Package URL, but I will try  to come up with something.

elear:
I agree that this is tricky.  On the whole, purl is stitching together different namespaces, and most of the threats are around attacks on those namespaces.  For example, if someone hacks my account, they can update whatever stuff I can drop onto github or pypi etc.  If someone hacks Debian, then they can whack the dpkg repos.

If there are examples where that's not the case, maybe indicate that...

coderpatros:
@pombredanne I can work on the security considerations if you haven't already.

elear:
The security considerations should contain remediations without fluff like “dedicated teams...".  The point is that you are **naming** the package, and not telling people to blindly do updates.  npm is probably festering with all source of hidden malware.  Same with pypi.  Certification of the quality and provenance of the object you are referencing is beyond your scope.  It's as simple as that.

JimFuller-RedHat:
What is the status ? - checking https://www.iana.org/assignments/uri-schemes/uri-schemes.xhtml I see pkg: is not registered

elear:
Thanks for the ping.  Looking back on this thread, I agree that we could simplify the security considerations,.  So maybe something like this:

> The underlying security considerations involved with PURLs is that they continue to resolve to the same packages over time.  The key threat to this isn't in the PURL but in the repository/package system itself not being properly maintained.

Anyway, I think there's enough there for at least a provisional assignment, and this really should go out

elear:
One additional point- the POC and controller should be an institutional address of some form.

pombredanne:
@JimFuller-RedHat there is very little left to do, the major item was the security consideration. This should be a relatively easy one. 

@elear And we can update the addresses alright for the POC and institution, please elaborate on what you mean there.

elear:
@pombredanne Email addresses should not be to individuals, and preferably directly related to the project.  So... update Contact: and Change-Controller accordingly.

pkg-iana-registration.txt
```diff
@@ -0,0 +1,44 @@
...
+
+Scheme syntax: pkg:<type>/<namespace/name>@version?<qualifiers>#<subpath>
+
+Scheme semantics: Identify and locate a software package.
+
+Applications/protocols that use this scheme name:
+
+  Binary Analysis Next Generation (BANG): https://github.com/armijnhemel/binaryanalysis-ng
+  CycloneDX: https://github.com/CycloneDX
+  DefectDojo: https://github.com/DefectDojo/django-DefectDojo
```
tschmidtb51:
```suggestion
  Common Security Advisory Framework (CSAF) - ISO/IEC 20153:2025: https://docs.oasis-open.org/csaf/csaf/v2.0/csaf-v2.0.html
  DefectDojo: https://github.com/DefectDojo/django-DefectDojo
```

elear:
@tschmidtb51 this is good enough, although I would still prefer to see an institutional controller.  I would go ahead and submit so you can start to get comments back from the IANA reviewers.

elear (approved):

pkg-iana-registration.txt
tschmidtb51: Alternatively, link to the GitHub:
```suggestion
  Common Security Advisory Framework (CSAF) - ISO/IEC 20153:2025: https://github.com/oasis-tcs/csaf
  DefectDojo: https://github.com/DefectDojo/django-DefectDojo
```

tschmidtb51:
> @tschmidtb51 this is good enough, although I would still prefer to see an institutional controller. I would go ahead and submit so you can start to get comments back from the IANA reviewers.

Sorry - you lost me there...
Is that about the addition of CSAF or the line
```
Change controller: nexB Inc. <info&nexb.com>
```
?

elear:
Yes, it's about the "Change controller:" line having some institutional connection to the registry for pkg.  But I think this is changeable later.  Better to get the registration in...

mjherzog:
I need confirmation from Philippe but I expect that we want to change the control info to:
Contact: Philippe Ombredanne <pombredanne&aboutcode.org>
Change controller: AboutCode Europe ASBL. <info&aboutcode.org>

AboutCode Europe ASBL is a non-profit that has taken over stewardship of AboutCode projects from nexB.

elear:
In my view, this has become too drawn out.  Let's get some registration into IANA.  @pombredanne and @mjherzog are you guys ready for the IANA people to do their (re)review?

mjherzog:
@elear @pombredanne 
I updated the text to prepare for registration following the template from https://www.rfc-editor.org/info/rfc7595/#section-7.4 - specifically
- Changed _Contact_ and _Controller_ to aboutcode.org
- Removed the _Scheme syntax,_ _Scheme semantics_ and _Encoding considerations_ sections according to latest registration template - these are covered by ECMA-427
- Updated _References / Scheme specification_ to ECMA-427
- Updated _References / Scheme implementation_  with a URL for more applications/protocols listed on the packageurl.org website

I think that we are down to 2 open questions:
- ECMA-427 covers the _Scheme syntax_, _Scheme semantics_ and _Encoding considerations_ but not _Security considerations_. How should we cover that within the current template - Add a section for that under References since that is where the other scheme information is located?
- We have 2 sets of overlapping information about applications that use the scheme under:
   -  _Applications/protocols that use this scheme name_ - bullet list
   - _References / Scheme implementation_  - https://packageurl.org/docs/getting-started/toolgrid
  See https://github.com/aboutcode-org/www.aboutcode.org/issues/172 for the list of applications not on the website list.
  Do keep only one list?

elear:
@mjherzog with regard to Security Considerations, I don't have a strong recommendation, other than to say that the security of the name is the responsibility of the selected package manager system.

Under applications/protocols, I think a brief narrative statement is sufficient, along the lines of:

This scheme is intended to be used to identity and, as appropriate, invoke a particular package manager relating to a specific software package.

The most important issue is change controller.  Now that this is an ECMA standard, it might make sense to have someone at ECMA be the change controller.  I suspect that the expert reviewers will insist on that.

elear:
yup.  Let's get this back to the IANA.

mjherzog:
@elear I updated the change controller to Ecma TC54-TG2 on recommendation of our Ecma Intl contacts.
I just need to clean up the application/protocols section and it will be ready to submit

mjherzog:
@elear @pombredanne Please take a final look before I merge. 
I expect that the request should be submitted by @pombredanne because he is the listed Contact

elear:
Good enough.  I'm sure the IANA reviewers will request **some** form of change, so the sooner it is sent, the sooner you'll find out.

elear:
@pombredanne ok, now it's on you to resubmit an IANA request.  You're aiming for https://www.iana.org/form/protocol-assignment.

Issue #127, opened 2021-09-27
Decoupling Location from Identity - Is this in the scope of purl?


SteveLasker:
I'm opening this issue as a question, as the [readme](https://github.com/package-url/purl-spec/blob/master/README.rst) states purl is scoped to:

> A purl or package URL is an attempt to standardize existing approaches to reliably identify _**and locate**_ software packages.

While I recognize this has been a known pattern to assume a location for an artifact, this has also been a challenge for users that wish to take ownership of the content they depend upon. The realization that even common/shared/oss artifacts must be pulled from multiple locations, making an individual location a problematic concept.

## A detailed post, with the context of the problem

[Separating Identity From Location](https://stevelasker.blog/2021/09/24/separating-identity-from-location/)

## TLDR:

From an SBoM community (CycloneDX and SPDX as examples), there's a desire to assure a reference within an SBoM points to a very specific artifact. It could be a container image, helm chart, wasm or other types where SBoMs are relevant.
There are two dimensions to this decoupling:
1. Users want to take control of the public content they depend upon, and promote it to their private registry. In many/most cases, they want those private environments to be locked down, with no access to the internet. In this case, they need to import the container image, the sbom, scan results, signatures and other types with the image so they're all local to the private network/environment. Any egress is blocked by network restrictions.
2. The same content is becoming available on multiple registries. For instance, the exact same debian image can be pulled from [docker hub](https://hub.docker.com/_/debian) or [ecr public](https://gallery.ecr.aws/debian/debian), with others _"coming soon"_.

For 1, you might be willing to say "this is the debian image from docker.io", however, it's currently in my private registry. As long as the image is in the same repository as the SBoM, it can be resolved, and the URL part of the identifier is ignored as the debain image is said to be unique as it was in docker.io.  Mirrors could also be resolved, _maybe_.
For 2, it's far more challenging. If the exact same debian image is pushed to docker hub, ecr public, github, mcr and quay, what would the URL be?  Should the debian owner have to pick one?
Whether the user pulls the debian image from hub, ecr, or their private registry, the SBoM should be able to resolve the debian image, independently from where they got the image.
The proposal in #123, focuses on decoupling location from identity. Location is an optional hint in the oci-artifact purl PR.
What we've been trying to understand is whether purl, the specification, can decouple identity from location, or is purl always about identity **&** location?

[...]

PR #145, opened 2021-12-18
Proposal: Add signature_url as qualifier to generic purls

shibumi:
Many packages provide a signature. We should add a signature_url qualifier to the generic package url.

### DIFF
--- a/PURL-TYPES.rst
+++ b/PURL-TYPES.rst
@@ -178,7 +178,7 @@ generic
 for "upstream-from-distro" packages. In particular this is handy for a plain
 version control repository such as a bare git repo.
 
-- There is no default repository. A ``download_url`` and ``checksum`` may be
+- There is no default repository. A ``download_url``, ``signature_url` or ``checksum`` may be
   provided in `qualifiers` or as separate attributes outside of a ``purl`` for
   proper identification and location.
 - When possible another or a new purl ``type`` should be used instead of using
@@ -190,6 +190,8 @@ version control repository such as a bare git repo.
      pkg:generic/openssl@1.1.10g
      pkg:generic/openssl@1.1.10g?download_url=https://openssl.org/source/openssl-1.1.0g.tar.gz&checksum=sha256:de4d501267da
      pkg:generic/bitwarderl?vcs_url=git%2Bhttps://git.fsfe.org/dxtr/bitwarderl%40cc55108da32
+     pkg:generic/openssl@1.1.10g?download_url=https://openssl.org/source/openssl-1.1.0g.tar.gz&signature_url=https://openssl.org/source/openssl-1.1.0g.tar.gz.asc&checksum=sha256:de4d501267da
+
 
 
 github


stevespringett:
Thanks for the suggestion and PR. 
* In some ecosystems, the signature can automatically be derived from an existing purl (e.g. Maven)
* Some ecosystems support multiple signatures (also Maven)

Since log4j is on everyones mind this week, how would signatures of [log4j-core v2.17.0](https://repo1.maven.org/maven2/org/apache/logging/log4j/log4j-core/2.17.0/ ) be represented?

shibumi:
To be honest I was mostly interested in tarballs, because that's what I have to deal with at most. I am an Arch Linux package maintainer and we compile software from source, via downloading the source tarball from Github/Website/etc.

i see that log4j provides signatures as well:

https://repo1.maven.org/maven2/org/apache/logging/log4j/log4j-core/2.17.0/log4j-core-2.17.0-test-sources.jar
https://repo1.maven.org/maven2/org/apache/logging/log4j/log4j-core/2.17.0/log4j-core-2.17.0-test-sources.jar.asc

It's a fat jar (I guess), hence I would just use the generic package url for downloading the jar as well.

pombredanne:
@shibumi I have mixed feeling about overloading things with standard extra qualifiers and URLs and .... I wonder if this could be tried out first for arch only? and see if this works OK at scale?

shibumi:
@pombredanne I understand your concerns about overloading the purl with just another qualifier. Would you have another, maybe better, solution for providing a signature? Maybe a 'new' purl called `generic secure url`?

@dlorenc What do you think about this issue here? Do you see another solution for cosign + purls?

Rush2112hale:
Sorry to me so long I've been in the hospital for three months

Rush2112hale:
Thanks I've been sick for a while sorry it took so long

pombredanne:
@shibumi re:

> I understand your concerns about overloading the purl with just another qualifier. Would you have another, maybe better, solution for providing a signature? Maybe a 'new' purl called generic secure url?

I am not saying that using a qualifier is not OK, rather that it would make sense to experiment first.

IMHO you should experiment at some scale first how to evaluate how this works for you (say for Arch Linux) with working code and then based on this, report back here?

shibumi:
> IMHO you should experiment at some scale first how to evaluate how this works for you (say for Arch Linux) with working code and then based on this, report back here?

We are months + lot's of community discussions away for testing this on scale for Arch Linux. Arch Linux was just an example.
Having signatures on the generic url specification is a good use case for every distribution and tools that download generic tarballs or any other artifact.

SteveLasker:
Is this asking to encode a fixed url/location for a signature?

shibumi:
@SteveLasker yes, correct. Having a fixed location for a detached signature is common practice in software distribution. 
The signature is usually computed over the archive of the source code or the checksum file.
Just to mention a few projects:

* glibc: https://ftp.gnu.org/gnu/glibc/
* gnupg: https://gnupg.org/ftp/gcrypt/
* gopass: https://github.com/gopasspw/gopass/releases/tag/v1.13.1
* Python: https://www.python.org/ftp/python/3.9.9/
* Golang: https://storage.googleapis.com/golang

I am sure I can find many other big, relevant projects. It's common practice to use detached signatures, hence I don't understand why this need testing to be honest.

pombredanne:
@shibumi re https://github.com/package-url/purl-spec/pull/145#issuecomment-1040037809

>  I am sure I can find many other big, relevant projects. It's common practice to use detached signatures, hence I don't understand why this need testing to be honest.

Yes! this is common enough: maven would be an example with million packages with a detached .asc signature. As in https://repo1.maven.org/maven2/org/glassfish/javax.json/1.1.4/ . In his case though the signature is IMHO something that's entirely derivable from the package type and the purl. 

The reason why I am taking some time is that each addition to a simple spec is making it a bit less simple, one tiny addition at a time... so it is always best not to rush things IMHO! I hope you can appreciate the motivation.

SteveLasker:
I'm totally supportive of detached signatures. It's the basis for [Notary v2](https://github.com/notaryproject/notaryproject) and detached references for anything with [ORAS Artifacts](https://github.com/oras-project/artifacts-spec/blob/main/scenarios.md). 
I was just clarifying the intent, before raising the concern with having the signature URL as a coupling of "identity and location". 
See #127 for a discussion, and this post: [Separating Identity From Location](https://stevelasker.blog/2021/09/24/separating-identity-from-location/)

When a user copies the artifact (package, container, wasm, sbom, scan result) to their private environment, [following best practices for consuming public content](https://opencontainers.org/posts/blog/2020-10-30-consuming-public-content/), is the user expected to reach that signature location from within their private network?

shibumi:
@pombredanne  yes your motivation is clear. No worries :)

@steveLasker normally you would download the artifact *and* the signature together from a public location and then verify the signature. In [your particular case](https://opencontainers.org/posts/blog/2020-10-30-consuming-public-content/) the signature is in the OCI image embedded (this is common practice as far as I know). This PR here, is mainly about generic artifacts (software archives, arbitrary files, etc). For downloading OCI Images I would use a different package URL to be honest.

SteveLasker:
Thanks @shibumi. Can this be clarified as a hint, as opposed to a definitive location?
In the promotion scenario, I'm hoping we help folks enable the artifact and the signature (even if from another location) can be promoted into the destination environment. 

This is the permanent location, vs. runtime discussion. Where is the signature now? Oh, there's a hint that's it's at storage.example.io. When I copy the package, and the signature into the private environment, the signature location is different.

The user wants to promote an artifact from a public location to private:
- Before pulling the artifact, verify the signature from storage.example.io
- Validation complete
- Copy the artifact to artifacts.acme-rockets.io
- Copy the signature to storage.acme-rockets.io

Copies from the internal staging location to a production environment that has a different vnet.
- Before pulling the artifact, verify the signature. 
- Validation complete
- Copy the artifact to prod-artifacts.acme-rockets.io
- Copy the signature to prod-storage.acme-rockets.io

PR #150, opened 2022-02-11
docs(types): oci type qualifiers must be percent encoded

masahiro331:
This PR modified qualifiers values like url.

Some url-like values ​​in qualifiers have been URL-encoded.
* vcs_url
* repository_url
* download_url

---

[OCI specification](https://github.com/package-url/purl-spec/blob/master/PURL-TYPES.rst#oci)

> repository_url: A repository URL where the artifact may be found, but not intended as the only location. This value is encouraged to identify a location the content may be fetched

```
pkg:oci/static@sha256:<digest>?repository_url=gcr.io/distroless/static&tag=latest
```

[qualifiers specification](https://github.com/package-url/purl-spec/blob/master/PURL-SPECIFICATION.rst#rules-for-each-purl-component) 
> A value must be a percent-encoded string

Must repository_url be percent-encoded?

### DIFF
--- a/tests/types/generic-test.json
+++ b/tests/types/generic-test.json
@@ -46,7 +46,7 @@
       "description": "Parse test for PURL type: generic",
       "test_group": "base",
       "test_type": "parse",
-      "input": "pkg:generic/openssl@1.1.10g?download_url=https://openssl.org/source/openssl-1.1.0g.tar.gz&checksum=sha256:de4d501267da",
+      "input": "pkg:generic/openssl@1.1.10g?download_url=https:%2F%2Fopenssl.org%2Fsource%2Fopenssl-1.1.0g.tar.gz&checksum=sha256:de4d501267da",
       "expected_output": {
         "type": "generic",
         "namespace": null,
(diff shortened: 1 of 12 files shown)


stevespringett:
@pombredanne  The test suite, in this spec repo and used by all implementations, do not have repository_url encoded.

johnmhoran:
Hi @masahiro331 and thank you for your PR.  I see that although most of the proposed changed PURLs are for `qualifiers` values, one is for a `subpath` value.  My sense is that the '/' in a qualifier value must be percent-encoded, but the '/' in a subpath value must not be encoded.  To enable us to tackle these 2 PURL components individually, could you please replace this PR 150 with two separate PRs?

PURL-TYPES.rst
```diff
@@ -248,8 +248,8 @@ hex
 
       pkg:hex/jason@1.1.2
       pkg:hex/acme/foo@2.3.
-      pkg:hex/phoenix_html@2.13.3#priv/static/phoenix_html.js
-      pkg:hex/bar@1.2.3?repository_url=https://myrepo.example.com
+      pkg:hex/phoenix_html@2.13.3#priv%2Fstatic%2Fphoenix_html.js
```
masahiro331, line 251:
```suggestion
      pkg:hex/phoenix_html@2.13.3#priv/static/phoenix_html.js
```

mjherzog:
We will actually need a new PR covering the updates to types/:
- generic-definition.json
- hex-definition.json
- npm-definition.json
- oci-definition.json

masahiro331:
Some url-like values ​​in qualifiers have been URL-encoded.
* vcs_url
* repository_url
* download_url

Issue #151, opened 2022-02-11
About Epoch in RPM and DEB Package Versioning


masahiro331:
# About Epoch in RPM and DEB Package Versioning

In [DEB](https://github.com/package-url/purl-spec/blob/master/PURL-TYPES.rst#deb) Type, the epoch is included in the version.
```
pkg:deb/debian/attr@1:2.4.47-2%2Bb1?arch=amd64
```

But, [RPM](https://github.com/package-url/purl-spec/blob/master/PURL-TYPES.rst#rpm) Type, the epoch is included in qualifiers.
```
pkg:rpm/centerim@4.22.10-1.el6?arch=i686&epoch=1&distro=fedora-25
```

In my opinion, it is preferable to include epoch in the version as in RPM.

johnmhoran:
Hello @masahiro331 .  Thank you for opening this issue.  I see that you opened a related PR on 2022-02-19 and closed it on 2023-03-30.  Does that mean this issue is ready to be closed as well?

PR #153, opened 2022-02-19
docs(types): rpm version combined epoch number

masahiro331:
issue: #151


dralley (approved):
This change mostly makes sense to me.  If epoch is not present for an RPM package than it is treated as zero.

dralley:
However, here is the discussion from when it was originally done this way - there are decent arguments there as well.

https://github.com/package-url/purl-spec/issues/69#issuecomment-551355102

PR #156, opened 2022-03-04
Add support for pub packages

ericlarssen-wf:
Add documentation and test data for pub packages.

### DIFF
--- a/PURL-TYPES.rst
+++ b/PURL-TYPES.rst
@@ -329,16 +329,20 @@ including container images built by Docker and others:
       pkg:oci/hello-wasm@sha256:<digest>?tag=v1
 
 pub
-----
-``pub`` for Dart and Flutter packages:
+---
+``pub`` for Dartlang packages:
 
-- The default repository is ``https://pub.dartlang.org``
-- Pub normalizes all package names to be lowercase and using underscores. The only allowed characters are `[a-z0-9_]`. 
+- Pub normalizes all package names to be lowercase and using underscores. The only allowed characters are `[a-z0-9_]`.
+- Optional qualifiers may include:
+  - ``hosted_url``: A URL representing the registry where the package is stored.
+  - ``vcs_url``: A URL representing the git repository information where the package is stored.
+  - Only one of ``hosted_url`` and ``vcs_url`` should be provided.
 - More information on pub naming and versioning is available in the [pubspec documentation](https://dart.dev/tools/pub/pubspec)
 - Examples::
 
-      pkg:pub/characters@1.2.0
-      pkg:pub/flutter@0.0.0
+      pkg:pub/foobar@1.2.34
+      pkg:pub/foobar-hosted@1.2.34?hosted_url=https://pub.hosted.org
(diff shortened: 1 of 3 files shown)


pombredanne (changes requested):
Oops... I merged another pub-related PR before seeing this one. Sorry for this!

But yours has good additions...
Do you mind to rebase this on the latest, adding your changes?
Also can you elaborate on the "hosted_url" attribute a bit? Is  this something defined in pub/dart upstream? is there a risk of redundancy with "vcs_url" ?

pombredanne:
After the merge of PR, PURL types are now defined in JSON:
- See  #514

PURL-TYPES.rst
```diff
@@ -329,16 +329,20 @@ including container images built by Docker and others:
...
+  - ``vcs_url``: A URL representing the git repository information where the package is stored.
+  - Only one of ``hosted_url`` and ``vcs_url`` should be provided.
 - More information on pub naming and versioning is available in the [pubspec documentation](https://dart.dev/tools/pub/pubspec)
 - Examples::
 
-      pkg:pub/characters@1.2.0
-      pkg:pub/flutter@0.0.0
+      pkg:pub/foobar@1.2.34
+      pkg:pub/foobar-hosted@1.2.34?hosted_url=https://pub.hosted.org
+      pkg:pub/fooboar-git@1.2.34?vcs_url=git://github.com/path/to/repo.git%404345abcd34343#path/to/package
```
ppkarwasz, line 345: Why are these examples using `-` in package names instead of `_`?

test-suite-data.json
```diff
@@ -466,5 +466,65 @@
...
+    "name": null,
+    "version": null,
+    "qualifiers": null,
+    "subpath": null,
+    "is_invalid": true
+  },
+  {
+    "description": "valid base pub purl",
+    "purl": "pkg:pub",
+    "canonical_purl": "pkg:pub/utils@0.1.0",
```
ppkarwasz, line 485:
```suggestion
    "description": "valid base pub purl",
    "purl": "pkg:pub/utils@0.1.0",
    "canonical_purl": "pkg:pub/utils@0.1.0",
```

`purl` should be a valid PURL that parsers canonicalize to obtain `canonical_purl`.

test-suite-data.json
```diff
@@ -466,5 +466,65 @@
...
+    "name": "utils",
+    "version": "0.1.0",
+    "qualifiers": null,
+    "subpath": null,
+    "is_invalid": false
+  },
+  {
+    "description": "valid hosted pub purl",
+    "purl": "pkg:pub",
+    "canonical_purl": "pkg:pub/utils@0.1.0?hosted_url=https://pub.hosted.org",
```
ppkarwasz, line 497: This and the examples below should be updated to the current percent-encoding rules, i.e. `/` must be escaped as `%2F` in the value of the qualifiers in the `canonical_purl`.

ppkarwasz (changes requested):

PR #168, opened 2022-05-19
elaborate conan purl type

jsteinhofff:
Include differentiation between different binary packages, revisions, and have user and channel consistently as qualifier.

### DIFF
--- a/PURL-TYPES.rst
+++ b/PURL-TYPES.rst
@@ -79,17 +79,26 @@ composer
 
 conan
 -----
-``conan`` for Conan C/C++ packages:
+``conan`` for Conan C/C++ packages. The purl is designed to closely resemble the Conan-native `<package-name>/<package-version>@<user>/<channel>` `syntax for package references <https://docs.conan.io/en/1.46/cheatsheet.html#package-terminology>`_.
 
-- The default repository is ``https://center.conan.io``
-- The ``namespace`` is the user if present
-- The ``name`` is the package name.
-- The ``version`` is the package version.
-- The qualifier ``channel`` must be not empty if namespace is present
-- Examples::
+- ``name``: The Conan ``<package-name>``.
+- ``version``: The Conan ``<package-version>``.
+- ``namespace``: The vendor of the package.
+- Qualifier ``user``: The Conan ``<user>``. Only required if the Conan package was published with ``<user>``.
+- Qualifier ``channel``: The Conan ``<channel>``. Only required if the Conan package was published with Conan ``<channel>``.
+- Qualifier ``rrev``: The Conan recipe revision (optional). If omitted, the purl refers to the latest recipe revision available for the given version.
+- Qualifier ``prev``: The Conan package revision (optional). If omitted, the purl refers to the latest package revision available for the given version and recipe revision.
+- Qualifier ``repository_url``: The Conan repository where the package is available (optional). If ommitted, ``https://center.conan.io`` as default repository is assumed.
 
-      pkg:conan/cctz@2.3
-      pkg:conan/bincrafters/cctz@2.3?channel=stable
+Additional qualifiers can be used to distinguish Conan packages with different settings or options, e.g. ``os=Linux``, ``build_type=Debug`` or ``shared=True``.
+
+If no additional qualifiers are used to distinguish Conan packages build with different settings or options, then the purl is ambiguous and it is up to the user to work out which package is being referred to (e.g. with context information).
+
+Examples::
+
+      pkg:conan/openssl@3.0.3
+      pkg:conan/openssl.org/openssl@3.0.3?user=bincrafters&channel=stable
+      pkg:conan/openssl.org/openssl@3.0.3?arch=x86_64&build_type=Debug&compiler=Visual%20Studio&compiler.runtime=MDd&compiler.version=16&os=Windows&shared=True&rrev=93a82349c31917d2d674d22065c7a9ef9f380c8e&prev=b429db8a0e324114c25ec387bfd8281f330d7c5c
 
 conda
 -----


PURL-TYPES.rst
```diff
@@ -81,15 +81,24 @@ conan
...
 ``conan`` for Conan C/C++ packages:
 
-- The default repository is ``https://center.conan.io``
-- The ``namespace`` is the user if present
-- The ``name`` is the package name.
-- The ``version`` is the package version.
-- The qualifier ``channel`` must be not empty if namespace is present
+- The default repository is ``https://conan.io/center``.
+- The ``namespace`` is the vendor.
+- The ``name`` is the conan package name.
```
sschuberth, line 86: Please capitalize "Conan" in prose text. Same below.

PURL-TYPES.rst
```diff
@@ -81,15 +81,24 @@ conan
...
+- The ``version`` is the conan package version.
+- The qualifier ``channel`` is optionally the conan channel.
+- The qualifier ``user`` is optionally the conan user.
+- The qualifier ``rrev`` is optionally the conan recipe revision. If none is given, the latest recipe revision available for the given version is meant.
```
sschuberth, line 90: "If none is given" could be interpreted as literally passing "rrev=none". Maybe better say "If omitted, ...".

Also, maybe better "is used" instead of "is meant"?

Similar below.

PURL-TYPES.rst
```diff
@@ -81,15 +81,24 @@ conan
...
+- The qualifier ``prev`` is optionally the conan package revision. If none is given, the latest package revision for the given version and recipe revision is meant.
+- Additional qualifiers can optionally be used to distinguish conan packages with different
+  settings or options, eg. os=Linux, build_type=Debug or shared=True.
+- Qualifier values shall be urlencoded where necessary.
+- If no additional qualifiers are used to distinguish conan packages build with different settings or options, then the
+  purl is ambiguous and it is up to the user to work out which package is meant (eg. with context information).
```
sschuberth, line 96: meant -> being referred to

PURL-TYPES.rst
```diff
@@ -81,15 +81,24 @@ conan
...
 - Examples::
 
-      pkg:conan/cctz@2.3
-      pkg:conan/bincrafters/cctz@2.3?channel=stable
+      pkg:conan/openssl@3.0.3
+      pkg:conan/openssl.org/openssl@3.0.3?user=bincrafters&channel=stable
+      pkg:conan/openssl.org/openssl@3.0.3?arch=x86_64&build_type=Debug&compiler=Visual%20Studio&compiler.runtime=MDd&compiler.version=16&os=Windows&shared=True&rrev=93a82349c31917d2d674d22065c7a9ef9f380c8e&pref=b429db8a0e324114c25ec387bfd8281f330d7c5c
```
sschuberth, line 101: Typo in "pref".

PURL-TYPES.rst
```diff
@@ -81,15 +81,24 @@ conan
 -----
 ``conan`` for Conan C/C++ packages:
 
-- The default repository is ``https://center.conan.io``
-- The ``namespace`` is the user if present
-- The ``name`` is the package name.
-- The ``version`` is the package version.
-- The qualifier ``channel`` must be not empty if namespace is present
+- The default repository is ``https://conan.io/center``.
```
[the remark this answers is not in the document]
jsteinhofff: good point, i will fix that

PURL-TYPES.rst
jsteinhofff: about None: good point
about "is used" vs. "is meant" i am not so sure. The case where no rrev is specified but multiple revisions exist is a bit strange because in the end the user would need to look up what the newest one is. "is used" implies that some implementation exists which would just do that, but this is not the case.
So what i want to convey here that writing a PURL without rrev/prev when multiple revisions exists shall be interpreted as "latest revision to be determined by the user".

sschuberth: Where "latest revision to be determined by the user" could be *any* of the available revisions, and not necessarily the one with the latest date or so? Then that's rather arbitrary, and maybe we should simply drop the second sentence completely, implying that it's undefined (similar to how we also don't explicitly say what omitting any of the other optional qualifiers means).

jsteinhofff: Conan behaves the way that if you dont specificy prev or rrev, then it will fetch the latest one from the server. I think the same behaviour we should also strive for. Just here in the PURL-spec document we dont have a server and we dont have an implemented tool, so it is just an intention we can formulate.

sschuberth: > I think the same behaviour we should also strive for.

Agreed, but I'm still wondering why we're documenting the implicit default behavior here, but not for the other optional qualifiers? I.e. which channel is used if none is specified?

jsteinhofff: Hmm, @sschuberth user and channel are also optional in Conan, so i think having them optional here as well kind of makes sense. Shall we define that they shall only be omitted if also on the conan package side they were omitted?

sschuberth: > Shall we define that they shall only be omitted if also on the conan package side they were omitted?

With my Conan-half-knowledge, I'd say that makes sense, yes.

PURL-TYPES.rst
```diff
@@ -81,15 +81,24 @@ conan
...
-- The default repository is ``https://center.conan.io``
-- The ``namespace`` is the user if present
-- The ``name`` is the package name.
-- The ``version`` is the package version.
-- The qualifier ``channel`` must be not empty if namespace is present
+- The default repository is ``https://center.conan.io``.
+- The ``namespace`` is the vendor.
+- The ``name`` is the Conan package name.
+- The ``version`` is the Conan package version.
+- The qualifier ``channel`` is the Conan channel, it is required if the Conan package was published with channel.
```
sschuberth, line 88: "with channel" -> "with *a* channel" / "to a channel"?

PURL-TYPES.rst
```diff
@@ -81,15 +81,24 @@ conan
...
+- The qualifier ``user`` is the Conan user, it is required if the Conan package was published with user.
```
sschuberth, line 89: "with user" -> "with *a* user" / "as a user" / "under a user"?

PURL-TYPES.rst
```diff
@@ -81,15 +81,24 @@ conan
...
+- The qualifier ``rrev`` is optionally the Conan recipe revision. If ommitted, the purl refers to the latest recipe revision available for the given version.
```
sschuberth, line 90: I think the original "optional Conan recipe revision" read better. Similar below.

PURL-TYPES.rst
```diff
@@ -81,15 +81,24 @@ conan
...
+- The qualifier ``prev`` is optionally the Conan package revision. If ommitted, the purl refers to the latest package revision available for the given version and recipe revision.
+- Additional qualifiers can optionally be used to distinguish Conan packages with different
+  settings or options, eg. os=Linux, build_type=Debug or shared=True.
+- Qualifier values shall be urlencoded where necessary.
+- If no additional qualifiers are used to distinguish Conan packages build with different settings or options, then the
+  purl is ambiguous and it is up to the user to work out which package is beeing referred to (eg. with context information).
```
sschuberth, line 96: "beeing" -> "being"
"eg." -> "e.g."

sschuberth (changes requested):

PURL-TYPES.rst
```diff
@@ -81,15 +81,24 @@ conan
 -----
 ``conan`` for Conan C/C++ packages:
 
-- The default repository is ``https://center.conan.io``
-- The ``namespace`` is the user if present
-- The ``name`` is the package name.
-- The ``version`` is the package version.
-- The qualifier ``channel`` must be not empty if namespace is present
+- The default repository is ``https://center.conan.io``.
```
sschuberth, line 84: BTW @jsteinhofff, how is a non-default repository declared via PURL?

jsteinhofff: Good point, i dont know, should it? I think PURL is actually rather a misnomer, it should be PURI since the location of the packages are at not in it. At least when i look at the examples in the core specification the repository is nowhere given.

I see no way to indicated the repository in the PURL, so mentioning the default is maybe misleading (but it would be also for other package types here in the document). @sschuberth  what do you think?

sschuberth: Well, there *are* [examples](https://github.com/package-url/purl-spec#some-purl-examples) for how to specify a non-default Maven repository by leveraging an optional `repository_url` qualifier. I'd simply propose to use the same approach for Conan and document it here.

PURL-TYPES.rst
jsteinhofff: How about

> Qualifier ``rrev``: Conan recipe revision (optional), if omitted, the purl refers to the latest recipe revision available for the given version.

I would change this then consistently for all the points.

sschuberth: Ok with me if you start a new sentence with "If omitted, ...".

PURL-TYPES.rst
sschuberth: Yes, I think the fact that these example qualifiers are not URL-encoded is simply an oversight.

PURL-TYPES.rst
jsteinhofff: i could not come up with a clear formulation for this, so now i have a new approach where i introduce the Conan package reference scheme so i can refer to the tokens in it. Does that work for you?

PURL-TYPES.rst
jsteinhofff: same as for channel

PURL-TYPES.rst
jsteinhofff: i now added a reference to the repository_url. Actually i restructured it so the default repository is mentioned for the repository_url qualifier.

PURL-TYPES.rst
jsteinhofff: The structure differs now from the other package types, but i think it is more clear and consistent in itself now.

PURL-TYPES.rst
```diff
@@ -79,17 +79,28 @@ composer
...
-- The ``name`` is the package name.
-- The ``version`` is the package version.
-- The qualifier ``channel`` must be not empty if namespace is present
-- Examples::
+    <package-name>/<package-version>@<user>/<channel>
+
+- ``name``: The Conan ``<package-name>``.
+- ``version``: The Conan ``<package-version>``.
+- ``namespace``: The vendor of the package.
+- Qualifier ``user``: The Conan ``<user>``. It is required if the Conan package was published with ``<user>``.
```
sschuberth, line 89: Just say "Only required if ...". Same below.

PURL-TYPES.rst
```diff
@@ -79,17 +79,28 @@ composer
...
+- Qualifier ``channel``: The Conan ``<channel>``. It is required if the Conan package was published with Conan ``<channel>``.
+- Qualifier ``rrev``: The Conan recipe revision (optional). If ommitted, the purl refers to the latest recipe revision available for the given version.
```
sschuberth, line 91: "omitted" (one "m"). Same below.

PURL-TYPES.rst
```diff
@@ -79,17 +79,28 @@ composer
 
 conan
 -----
-``conan`` for Conan C/C++ packages:
+``conan`` for Conan C/C++ packages. Most information is taken from the respective Conan package reference which has the following format::
```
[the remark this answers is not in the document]
sschuberth: And maybe link "syntax for package references" to Conan docs that define that syntax.

sschuberth (changes requested):

sschuberth:
Some additional commit message nits:

The title "elaborated conan purl type" should probably say "Elaborate on the Conan purl type". Hard-wrapping the commit message body at column 75 would be nice, as well as adding a `Signed-off-by` footer (`git commit -s`).

PURL-TYPES.rst
jsteinhofff: done

PURL-TYPES.rst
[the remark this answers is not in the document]
jsteinhofff: done

PURL-TYPES.rst
maxhbr: Just to cross reference relevant discussion: https://github.com/package-url/purl-spec/pull/184

I think rfc2396.txt is not applicable and https://datatracker.ietf.org/doc/html/rfc3986#section-3.4 states something different:
> However, as query components are often used to carry identifying information in the form of "key=value" pairs and one frequently used value is a reference to another URI, it is sometimes better for usability to avoid percent-encoding those characters.

sschuberth (approved):

PURL-TYPES.rst
```diff
@@ -79,17 +79,26 @@ composer
...
+
+If no additional qualifiers are used to distinguish Conan packages build with different settings or options, then the purl is ambiguous and it is up to the user to work out which package is being referred to (e.g. with context information).
+
+Examples::
 
-      pkg:conan/cctz@2.3
-      pkg:conan/bincrafters/cctz@2.3?channel=stable
+      pkg:conan/openssl@3.0.3
+      pkg:conan/openssl.org/openssl@3.0.3?user=bincrafters&channel=stable
+      pkg:conan/openssl.org/openssl@3.0.3?arch=x86_64&build_type=Debug&compiler=Visual%20Studio&compiler.runtime=MDd&compiler.version=16&os=Windows&shared=True&rrev=93a82349c31917d2d674d22065c7a9ef9f380c8e&prev=b429db8a0e324114c25ec387bfd8281f330d7c5c
```
jdalton, line 102: The spec unit tests for this were not updated when the rules where changed here. The old way required channel to be present if qualifiers exist.

PR #170, opened 2022-05-27
Add `filename` qualifier to pypi.

### LINKED ISSUE #90: Clarification on how to refer to platform-specific pypi packages
A Python package sometimes differs in content/dependencies depending on things such as the targeted operating system, version of the Python interpreter (2 vs 3, for instance) and depending on which "extras" (optional features) of the package the consumer is interested in. This effectively results in different "editions" of the same package.

I assume that purl qualifiers would be used to represent these different editions, correct? 
For example:

    pkg:pypi/requests@2.22.0?extras=security,socks
    
    pkg:pypi/Jinja2@2.9.0?python_version=2.7

and similar. The [README](README.rst) doesn't elaborate on any special cases like these for `pypi` purls, whereas for example the description of `maven` purls contain quite a few examples with qualifiers.

So I guess I'd like to know if qualifiers are the natural construct to represent these python package "variations" and it would also be interesting to learn of any example use in the wild of `pypi` purls?

Edit: this is not necessarily a request to update the documentation. Consider it a question.

MarkLodato:
This allows the PURL to specify the exact file that was used.

Fixes #90.

### DIFF
--- a/PURL-TYPES.rst
+++ b/PURL-TYPES.rst
@@ -461,9 +461,11 @@ pypi
 - PyPI treats ``-`` and ``_`` as the same character and is not case sensitive.
   Therefore a PyPI package ``name`` must be lowercased and underscore ``_``
   replaced with a dash ``-``.
+- The ``filename`` qualifier selects a particular distribution file (case-sensitive).
 - Examples::
 
       pkg:pypi/django@1.11.1
+      pkg:pypi/django@1.11.1?filename=Django-1.11.1-py2.py3-none-any.whl
       pkg:pypi/django-allauth@12.23
 
 rpm
--- a/PURL-TYPES.rst
+++ b/PURL-TYPES.rst
@@ -519,10 +519,15 @@ pypi
 - PyPI treats ``-`` and ``_`` as the same character and is not case sensitive.
   Therefore a PyPI package ``name`` must be lowercased and underscore ``_``
   replaced with a dash ``-``.
-- The ``filename`` qualifier selects a particular distribution file (case-sensitive).
+- The ``file_name`` qualifier selects a particular distribution file
+  (case-sensitive). For naming convention, see the Python Packaging User Guide on
+  `source distributions <https://packaging.python.org/en/latest/specifications/source-distribution-format/#source-distribution-file-name>`_,
+  `binary distributions <https://packaging.python.org/en/latest/specifications/binary-distribution-format/#file-name-convention>`_,
+  and `platform compatibility tags <https://packaging.python.org/en/latest/specifications/platform-compatibility-tags/>`_.
 - Examples::
 
       pkg:pypi/django@1.11.1
+      pkg:pypi/django@1.11.1?filename=Django-1.11.1.tar.gz
       pkg:pypi/django@1.11.1?filename=Django-1.11.1-py2.py3-none-any.whl
       pkg:pypi/django-allauth@12.23
 


PURL-TYPES.rst
```diff
@@ -348,9 +348,12 @@ pypi
 - PyPi treats ``-`` and ``_`` as the same character and is not case sensitive.
   Therefore a Pypi package ``name`` must be lowercased and underscore ``_``
   replaced with a dash ``-``
+- The ``format`` qualifier specifies the type of distribution; one of ``sdist``, ``wheel``, or ``egg``.
```
iamwillbar, line 351: Should consumers assume one of these values if it is present?

MarkLodato: Now that you mention it, I'm doubting that `format` + `tag` is the right choice.

- I can't think of a situation where you'd want to specify one but not the other (except for sdist, where there is no tag).
- The url is still ambiguous with respect to file extension:
  - Source distributions support zip and tar.gz, and formerly tgz, tar, tar.Z, etc.
  - Binary distributions support egg and exe, possibly others. Example: [zope 4.0.2](https://pypi.org/project/zope.interface/4.0.2/#files), which has both .egg and .exe.

Instead, what about a single qualifier for the full tag + extension suffix? (Could use a better name than `suffix`.) Examples:

- `pkg:pypi/zope.interface@4.0.2?suffix=4.0.2.win-amd64-py2.7.exe`
- `pkg:pypi/django@1.11.1?suffix=py3-none-any.whl`

@di thoughts? As a background, PURL is a universal URL format for all package types, and I'm trying to add a query parameter to narrow scope to a specific artifact for packages on pypi.

di: > Should consumers assume one of these values if it is present?

No, none of these are guaranteed to be present

> Now that you mention it, I'm doubting that format + tag is the right choice.

Why not just use the filename? It's a) guaranteed to be unique and b) contains all possible 'qualifiers' you're adding here and c) does it in a format that's standardized in the Python community.

e.g.
* `zope.interface-4.0.2.win-amd64-py2.7.exe`
* `Django-1.11.1-py2.py3-none-any.whl`

MarkLodato: [Sorry I let this drop. Picking this up again...]

I was trying to de-duplicate with the package name. In other words, it's not super desirable to say `pkg:pypi/zope.interface@4.0.2?suffix=zope.interface-4.0.2.win-amd64-py2.7.exe`. See how `zope.interface` and `4.0.2` is listed twice? Is there a good way to avoid that?

sethmlarson: I'm in agreement to use the filename, that would also future-proof the implementation against new wheel platforms/tags.

MarkLodato: Switched to `filename`.

PURL-TYPES.rst
```diff
@@ -461,9 +461,11 @@ pypi
 - PyPI treats ``-`` and ``_`` as the same character and is not case sensitive.
   Therefore a PyPI package ``name`` must be lowercased and underscore ``_``
   replaced with a dash ``-``.
+- The ``filename`` qualifier selects a particular distribution file (case-sensitive).
```
sethmlarson, line 464: Using the name instead of individual tags is much better, thank you! Because naming is hard, I figured I'd throw `dist` out there as an option ;) Either way I'm happy with this.

di: IMO `filename` is better, `dist` feels a bit like jargon.

di (approved):

PURL-TYPES.rst
MarkLodato: Agreed. In fact, I struggled for what to call it in the text. Actually, should I say "distribution package file" to be unambiguous? (I want to clarify that it's not a file *within* an archive, which would instead be the `#path` part of the URL.)

sethmlarson: I agree too, let's go with `filename` :shipit:

sethmlarson (approved):

MarkLodato:
@iamwillbar friendly ping :) I think this is ready to be merged if you agree with the changes.

MarkLodato:
Friendly ping. Any chance this can get merged?

stevespringett:
@pombredanne there are likely other purl types that could benefit from the `filename` qualifier. Is this something we want to include as a global qualifier in the next purl spec?

MarkLodato:
@pombredanne Thanks for the suggestions! Please take a look at the latest version.

PR #174, opened 2022-06-10
Initial add of swid as a purl type

### LINKED ISSUE #173: Add swid purl type
This proposal adds SWID as a purl type.  [ISO-IEC 19770-2 Software Identification (SWID) Tags](https://www.iso.org/standard/65666.html) is a standard used to identify software and the contents of the software. It is an XML specification used primarily for commercial software for the purpose of license entitlement and software asset management. 

For the purpose of software identity, there are only a few important pieces of data that are necessary, which can easily be mapped to purl.

stevespringett:
Resolves #173 

Signed-off-by: Steve Springett <steve@springett.us>

### DIFF
--- a/PURL-TYPES.rst
+++ b/PURL-TYPES.rst
@@ -382,19 +382,19 @@ swid
 - The ``namespace`` is the optional name and regid of the entity with a role of softwareCreator. If specified, name is required and is the first segment in the namespace. If regid is known, it must be specified as the second segment in the namespace. A maximum of two segments are supported.
 - The ``name`` is the name as defined in the SWID SoftwareIdentity element
 - The ``version`` is the version as defined in the SWID SoftwareIdentity element
-- The qualifier ``tagId`` must not be empty and corresponds to the tagId as defined in the SWID SoftwareIdentity element. Per the SWID specification, GUIDs are recommended. If a GUID is used, it must be lowercase. If a GUID is not used, the tagId is case aware but not case sensitive
-- The qualifier ``tagVersion`` is an optional integer and corresponds to the tagVersion as defined in the SWID SoftwareIdentity element. If not specified, defaults to 0
+- The qualifier ``tag_id`` must not be empty and corresponds to the tagId as defined in the SWID SoftwareIdentity element. Per the SWID specification, GUIDs are recommended. If a GUID is used, it must be lowercase. If a GUID is not used, the tag_id qualifier is case aware but not case sensitive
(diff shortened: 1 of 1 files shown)


stevespringett:
cc: @dkruszew

stevespringett:
From the [XSD](https://standards.iso.org/iso/19770/-2/2015-current/schema.xsd) of the ISO spec:

tagId:
> tagId shall be a globally unique identifier and should be assigned a GUID reference (see ISO/IEC 19770-5 definition for GUID). The tagID provides a unique reference for the specific product, version, edition, revision, etc (essentially, the same binary distribution). If two tagIDs match and the tagCreator is the same, the underlying products they represent are expected to be exactly the same. This allows IT systems to identify if a software item (for example, a patch) is installed simply by referencing the specific tagID value which is likely to be readily available in a software inventory. It is recommended, when possible, that a 16 byte GUID be used for this field as this provides global uniqueness without a significant amount of overhead for space. If use of a 16 byte GUID is not possible, a text based globally unique ID may be constructed, this ID should include a unique naming authority for the tagCreator and sufficient additional details that the tagId is unique for the software product, version, edition, revision, etc. This would likely look as follows (+ is used as a string concatenation symbol): regid + productName + version + edition + revision + ...

tagVersion:
> The tagVersion indicates if a specific release of a software product has more than one tag that can represent that specific release. This may be the case if a software tag producer creates and releases an incorrect tag that they subsequently want to fix, but with no underlying changes to the product the SWID tag represents. This could happen if, for example, a patch is distributed that has a Link reference that does not cover all the various software releases it can patch. A newer SWID tag for that patch can be generated and the tagVersion value incremented to indicate that the data is updated.

So for tagId, I think we need to state that:
* if a GUID is specified, then lowercase is required. 
* If a GUID is not used, then the value of the tagId is case aware, but not case sensitive.

For tagVersion, I will update the PR to indicate that this is an optional integer value.

dkruszew (dismissed):
Looks good to me.

stevespringett:
@pombredanne @iamwillbar can one of you review, provide feedback and/or approve?

pombredanne (commented):
LGTM... with just one tiny nit: we have not used camelCase anywhere so far for qualifier keys.
Could it make sense to use `tag_id`, `tag_version`, `tag_creator_name`, `tag_creator_reg_id` and so on instead?

stevespringett:
Good catch @pombredanne. I've updated the qualifiers to use underscores.

pombredanne (approved):
LGTM! thank you++

PR #178, opened 2022-07-02
feat: add language qualitifer for multi language package managers

cpendery:
## 📝 Description
This adds a language qualifier to the purls for package managers where the packages can be written in multiple different source codes in order to allow for the identification of the package's language

Related to:
- https://github.com/anchore/syft/pull/1083
- https://github.com/anchore/syft/pull/1081
- https://github.com/anchore/syft/pull/1073

### DIFF
--- a/PURL-TYPES.rst
+++ b/PURL-TYPES.rst
@@ -44,6 +44,7 @@ cocoapods
 - The default repository is ``https://cdn.cocoapods.org/``
 - The ``name`` is the pod name and is case sensitive, cannot contain whitespace, a plus (+) character, or begin with a period (.).
 - The ``version`` is the package version.
+- The qualifier ``language`` is the source language of the package: ``swift`` or ``objectiveC``
 - The purl subpath is used to represent a pods subspec (if present)
 - Examples::
 
@@ -51,6 +52,7 @@ cocoapods
       pkg:cocoapods/MapsIndoors@3.24.0
       pkg:cocoapods/ShareKit@2.0#Twitter
       pkg:cocoapods/GoogleUtilities@7.5.2#NSData+zlib
+      pkg:cocoapods/MapsIndoors@3.24.0?language=swift
 
 cargo
 -----
@@ -86,10 +88,11 @@ conan
 - The ``name`` is the package name.
 - The ``version`` is the package version.
 - The qualifier ``channel`` must be not empty if namespace is present
+- The qualifier ``language`` is the source language of the package: ``c`` or ``cpp``
 - Examples::
 
       pkg:conan/cctz@2.3
-      pkg:conan/bincrafters/cctz@2.3?channel=stable
+      pkg:conan/bincrafters/cctz@2.3?channel=stable&language=cpp
 
 conda
 -----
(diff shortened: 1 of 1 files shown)


spiffcs (commented):
Just a heavy +1 from the [syft](https://github.com/anchore/syft/tree/main/syft) project. Being able to differentiate between languages as a parameter to the package system would help in making our cataloging process more accurate.

bureado (commented):
This is very interesting! Thank you for this! I can see how this helps cataloging and enriches dependency enumeration. As I'm unfamiliar with the package managers involved, I have a couple questions in the spirit of improving this proposal:

1. Can there ever be a package which contains source code written in two or more languages?
2. Looks like there's a fixed set of languages suggested for these package managers, is there an authoritative source for those? (For futureproofing purposes)
3. Where is the `language` property for these packages defined? Is there a metadata file that has this attribute? How is `language` expected to be populated in package URLs?
    * For example, in the case of `conan/bincrafters/cctz`, is the only way (other than inspecting the tarball) to know this is `language=cpp` by looking for the presence of `cpp_info`  in [`package_info()`](https://github.com/bincrafters/conan-center-index/blob/main/recipes/cctz/all/conanfile.py#L79)? And how would I go about `c`? What about the other ones proposed here?
4. Do we know of prior art in other ecosystems? For example, I see `pom.xml` has `type` and `classifier` but nothing for languages; how would Maven deal with this?
    1. In the case of Linux package managers, e.g. `deb`, I would expect this to be dealt with via tags (debtags) which aren't mapped in purl today and would need off-band resolution just like for `generic` or `github`

bureado:
@stevespringett for input

bureado:
Also see #168 appearing to introduce arbitrary qualifiers for `conan`.

stevespringett:
> Can there ever be a package which contains source code written in two or more languages?

Yes. Occurs in Maven Central where a project may have a combination of Java, Scala, and Kotlin source languages. These projects will also have XML or groovy depending on the build system used (Maven uses XML, Gradle uses groovy). Maven projects can also include "resources" which will typically consist of key/value properties, xml and json configuration, but technically, any language can be included there. Any application that has a dependency on Mozilla Rhino for example, would also typically include Javascript in one or more of the dependencies as well. Same thing for any of the other library that allows the execution of non-native languages on the JVM. 

Will also occur with npm where both Javascript and Typescript are used. npm also allows arbitrary scripts to run when installing by default. These scripts could be bash, powershell, etc, which would be yet another language.

stevespringett:
One additional point on Maven, may likely be true for others as well, if the Maven type is `war` (web archive), then you can expect either Java, Scala, or Kotlin source languages for the backend, and HTML, CSS, and Javascript for the frontend, all in a single package.

cpendery:
> 2. Looks like there's a fixed set of languages suggested for these package managers, is there an authoritative source for those? (For futureproofing purposes)

I don't know of any fixed set for these package managers, so it may be better to have a naming convention for this fields so the spec doesn't need to change if support of a new language gets added. I think lowercased snake case where characters should not require html encoding (thinking of `cpp` vs `c%2b%2b`) which seems to follow purl-spec convention. This would yield: `objective_c`, `c`, `cpp`, `elixir`, `swift`, `erlang`.

PR #179, opened 2022-07-04
Proposing optional os qualifier for oci type

puerco:
This PR proposes adding an optional `os` qualifier to the `oci` type. We currently have `arch` but we also need `os` to complete the platform information for container images.

Signed-off-by: Adolfo García Veytia (Puerco) <puerco@chainguard.dev>


bureado (dismissed):

iamwillbar:
This addition seems reasonable to me, @rnjudge since you originally added the `oci` type do you have any concerns with this?

captn3m0:
Perhaps a clarification that `os` here refers to the runtime the image is built for, and not the OS inside the container image?

PURL-TYPES.rst
```diff
@@ -317,6 +317,7 @@ including container images built by Docker and others:
 - Optional qualifiers may include:
 
   - ``arch``: key for a package architecture, when relevant
+  - ``os``: key for package os, when applicable
```
pombredanne, line 320:
```suggestion
  - ``os``: key for package operating system (os) with value such as ...
```
Also could explain how is this value discovered?

puerco: Changed it to:

`key for package operating system (os) with value such as linux, darwin, windows, etc.`

puerco:
Thanks for the review @pombredanne! I completely missed it back then. I've rebased the PR and modified the description as suggested.

pombredanne:
After the merge of PR, PURL types are now defined in JSON:
- See  #514

PR #180, opened 2022-07-06
Fix missing percent-encoding in purl example with repository_url

jsteinhofff:


jsteinhofff:
@sschuberth this fixes the issue we discussed in my Conan related PR when i was irritated by the missing encoding of a slash in the example, maybe you would be so kind to also review here?

sschuberth (approved):

maxhbr:
This is not necessary and does not match the test cases:
https://github.com/package-url/purl-spec/blob/37626c43ace346f38f390f9266ef99eef654bf66/test-suite-data.json#L110-L121

maxhbr:
created a revert PR: https://github.com/package-url/purl-spec/pull/184

PR #182, opened 2022-07-06
make the repo reuse compliant

maxhbr:
(close that PR, if it is not in the projects interest)


mjherzog:
Closing because most of the files referenced are obsolete.

PR #183, opened 2022-07-08
Add kubernetes bom tool as user

sbs2001:
Link: https://github.com/kubernetes-sigs/bom 

It outputs bill of materials with purl references.


jkowalleck (commented):
so the tool you want to add actually generates SPDX? well, SPDX is already on that list.
Could you elaborate why to add that tool?

PR #189, opened 2022-08-04
Fix URL of package-url rust-lang repository

Xiphoseer:
Changed from `packageurl-rs` to `packageurl.rs` because that seems to be what the actual repo/fork is using.


iamwillbar:
@pombredanne and @stevespringett is there a preference for having a fork in the package-url org vs just linking to the upstream?

stevespringett:
@iamwillbar it may be more difficult to govern a fork. IMO, I'd prefer the package-url org to host the authoritative/original repo, or if that's not an option, simply provide a link to another repo in the docs.

Xiphoseer:
Superseded by #211

PR #192, opened 2022-09-27
Request to reserve the qpkg type

jlb-bb:
BlackBerry respectfully requests reservation of the qpkg type to identify packages belonging to QNX's package manager. QNX is a commercial Unix-like real-time operating system, aimed primarily at the embedded systems market.

### DIFF
--- a/PURL-TYPES.rst
+++ b/PURL-TYPES.rst
@@ -326,6 +326,20 @@ nuget
 
       pkg:nuget/EnterpriseLibrary.Common@6.0.1304
 
+qpkg
+----
+``qpkg`` for QNX packages:
+
+- There is no default package repository: this should be implied either from
+  the ``namespace`` or using a repository base URL as ``repository_url`` 
+  qualifiers key.
+- The ``namespace`` is the vendor. It is not case sensitive and must be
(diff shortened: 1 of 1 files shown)


PURL-TYPES.rst
```diff
@@ -326,6 +326,20 @@ nuget
...
 
+qpkg
+----
+``qpkg`` for QNX packages:
+
+- There is no default package repository: this should be implied either from
+  the ``namespace`` or using a repository base URL as ``repository_url`` 
+  qualifiers key.
+- The ``namespace`` is the vendor. It is not case sensitive and must be
+  lowercased. The default ``namespace`` is ``blackberry``.
```
iamwillbar, line 337: Today we don't have default namespaces for purl types and clients would need to know to substitute `blackberry` for missing namespaces. I'd suggest instead making the `namespace` required and explicitly include `blackberry` in the example below (i.e. no default namespace).

pombredanne (approved):
LGTM... I wonder if there is at least some public package repository that shows what a QNX package looks like?

jlb-bb:
There’s an application that developers/engineers use to access qpkg, QNX Software Center. I think you need a license to access qpkg.

PR #198, opened 2022-10-28
Consistent punctuation in purl types spec

mprpic:
Just some minor nits here and there while I was working on a separate change to this file.


pombredanne (approved):
Than you ++.
LGTM!

PR #206, opened 2022-12-10
fix(test-suite-data): expect lexicographical sorting of qualifiers

williamboman:
Running into test failure for this test case only. I might've misunderstood the spec but I believe these should be lexicographically ordered?


williamboman:
Superseded by #210.

PR #207, opened 2022-12-10
feat(test-suite-data): add test case for checksum qualifier

williamboman:
Figured a test case like this would be helpful.


pombredanne:
This is a test case for:
- https://github.com/package-url/purl-spec/pull/209
This should be rebased and merged in that PR to have a consistent change.
@williamboman I know this was a long time ago.... but do you mind to do this merge?

pombredanne:
- See  #514

1. you update your PR to use the new approach, e.g., adopt the new tests schema and move your fix to the corresponding test file OR

PR #209, opened 2022-12-10
docs: consistent naming of the `checksum` qualifier key

williamboman:
This confused me a bit as the spec includes to different variants of this
qualifier: (`checksum` and `checksums`). From reading the spec I believe the
intended variant is `checksum`; this PR consolidates all references to this
qualifier.

Related PR: #207.


PURL-SPECIFICATION.rst
```diff
@@ -377,7 +377,7 @@ To parse a `purl` string in its components:
     - The `value` is the percent-decoded right side
     - UTF-8-decode the `value` if needed in your programming language
     - Discard any key/value pairs where the value is empty
-    - If the `key` is `checksums`, split the `value` on ',' to create
+    - If the `key` is `checksum`, split the `value` on ',' to create
       a list of `checksums`
```
johnmhoran, line 381: @williamboman One question: on line 381, should `checksums` be changed to remove the backticks, as you did on line 324?

pombredanne, line 381:
```suggestion
      a list of checksums
```

pombredanne (changes requested):
We have a problem here as I messed things up by specifying both "checksum" (singular) and "checksums" (plural) as keys for this qualifier originally. 

We need to recover from this before pushing further changes.
1. We need to survey what the PURL libraries and SBOM docs do today to assess the damage
2. We should decide on the path of least resistance, and least damage with either plural or singular
3. We should update the spec then, but also mentioning the legacy form so that tools authors can accept and fix either plural to singular or singular to plural

pombredanne:
And also update to use proper Rst with double backticks as in the latest main, as fixed by @andrewpollock

johnmhoran:
Hi @williamboman  .  Thank you for your PR.  When you have the chance, could you please resolve the conflicts referred to below?

pombredanne:
- Most implementations do not have specific ode for parsing checksum-related qualifiers
- @matt-phylum in Rust and @jdalton in JS as well as the PHP implementation are using `checksum` singular. Same for Kotlin.
- @matt Ruby uses plural https://github.com/package-url/packageurl-ruby
- @giterlizzi Perl uses plural https://github.com/giterlizzi/perl-URI-PackageURL

Therefore I am inclined to go singular and help Ruby and Perl authors to adopt the singular form.

We should also provide suggestions to also accept plural form for lenient parsing in the FAQ.

pombredanne (approved):
Let's merge this!

Issue #195, opened 2022-10-08
missing namespace in the examples of rpm section from PURL-TYPES.rst


sify21:
`pkg:rpm/centerim@4.22.10-1.el6?arch=i686&epoch=1&distro=fedora-25`
I guess namespace `fedora` is missing here, or is namespace optional?

johnmhoran:
Hello @sify21 .  The `namespace` component is optional.  See:

- The [core-spec](https://github.com/package-url/purl-spec/blob/master/PURL-SPECIFICATION.rst#package-url-specification-v10x) provides that `namespace` is optional -- "namespace: some name prefix such as a Maven groupid, a Docker image owner, a GitHub user or organization. Optional and type-specific."
- The `rpm` [type spec](https://github.com/package-url/purl-spec/blob/master/PURL-TYPES.rst#rpm) includes two examples, the second with no namespace (and both of which use the `distro` qualifier to identify `fedora-25`): 

  ```
  pkg:rpm/fedora/curl@7.50.3-1.fc25?arch=i386&distro=fedora-25
  pkg:rpm/centerim@4.22.10-1.el6?arch=i686&epoch=1&distro=fedora-25
  ```

johnmhoran:
If this ^ is inaccurate or incomplete, please chime in here so we can update the core spec as needed.

johnmhoran:
@sify21 Thanks for opening this issue.  If my response above answers your question, please feel free to close this issue.

PR #214, opened 2023-01-05
[rpm] Add VMWare Photon Example

captn3m0:
just for example purposes, to avoid confusion elsewhere.

There are lots of ways to represent this:

- `pkg:rpm/photon/systemd?distro=vmware-photon-1`
- `pkg:rpm/photon/systemd?distro=photon-1`
- `pkg:rpm/systemd?distro=vmware-photon-1`
- `pkg:rpm/systemd?distro=photon-1`

However, the representation in the example should hopefully be clear and provide some guidance.

### DIFF
--- a/PURL-TYPES.rst
+++ b/PURL-TYPES.rst
@@ -470,6 +470,7 @@ rpm
 
       pkg:rpm/fedora/curl@7.50.3-1.fc25?arch=i386&distro=fedora-25
       pkg:rpm/centerim@4.22.10-1.el6?arch=i686&epoch=1&distro=fedora-25
+      pkg:rpm/vmware/systemd?distro=photon-1
 
 swid
 -----


bureado:
@captn3m0, I'm drawing a bunch of inferences on what this one-line PR intends to convey, so please call me out if I incorrectly inferred anything.

The proposed example implies that, in the case of Photon, the value for `namespace` should be `vmware`. The PR further implies that a value of `photon` would be suboptimal.

For the record, we're discussing `namespace` as defined in https://github.com/package-url/purl-spec/blob/master/PURL-TYPES.rst#rpm, which says the `namespace` is the vendor, with vendor implied to be the "provider" of the distro that ships the package being referenced.

"Fedora" and "OpenSUSE" are used in the original examples. But is "Fedora" the "vendor" of "Fedora Linux"? Or is it "Fedora Project"? And "OpenSUSE Project" for the "OpenSUSE" distribution? In the case of Fedora and OpenSUSE, these are not important considerations.

In the case of Photon, though, it means that one would consider "VMWare" as the namespace, hence this PR.

However, I think that interpretation actually suggests the spec should evolve so the namespace is not the vendor but the distributor ID [as per `lsb_release`](https://refspecs.linuxfoundation.org/LSB_3.0.0/LSB-PDA/LSB-PDA/lsbrelease.html) or the distro ID [as per `systemd`](https://www.linux.org/docs/man5/os-release.html).

This would mean that the proposed example would actually read `pkg:rpm/photon/systemd?distro=photon-4.0` (the `distro` value being derived from Photon's actual [os-release](https://github.com/vmware/photon/blob/9b0f95da5e30be6e254e27c46dd765929c148c5f/SPECS/photon-release/photon-release.spec#L48))

An alternative to this approach would be to use the vendor as defined for example [in a CPE string](https://cpe.mitre.org/specification/). Using this definition, then `vmware` is acceptable, but it introduces a number of problems including the fact that in the CPE 2.3 dictionary calls `photon` differently: `photon_os`. I would find using CPE string parts a restrictive approach for a number of other reasons, too.

So this brings up an interesting `purl`-wide question. What is a vendor, particularly when it comes to Linux packages? There are 7 mentions to `vendor` in the types specification, one being `rpm` and the other ones being:

* For `alpm` (pacman), some are distro names, some are something else (e.g., `msys`)
* For `apk` and `deb`, it's distro names
* For `composer`, `conan` and `qpkg`, it appears to be the publisher name

The spirit behind this seems to be that `vendor` is a part that helps locate the package univocally on the Internet. To that extent, some ecosystems use publisher (like `qpkg`) and others use the distro name.

Following this spirit, the appropriate `namespace` for Photon would be, IMHO, `photon`.

If this is in fact the spirit and the logic that `purl` wants for using "vendor as namespace" then I suggest it's made more clear in places where it's ambiguous, perhaps `rpm` being one of them (I don't see `deb` as being particularly ambiguous, but it's situational and ecosystem-dependent)

Of course, ambiguity can always be resolved with `repository_url`. Maybe we also say that if someone intends to use a `purl` to assemble a URL to fetch a package from a repository (which is not a great idea considering snapshot repositories aren't too common) then they must use a `repository_url` or they risk ambiguity.

We might also want to recommend that `purl` users do some due diligence on what the preferred namespace for the ecosystem they're referencing to is. Maybe `photon` users prefer to use `vmware` as their `namespace`.

captn3m0:
I agree with most of the analysis, but this shouldn't be left to the end-users - I'm facing these conundrums everyday while generating PURLs). It should be made clear in the specification as to where the vendor/namespace are expected to be picked up from (such as deferring to /etc/os-release).

Even if we pick /etc/os-release, there's still more issues, such as both centos linux and centos stream (which are quite different operating systems), both using the exact same details in the lsb-release, even down to the same CPE: https://github.com/endoflife-date/endoflife.date/pull/1255#issuecomment-1139383709

For differentiating between centos 8 and centos stream 8 packages, you need to encode this in the distro field (such as centos-8, centos-stream-8), but there's no guidance provided to RPM users currently on what forms this will take.

Even in case of debian packages, there are scenarios where the PURL spec falls short, such as in the case of Debian ELTS:

1. The /etc/os-release doesn't change
2. However, the package repository changes to the ELTS repository

In such case, imo - the distro should remain `jessie`, as earlier, with the `repository_url` field for upgraded packages pointing to the new repository. However, since such a PURL might not be clear enough, we're discussing whether to change the [ecosystem field](https://github.com/ossf/osv-schema/pull/104#issuecomment-1372871791) to accommodate this additional information.

> The spirit behind this seems to be that vendor is a part that helps locate the package univocally on the Internet. 

From my usual understanding of the word "vendor", I'd just assumed this to mean "the supplier behind the package". "locate the package un-ambigiously" is something other fields also do, not just the vendor field, so it's not a clear definition.

Either the specification should provide a clear definition of what vendor is supposed to be, or it should provide guidance as to how the field can be inferred (with appropriate fallbacks).

imo, it is quite impossible for PURL to fulfill both of the obligations at the same time:

1. Not maintain a dictionary of keywords/mappings for fields like namespaces/distro/os/arch etc
4. Become a widely interoperable and easily-used specification.

bureado:
See:
- #195

pombredanne:
@captn3m0 Hey!

- See  #514

Thanks for your understanding !

captn3m0:
Updated it based on what `syft` is now already using:

`pkg:rpm/photon/bash@5.2-7.ph5?arch=x86_64&distro=photon-5.0` (likely based on the `/etc/os-release` data, using the same approach @bureado mentioned above).

I've added a simple test, and an example in the definition json.

One thing I had to remove was the `upstream` qualifier which is used by syft to document the source package name. [Added here](https://github.com/anchore/syft/pull/769) and noted as non-standard ("The pURL spec does not strictly support this qualifier,"). There's some more discussion at https://github.com/anchore/syft/issues/1700. I don't use RPM systems so someone else can perhaps look at making that part of the spec.

PR #218, opened 2023-02-21
Poposing optional media_type qualifier for OCI type purls

puerco:
This commit introduces a new `media_type` optional qualifier available to `oci`-typed purls to express the
type of OCI object they are identifying (image, manifest, layout, layer, etc).

Example purl identifying an index:

`pkg:oci/static@sha256%3A953aad6dd4eb0?media_type=application%2Fvnd.oci.image.index.v1%2Bjson`

/cc @rnjudge

### DIFF
--- a/PURL-TYPES.rst
+++ b/PURL-TYPES.rst
@@ -415,6 +415,8 @@ including container images built by Docker and others:
 - Optional qualifiers may include:
 
   - ``arch``: key for a package architecture, when relevant.
+  - ``media_type``: [Media type](https://github.com/opencontainers/image-spec/blob/main/media-types.md)
+    of the OCI object identified by the package URL (layer, manifest, config, etc).
   - ``repository_url``: A repository URL where the artifact may be found, but not
     intended as the only location. This value is encouraged to identify a
     location the content may be fetched.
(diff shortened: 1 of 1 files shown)


stevespringett:
Thanks @puerco. LGTM. Any comments @rnjudge?

rnjudge:
I need to go take a look at the original discussion but I thought I remember `media_type` being a rather contentious qualifier when OCI was first proposed...

rnjudge:
@puerco I think the confusion for me with `media_type` would be around whether folks are supposed to use the IANA media types, OCI media types or just a free-form string? Can you elaborate on use cases here?

rnjudge:
> I need to go take a look at the original discussion but I thought I remember `media_type` being a rather contentious qualifier when OCI was first proposed...

Ok, looks like I had proposed `type` (instead of media_type) as an optional qualifier with my original PR that was meant to indicate the type of OCI object being described. @dlorenc pushed back [saying](https://github.com/package-url/purl-spec/pull/123#issuecomment-921831347):

```
The type qualifier feels out of place, what's the rationale for inclusion? It seems a bit redundant
and error-prone, since things stored in OCI registries are generally self-describing. From a 
separation of concerns standpoint, a PURL would be used to identify and locate the object, then
the object/API contains information about what type of object it is (Image, Index, other). If the
type here didn't match the type in the object/API, I think that could cause issues? I don't see
a type qualifier on any of the other pURL types either, even the generic one where it might be
useful like here. The maven example seems a bit off, since for that the type is actually required by
maven to correctly resolve the artifact, where here it's just additional information for a client.
```

dlorenc:
Yeah - I think I have the same questions here. What happens if it doesn't match? If you're pulling by digest the media type will be in the contents of the thing you're referencing.

PR #220, opened 2023-03-01
Add lowercase requirement to composer name

oatovar:
This MR adds a requirement to have the `namespace` and `name` be lowercased since this is how names are expected by the [composer.lock](https://getcomposer.org/doc/04-schema.md#name) schema.


iamwillbar (approved):
Thanks.

PR #232, opened 2023-05-30
Remove cocoapods from canidate list

cs-shadow:
cocoapods type appears to be defined in #125, this entry would seems to be outdated now.


stevespringett (approved):
Thanks for the correction

PR #234, opened 2023-06-19
Add new PURL type: 'macos'

awsactran:
Request to add macos type to purl-spec documentation. Reserving macos for MacOS applications.

### DIFF
--- a/PURL-TYPES.rst
+++ b/PURL-TYPES.rst
@@ -314,15 +314,16 @@ huggingface
 
 macos
 -----
-``macos`` for native macOS packages and applications which are either pre-installed by Apple or installed via non-standard method (i.e. installed from .dmg file, downloaded from external sources)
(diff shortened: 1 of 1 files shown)


stevespringett:
How would appstore apps be specified? Some are platform specific while others work across multiple platforms (macOS, iOS, iPadOS, watchOS, tvOS, visionOS, etc)

awsactran:
`macos` type is reserved for applications that are inventoried from mac. With other platforms such as iOS, iPadOS, tvOS, etc, I think we can use `pkg:apple/macos` `pkg:apple/ios` instead. Would this be a blocker?

awsactran:
@stevespringett the new commit is meant to specify apple namespace and its platforms.

pombredanne:
@awsactran After the merge of PR #514, PURL types are now defined in JSON :angel: :innocent: :
- See  #514

Note also that there are some inconsistencies in your PR between the type and namespace in your latest push... it would be great to have someone from Apple chime in too.

PR #236, opened 2023-06-21
Add new PURL type: "helm"

tstephen-relx:


javierfreire:
I was also interested in having this purl type, but I would like it to also support an OCI registry.

Reviewing the [spec](https://github.com/package-url/purl-spec/blob/master/PURL-SPECIFICATION.rst#rules-for-each-purl-component): 
> A URL host or Authority must NOT be used as a namespace. Use instead a repository_url qualifier

So I think using the `repository_url` attribute instead of the namespace would be more correct and would allow using an OCI registry.

Also, I don't know if it would be interesting to add the `sha256` as an attribute (could be `digest` or `checksum` like the generic type)  in the case of OCI.

Examples::

    pkg:helm/postgresql@13.0.0?repository_url=oci://registry-1.docker.io/bitnamicharts
    pkg:helm/postgresql@13.0.0?repository_url=oci://registry-1.docker.io/bitnamicharts&digest=sha256:e48a5f264166fa36a778501b9df0d61528e5e5fba8c9b2b55dbb914743025df9

beltran-rubo:
@stevespringett from Bitnami we are interested on using this specification for Helm charts. Anything we can help to add this proposal to the purl standard?

pombredanne:
@tstephen-relx @beltran-rubo After the merge of PR #514, PURL types are now defined in JSON :angel: :innocent: :
- See  #514

If you can help we should be able to merge this quick enough

andrew:
Would you like help updating this purl type for the next json schema? Happy to update this pr or open a new one. 

For context I'm working on adding support for helm charts to various https://github.com/git-pkgs tools

PR #237, opened 2023-06-21
Add Julia type and tests

Octogonapus:
This PR adds a Julia type, along with some tests based on the provided examples.

cc https://github.com/package-url/packageurl-go/pull/44

### DIFF
--- a/PURL-TYPES.rst
+++ b/PURL-TYPES.rst
@@ -326,6 +326,24 @@ huggingface
       pkg:huggingface/microsoft/deberta-v3-base@559062ad13d311b87b2c455e67dcd5f1c8f65111?repository_url=https://hub-ci.huggingface.co
 
 
+julia
+------
+``julia`` for Julia packages:
+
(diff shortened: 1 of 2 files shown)


Octogonapus:
I have used this spec to implement https://github.com/aquasecurity/trivy/pull/5635. Without further review from the maintainers of this organization, I would just assume this spec is sufficient and use it for whatever you need.

g-sahil22:
Hi @Octogonapus , I have tried trivy to generate SBOM for an Open-Source Julia project, but in the SBOM I didn't report any components

Octogonapus:
Let's move the discussion to https://github.com/aquasecurity/trivy/pull/5635
Also you will need to tell me which julia project, and keep in mind you need to build trivy from source using my PR

pombredanne:
After the merge of PR #514, PURL types are now defined in JSON :angel: :innocent: :
- See  #514

Octogonapus:
Replaced by https://github.com/package-url/purl-spec/pull/540

PR #243, opened 2023-07-19
Add GitHub Actions type

jhutchings1:
This PR adds GitHub Actions as a distinct type. We use this in the GitHub Dependency graph because GitHub Actions are distinct in meaning from the GitHub repository package references, and sometimes get CVEs published on them.

### DIFF
--- a/PURL-TYPES.rst
+++ b/PURL-TYPES.rst
@@ -253,6 +253,23 @@ github
       pkg:github/package-url/purl-spec@244fd47e07d1004
       pkg:github/package-url/purl-spec@244fd47e07d1004#everybody/loves/dogs
 
+githubactions
+------
+``githubactions`` for GitHub Actions:
+
+- The ``namespace`` is the user or organization. It is not case sensitive and
+  must be lowercased.
+- The ``name`` is the repository name. It is not case sensitive and must be
+  lowercased.
+- The ``subpath`` is used to point to a secondary action within a repository. It is not case sensitive and must be lowercased.
+- The ``version`` is a commit or release tag. 
+- Examples::
+
+      pkg:githubactions/package-url/purl-spec@244fd47e07d1004
+      pkg:githubactions/package-url/purl-spec@v1.2
+      pkg:githubactions/github/codeql-action/analyze@v2
+
+
 golang
 ------
 ``golang`` for Go packages:
--- a/PURL-TYPES.rst
+++ b/PURL-TYPES.rst
@@ -261,7 +261,7 @@ githubactions
   must be lowercased.
 - The ``name`` is the repository name. It is not case sensitive and must be
   lowercased.
-- The ``subpath`` is used to point to a secondary action within a repository. It is not case sensitive and must be lowercased.
+- The ``subpath`` is used to point to the location of an action within a repository in the event there are multiple defined. It is not case sensitive and must be lowercased. If omitted, the package-url refers to the default action for the repository only.
 - The ``version`` is a commit SHA or release tag. 
 - The ``repository_url`` refers to the GitHub instance that hosts the action you are using. If omitted, the default is assumed to be `https://github.com/`.
 - Examples::
--- a/PURL-TYPES.rst
+++ b/PURL-TYPES.rst
@@ -262,7 +262,8 @@ githubactions
 - The ``name`` is the repository name. It is not case sensitive and must be
   lowercased.
 - The ``subpath`` is used to point to a secondary action within a repository. It is not case sensitive and must be lowercased.
-- The ``version`` is a commit or release tag. 
+- The ``version`` is a commit SHA or release tag. 
+- The ``repositoryUrl`` refers to the GitHub instance you are using. If omitted, the default is assumed to be `https://github.com/`.
 - Examples::
 
       pkg:githubactions/package-url/purl-spec@244fd47e07d1004
--- a/PURL-TYPES.rst
+++ b/PURL-TYPES.rst
@@ -263,7 +263,7 @@ githubactions
   lowercased.
 - The ``subpath`` is used to point to a secondary action within a repository. It is not case sensitive and must be lowercased.
 - The ``version`` is a commit SHA or release tag. 
-- The ``repositoryUrl`` refers to the GitHub instance you are using. If omitted, the default is assumed to be `https://github.com/`.
+- The ``repository_url`` refers to the GitHub instance that hosts the action you are using. If omitted, the default is assumed to be `https://github.com/`.
 - Examples::
 
       pkg:githubactions/package-url/purl-spec@244fd47e07d1004


jhutchings1:
@pombredanne @stevespringett  Can you take a look at this? We're already using it in practice within GitHub, and I'd love to make sure that it's an accepted type definition.

PURL-TYPES.rst
```diff
@@ -253,6 +253,23 @@ github
...
+  must be lowercased.
+- The ``name`` is the repository name. It is not case sensitive and must be
+  lowercased.
+- The ``subpath`` is used to point to a secondary action within a repository. It is not case sensitive and must be lowercased.
+- The ``version`` is a commit or release tag. 
+- Examples::
+
+      pkg:githubactions/package-url/purl-spec@244fd47e07d1004
+      pkg:githubactions/package-url/purl-spec@v1.2
+      pkg:githubactions/github/codeql-action/analyze@v2
```
stevespringett, line 270: What is the relationship between `pkg:githubactions/github/codeql-action/analyze@v2` and the Marketplace where actions are found? For example, this action is not listed in the marketplace, but https://github.com/marketplace/actions/codeql-bundle is. I think this needs to be clarified as many people will associate the `githubactions` purl type with the actions that are in the Marketplace, but that's not true in this case.

PURL-TYPES.rst
```diff
@@ -253,6 +253,23 @@ github
       pkg:github/package-url/purl-spec@244fd47e07d1004
       pkg:github/package-url/purl-spec@244fd47e07d1004#everybody/loves/dogs
 
+githubactions
+------
+``githubactions`` for GitHub Actions:
```
stevespringett, line 258: What is the default `repository_url`? Does this apply to Actions? If not, why not? The purl spec does not differentiate between public Internet facing repositories and internal ones. If there's a default `repository_url`, even an internal one, it should be part of the purl type definition.

PURL-TYPES.rst
```diff
@@ -253,6 +253,23 @@ github
...
+
+- The ``namespace`` is the user or organization. It is not case sensitive and
+  must be lowercased.
+- The ``name`` is the repository name. It is not case sensitive and must be
+  lowercased.
+- The ``subpath`` is used to point to a secondary action within a repository. It is not case sensitive and must be lowercased.
```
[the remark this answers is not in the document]
jhutchings1: It's arbitrary. codeql-action is a good example, because we have 4 different actions in that repository. That field just specifies the path to that specific action within the home repository. 
github/codeql-action/init
github/codeql-action/analyze
github/codeql-action/autobuild
github/codeql-action/upload-sarif

PURL-TYPES.rst
jhutchings1: The Marketplace listings are basically advertisements and disconnected from the actual software distribution. You can reference an action in any repository regardless of whether there is a Marketplace listing. The only requirement is that there's an action.yml file in the repository and that the version matches a SHA, branch, or release tag. You'll see some of the scenarios around how things get referenced here: https://docs.github.com/en/actions/learn-github-actions/finding-and-customizing-actions#adding-an-action-to-your-workflow

PURL-TYPES.rst
jhutchings1: Sure, I could add one of those if you think it's helpful. GitHub Actions are always in practice pulled from the current GitHub instance you're on. So if you're on github.com, it's that, and if you're on an enterprise server instance, it's that.

PURL-TYPES.rst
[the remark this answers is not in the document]
hmaurer: I agree that the wording "secondary" is misleading, e.g. in https://github.com/github/codeql-action's case there is no GitHub Action at the root of the repository (no "primary" GitHub Action in this wording). We should somehow say that `subpath` is the directory containing the GitHub Action within the repository.

PURL-TYPES.rst
```diff
@@ -253,6 +253,24 @@ github
...
+- The ``name`` is the repository name. It is not case sensitive and must be
+  lowercased.
+- The ``subpath`` is used to point to the location of an action within a repository in the event there are multiple defined. It is not case sensitive and must be lowercased. If omitted, the package-url refers to the default action for the repository only.
+- The ``version`` is a commit SHA or release tag. 
+- The ``repository_url`` refers to the GitHub instance that hosts the action you are using. If omitted, the default is assumed to be `https://github.com/`.
+- Examples::
+
+      pkg:githubactions/package-url/purl-spec@244fd47e07d1004
+      pkg:githubactions/package-url/purl-spec@v1.2
+      pkg:githubactions/github/codeql-action/analyze@v2
```
prabhu, line 271: It appears like the name attribute could be a single word `purl-spec` or could have subpath `codeql-action/analyze`. This appear to be different from other schemes where subpath are after qualifiers separated by a hash.

```
pkg:githubactions/github/codeql-action@v2?repository_url=...#subpath=analyze
```

emilwareus:
Hi! I'm trying to use this schema, how would you say this vulnerability would fit into the actions pURL? https://github.com/advisories/GHSA-hw6r-g8gj-2987

PURL-TYPES.rst
matt-phylum: I think subpath has a special meaning for referring to files within a package, which is not the case here if the action is its own package which happens to be in the same repository as other packages. This seems consistent with Go where the namespace/name is treated as a single value which means something in Go instead of two separate values. (eg a lot of Go PURLs have the PURL name "v2" because of the way Go handles version epochs)

jhutchings1: @matt-phylum has it. The subpath is about picking a non-default action in a repository in instances where multiple actions exist.

tjwp: This seems inconsistent with the description given above:

> ``namespace`` is the user or organization
> ``name`` is the repository name
> ``subpath`` is used to point to the location of an action within a repository in the event there are multiple defined.

If the `name` component is strictly the repository name, then based on the components from the spec:

`scheme:type/namespace/name@version?qualifiers#subpath`

This example should be:

`pkg:githubactions/github/codeql-action@v2#analyze`

martincostello:
Poke for progress, as I just wasted 10 minutes trying to work out what the prefix was to use with [actions/dependency-review-action](https://github.com/actions/dependency-review-action)'s `allow-dependencies-licenses` option.

`github` and `github-actions` didn't work, despite Copilot's claim the latter was correct.

I worked out it should be `githubactions` eventually by tracing the action's code to get to [this line](https://github.com/actions/dependency-review-action/blob/89c73830747a68fcfdecb3b682ff934fc8c3abf1/src/dependency-graph.ts#L31C13-L31C70), at which point I realised I could download an SBOM from the repo, open it and search for an action to find the emitted purl prefix from there.

pombredanne:
@martincostello Hey, there have been discussions recently in a community whether this should exist as a type and whether not just use `github` for actions... and a `github-action` may not be needed?

Also, after the merge of PR #514, PURL types are now defined in JSON :angel: :innocent: :
- See  #514

With the new approach... this PR would need to be updated if we go with a separate type.

jhutchings1:
2 years, quite the zombie PR! @trevrosen may have thoughts on this. I'm no longer working on supply chain security, so I can't say if GitHub still wants this capability or how.

sjn:
Would it make sense to revive this ticket, in light of https://www.getsafety.com/blog-posts/shai-hulud-npm-attack and https://www.stepsecurity.io/blog/ctrl-tinycolor-and-40-npm-packages-compromised ?

I'd love to see github produce SBOMs describing what actions where used during a run, where any actions that were "pulled in" were referred to in the form of a PackageURL that clearly distinguishes actions from other types of package dependencies.

>  Different things should look different. Similar things should look similar.
>  – Larry Wall

nicorikken:
@sjn good idea. I created #698 as placeholder issue.

PR #250, opened 2023-10-02
pypi: change default repository

schischi:
`python.pypi.org` is the legacy index domain name and was replaced by `pypi.org` in 2017. See: https://packaging.python.org/en/latest/glossary/#term-pypi.org

### DIFF
--- a/PURL-TYPES.rst
+++ b/PURL-TYPES.rst
@@ -457,7 +457,7 @@ pypi
 ----
 ``pypi`` for Python packages:
 
-- The default repository is ``https://pypi.python.org``.
+- The default repository is ``https://pypi.org``.
 - PyPI treats ``-`` and ``_`` as the same character and is not case sensitive.
   Therefore a PyPI package ``name`` must be lowercased and underscore ``_``
   replaced with a dash ``-``.
--- a/PURL-TYPES.rst
+++ b/PURL-TYPES.rst
@@ -457,7 +457,7 @@ pypi
 ----
 ``pypi`` for Python packages:
 
-- The default repository is ``https://pypi.org``.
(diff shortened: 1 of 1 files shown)


sethmlarson (approved):

woodruffw (approved):

pombredanne:
@schischi Thanks... this makes perfect sense. We may want to keep a short note about the previous URL so folks are not confused.

PURL-TYPES.rst
```diff
@@ -457,7 +457,7 @@ pypi
 ----
 ``pypi`` for Python packages:
 
-- The default repository is ``https://pypi.python.org``.
+- The default repository is ``https://pypi.org``.
```
pombredanne, line 460:
```suggestion
- The default repository is ``https://pypi.org``. (Previously  ``https://pypi.python.org``.)
```

pombredanne (changes requested):
Here is a simple suggestion

criminosis:
@pombredanne out of curiosity does this have any implications for purl equality?

Given a purl `pkg:pypi/Django@4.2.7` and the prior implicit default of `https://pypi.python.org`.

Was it intended for `pkg:pypi/Django@4.2.7` and `pkg:pypi/Django@4.2.7?repository_url=https://pypi.python.org` to be "equal" purls?

Now that it's `https://pypi.org` are there now 3 cases that are intended to be canonically equal to maintain backwards compatibility with "old" purls?
-  `pkg:pypi/Django@4.2.7`(No `repository_url` so implied default)
- `pkg:pypi/Django@4.2.7?repository_url=https://pypi.python.org`
- `pkg:pypi/Django@4.2.7?repository_url=https://pypi.org`

PR #256, opened 2023-10-24
Disambiguate purl from PURL with historical context

edburns:


pombredanne:
This make sense... but also everyone wants to use PURL for this .... in all case I will grab this for the Ecma standard... as important background info

PR #259, opened 2023-10-29
Add CodeProject PURL type

juliancoccia:


PURL-TYPES.rst
```diff
@@ -115,6 +115,21 @@ cargo
       pkg:cargo/clap@2.33.0
       pkg:cargo/structopt@0.3.11
 
+codeproject
+-----------
+
+``codeproject`` for CodeProject scripts:
+
+- The default repository is ``https://www.codeproject.com``.
```
prabhu, line 123: At a minimum, there must be a package manager or an automated way to download and verify packages to qualify for a package-url. An article or a blog post is not a package.

juliancoccia: Hi @prabhu , there is not such thing in Code Project. However, there are quite a number of Open Source components published in that site many years ago and these packages are not available anywhere else. Is then the PURL specification restricted only to repositories that offer "a package manager or an automated way to download and verify packages"? That would be quite unfortunate. If that is the case, what would be your advice for naming those components?

mjherzog:
The general recommendation for using PURL to identify software components without a package manager is to use the generic type and specify the download_url in the qualifier section.

juliancoccia:
Understood. Neither GitHub nor CodeProject are package managers per se. The PURL spec appears to make accommodations for GitHub repositories despite them not having a traditional package manager. Might CodeProject deserve similar consideration as a hosting platform for software components, especially for those unique packages not available elsewhere?​​​​​​​​​​​​​​​​

pombredanne:
After the merge of PR #514, PURL types are now defined in JSON :angel: :innocent: :grin: :
- See  #514

With the new approach... this PR would need to be updated .... but I am not convinced that codeproject deserves a type. It was to close BTW, and was recently revived.... and it is for sure an important source code historica; C#, C/C++ and VB code snippets using the infamous CPOL license!
So the debate is still open.

PR #269, opened 2023-11-10
[Bitnami] Add distro field

gongomgra:
Add the `distro` qualifier key to Bitnami purl definition.


gongomgra:
@stevespringett can we get your inputs here? Thanks!

stevespringett (approved):
Thanks for the ping. LGTM

PR #273, opened 2023-11-20
add test for percent signs

matt-phylum:
This PR adds a simple test that percent signs are correctly encoded and decoded in all parts of a PURL. Some implementations don't correctly handle percent encoding in all allowed positions (giterlizzi/perl-URI-PackageURL, maennchen/purl, package-url/packageurl-ruby, package-url/packageurl-swift, sonatype/package-url-java) and some implementations don't encode percent signs (althonos/packageurl.rs, anchore/packageurl-go, phylum-dev/purl).


pombredanne:
@matt-phylum here too, after the merge of PR #514, PURL tests and defs are now defined in new JSON schema :angel: :innocent: :grin: :
- See  #514

....  this PR needs to be updated. Sorry for the pain. Thank you for your patience!

PR #276, opened 2023-11-24
Add distro to the Bitnami type

javierfreire:
The Bitnami packages include the distro in addition to the name, version, and architecture. 

Example: https://downloads.bitnami.com/files/stacksmith/wordpress-6.2.0-9-linux-amd64-photon-4.tar.gz

We (Bitnami) have started to include this format in our SPDXs and we wanted it to be reflected in the specification.


stevespringett:
This appears to be a duplicate of #269

javierfreire:
You're right. I'll close this.

PR #277, opened 2023-11-27
add note about nuget checksums

matt-phylum:
This PR adds clarification around NuGet package checksum qualifiers.

Should there also be a `nuget_lock_sha512` qualifier or similar for holding the package hash? It could be useful for cases where somebody is trying to generate an SBOM from a `packages.lock.json` file. NuGet packages are not namespaced by their source repository and the source repository information is not included in the `packages.lock.json` file. The package hash is the only disambiguating information in that case if there are multiple packages with the same name and version.


johnmhoran:
Thank you for this PR @matt-phylum . When you have the chance, could you please merge the latest master into this branch?

pombredanne:
Here too, after the merge of PR #514, PURL tests and defs are now defined in new JSON schema :angel: :innocent: :grin: :
- See  #514

With the new approach... this PR needs to be updated. Sorry for the pain. Thank you for your patience! :)

PR #288, opened 2024-02-02
Add "cpan" Perl Packages type

giterlizzi:


sjn:
@pombredanne here's the CPAN spec addition, as we discussed! 😊

sjn:
Is it necessary to mention metacpan at all?

PURL-TYPES.rst
```diff
@@ -168,6 +168,35 @@ conda
 
       pkg:conda/absl-py@0.4.1?build=py36h06a4308_0&channel=main&subdir=linux-64&type=tar.bz2
 
+cpan
+----
+``cpan`` for CPAN Perl packages:
+
+- The default respository is ``https://www.cpan.org/``.
+- To search CPAN it is recommended to use ``https://metacpan.org``.
```
sjn, line 176: I think we safely can remove this line, since the module search web page isn't relevant for how PackageURLs work.

giterlizzi: Corrected ;)

pombredanne (approved):
LGTM! Thank you ++

giterlizzi:
Thank you !!!

sjn:
Would it be meaningful to change the text to use only the terms defined in https://datatracker.ietf.org/doc/html/rfc2119? (e.g. MUST, and MUST NOT, and do not use the term ~~NEVER~~)

PR #205, opened 2022-12-02
Add `luarocks` package type

hishamhm:
Hi, I am the lead mantainer of the [LuaRocks](https://luarocks.org) package manager and we'd like to add its entry to the `purl` spec. Thanks!

PR #297, opened 2024-03-13
Add the spec for luarock packages

LaurentGoderre:
This adds documentation for the `luarock` package type used for the Lua language.

### DIFF
--- a/PURL-TYPES.rst
+++ b/PURL-TYPES.rst
@@ -356,6 +356,16 @@ huggingface
       pkg:huggingface/microsoft/deberta-v3-base@559062ad13d311b87b2c455e67dcd5f1c8f65111?repository_url=https://hub-ci.huggingface.co
 
 
+luarock
+--------
+``luarock`` for Lua language packages
+
+- The default repository is ``https://luarocks.org/``.
+- Examples::
+
+  pkg:luarock/lua-resty-http@0.17.1-0
+  pkg:luarock/luasql-mysql@2.6.0-3
+
 maven
 -----
 ``maven`` for Maven JARs and related artifacts:
@@ -579,7 +589,6 @@ Other candidate types to define:
 - ``haxe`` for Haxe packages:
 - ``helm`` for Kubernetes packages
 - ``julia`` for Julia packages:
-- ``lua`` for LuaRocks packages:
 - ``melpa`` for Emacs packages
 - ``meteor`` for Meteor JavaScript packages:
 - ``nim`` for Nim packages:


LaurentGoderre:
@spiffcs thoughts?

LaurentGoderre:
Also cocoapods seems to be the only one that is pluralized like this. It seems weird to have a plural in a url that specifies a single package

matt-phylum:
LuaRocks is the name of the ecosystem and it calls a package a rock. LuaRock doesn't seem right.

spiffcs (approved):

spiffcs:
Apologies - submitted this on the wrong tab on my machine while going through another PR.

I'll add my thoughts on this in a separate comment about how we're looking to consume this over in syft which might bend the discussion towards a solution for how to represent this package type.

PURL-TYPES.rst
```diff
@@ -356,6 +356,16 @@ huggingface
       pkg:huggingface/microsoft/deberta-v3-base@559062ad13d311b87b2c455e67dcd5f1c8f65111?repository_url=https://hub-ci.huggingface.co
 
 
+luarock
```
Tieske, line 359: the tool is called `luarocks` (plural), so the singular version seems a bit weird to me.

LaurentGoderre: My thought process is that if the ecosystem was called for example `RandomBinaries` then the Purl would `pkg:RandomBinary/SomeRandomBinary`

matt-phylum: LuaRocks is actually singular despite ending with an s. That's the reason LuaRock is weird. LuaRocks (singular) is the name of the package manager which installs rocks (plural of rock).

It's the same as RubyGems. RubyGems (singular) is the name of the package manager which installs gems (plural of gem). PURL uses `pkg:gem`.

LaurentGoderre: So there are examples for both cases then where a single packager is referred to in singular or plural. The reason we named it LuaRock is that rock might conflict so a rock for Lua.

Ultimately, it doesn't really matter to me, whatever the community choses is fine with me

matt-phylum: If we're worried about collisions it should probably be `luarocks` like how it is with `cocoapods`. LuaRock is not the singular form of LuaRocks. Package names are lowercase so I guess `luarock` could be read as "Lua rock", but it seems close enough to LuaRocks to cause confusion.

Tieske: Not natively English, but the common way to refer to them is `luarocks` for the tool, and a `rock` or `rockspec` for the packages and mainfests. If you use `luarock` anywhere, users will be confused probably whether you mean the tool/group, or an individual rock.

So to me `pkg:luarocks/penlight` would make most sense.

hishamhm:
Hello, LuaRocks lead maintainer here!

> The type must be rock (like gem) or luarocks (like cocoapods) but not luarock.

On naming: yes, definitely _not_ `luarock`, that denomination is never used, anywhere in the system. Definitely recommend `luarocks` because `rock` is not distinct enough (only LuaRocks users would look at `pkg:rock` and think "ah, that's Lua"). For RubyGems, `gem` is actually the CLI name so that's a well-known name even for non Rubyists, but for LuaRocks, a much smaller ecosystem, the CLI is `luarocks`.

We only use `rock` on its own to refer to the packages (as in "install this rock with LuaRocks" and as the filename extension for packages (`.rock`), but for a cross-language tool I don't think `rock` is distinct enough. But in any case, please do _not_ use `luarock` singular for any purpose.

> @fperrad when. installing rock packages, there is no namespace and the rock files don't have namespace either.

LuaRocks has optional namespacing. It has the concept of a "root manifest" which is a collection of packages that get "first dibs" on each name, so that, for example, if you run `luarocks install luasocket` you get the one from the user linked in the root manifest (currently, [lunarmodules/luasocket](https://luarocks.org/modules/lunarmodules/luasocket)). However, you can also do, for example, `luarocks install fyp/luasocket` and get [a fork from a different user](https://luarocks.org/modules/fyp/luasocket). You _can_ specify namespaced entries in rock dependencies. If a dependency is non-namespaced, a user can choose to install a fork ahead of time and use that custom version to resolve the non-namespaced dependency, but if a rock specifies a namespaced dependency, it will check whether the installed dependency has the namespace metadata.

I was pretty sure I had submitted a PR to add LuaRocks to the purl-spec 2 years ago(!) but I can't find the PR, so maybe I wrote it but ended up not submitting it... But I did find my fork with what I wrote back then:

https://github.com/hishamhm/purl-spec/commit/783ba8c17a1e303aaa7214915e3e47755f832f50

```md
luarocks
--------
``luarocks`` for Lua packages installed with LuaRocks:

- ``namespace``: The user manifest under which the package is registered.
  If not given, the root manifest is assumed.
- ``name``: The LuaRocks package name.
- ``version``: The full LuaRocks package version, including module version and rockspec revision.
- Qualifier ``repository_url``: The LuaRocks rocks server to be used;
  useful in case a private server is used (optional).
  If ommitted, ``https://luarocks.org`` as default server is assumed.

Examples::

      pkg:luarocks/luasocket@3.1.0-1
      pkg:luarocks/hisham/luafilesystem@1.8.0-1
      pkg:luarocks/username/packagename@0.1.0-1?repository_url=https://example.com/private_rocks_server/
```

matt-phylum:
> I was pretty sure I had submitted a PR to add LuaRocks to the purl-spec 2 years ago(!) but I can't find the PR, so maybe I wrote it but ended up not submitting it... But I did find my fork with what I wrote back then:

You did. #205

stevespringett:
Lets flush out #205 since its more complete than this PR. @hishamhm can you take a look at #205 again and lets just clarify a few things and we'll get luarocks support added.

PR #312, opened 2024-07-12
apk: define `arch=source` qualifier

gernot-h:
This proposes a similar approach for Alpine to what we agreed on for Debian in #57. 

Thanks for reviewing, let me know what you think and if you need an additional issue for it!

### DIFF
--- a/PURL-TYPES.rst
+++ b/PURL-TYPES.rst
@@ -53,7 +53,8 @@ apk
 - The ``name`` is the package name. It is not case sensitive and must be
   lowercased.
 - The ``version`` is a package version as expected by apk.
-- The ``arch`` is the qualifiers key for a package architecture.
+- The ``arch`` is the qualifiers key for a package architecture. The special value
+  ``arch=source`` identifies an Alpine source package as built by ``abuild srcpkg``.
 - Examples::
 
       pkg:apk/alpine/curl@7.83.0-r0?arch=x86


gernot-h:
@Foxboron, as you contributed apk, I would also be interested in your opinion.

Foxboron:
I've read the Debian discussion and I don't think I agree with lamby and would prefer having a `src` qualifier to go along with `pkg`, but considering this was decided in 2021 I don't think my opinion matters on this.

Generally I think overloading the `arch` field is a mistake that is making this spec complicated.

gernot-h:
Thanks, @Foxboron, for your opinion! Apart from the _how_ we express this, you agree that it actually makes sense _if_ we define a way to specify package sources and also use `abuild srcpkg` as definition, right?

Foxboron:
> Apart from the how we express this, you agree that it actually makes sense if we define a way to specify package sources and also use abuild srcpkg as reference, right?

Yes

oej (approved):

gernot-h:
Is there anything I can do to complete or improve this PR?

gernot-h:
@pombredanne @stevespringett, as you both engaged intensively in [my former `arch=source` PR](https://github.com/package-url/purl-spec/pull/57), would you mind having a look into this one, too? TIA!

pombredanne:
@gernot-h Thanks for the patience! And I am sorry for the pain... but after the merge of PR #514, PURL tests and defs are now defined in new JSON schemas :angel: :innocent: :grin: :
- See  #514

With the new approach... this PR needs to be updated. Sorry for the churn. :heart:

gernot-h:
> @gernot-h Thanks for the patience! And I am sorry for the pain... but after the merge of PR #514, PURL tests and defs are now defined in new JSON schemas 👼 😇 😁 :
> 
>     * See  [Refine Purl type schema #514](https://github.com/package-url/purl-spec/pull/514)
> 
> 
> With the new approach... this PR needs to be updated. Sorry for the churn. ❤️

Thanks for the heads-up, @pombredanne! Sorry that I didn't answer earlier! I will look into it in September after the vacation period!

gernot-h:
Sorry for the delay, @pombredanne! But I adapted my changes to the new schemas now. I hope I got it right!

PR #315, opened 2024-07-24
Add new PURL type: 'pack' for SWI Prolog packages

LaurentGoderre:
Package specification: https://www.swi-prolog.org/howto/PackInfo.md


pombredanne:
- See  #514

... therefore with the new approach... this PR would need to be updated. Sorry for the churn. :heart: 


Do you think you can update this?

PR #317, opened 2024-08-01
Add `bazel` type for Bazel modules

fmeum:
Bazel 6 introduced a new system for managing external dependencies centered around the concept of [Bazel modules](https://bazel.build/external/module), which are hosted in a [registry](https://bazel.build/external/registry). The default registry is the [Bazel Central Registry](https://bcr.bazel.build). This system will become the default this year and its predecessor will be turned off next year.

As discussed in https://github.com/bazelbuild/bazel/discussions/23166, we would thus like to register the `bazel` purl type for Bazel modules, as specified in this PR.

(currently waiting for feedback from the [Rules Authors SIG](https://github.com/bazel-contrib/SIG-rules-authors))

### DIFF
--- a/PURL-TYPES.rst
+++ b/PURL-TYPES.rst
@@ -59,6 +59,24 @@ apk
       pkg:apk/alpine/curl@7.83.0-r0?arch=x86
       pkg:apk/alpine/apk@2.12.9-r3?arch=x86
 
+bazel
+-----
+``bazel`` for Bazel modules:
+
+- The default repository ("registry") is ``https://bcr.bazel.build``, the
+  Bazel Central Registry (BCR).
+- The ``namespace``, if specified, is the host and path of the URL of an
+  alternative registry. It is not case sensitive and must be lowercased.
+- The ``name`` is the module name. It is case sensitive, but a single registry
+  will never contain two different module names that compare equal
+  case-insensitively.
+- The ``version`` is the module version.
+- Examples::
+
+      pkg:bazel/rules_java@7.8.0
+      pkg:bazel/curl@8.8.0.bcr.1
+      pkg:bazel/example.org/bazel-registry/curl@8.8.0
+
 bitbucket
 ---------
 ``bitbucket`` for Bitbucket-based packages:
--- a/PURL-TYPES.rst
+++ b/PURL-TYPES.rst
@@ -65,17 +65,23 @@ bazel
 
 - The default repository ("registry") is ``https://bcr.bazel.build``, the
   Bazel Central Registry (BCR).
-- The ``namespace``, if specified, is the host and path of the URL of an
-  alternative registry. It is not case sensitive and must be lowercased.
 - The ``name`` is the module name. It is case sensitive, but a single registry
   will never contain two different module names that compare equal
   case-insensitively.
-- The ``version`` is the module version.
+- The ``version`` is the module version in `Bazel's relaxed semver format
+  <https://bazel.build/external/module#version_format>`_.
+- The optional ``repository_url`` can be used to specify the URL of an
+  alternative registry.
+- The optional ``subpath`` can name a particular Bazel target in the module via
+  a label with the leading double slash (``//``) removed and canonicalized by
+  omitting the target name if it is equal to the name of the containing package. 
 - Examples::
 
       pkg:bazel/rules_java@7.8.0
       pkg:bazel/curl@8.8.0.bcr.1
-      pkg:bazel/example.org/bazel-registry/curl@8.8.0
+      pkg:bazel/curl@8.8.0?repository_url=https://example.org/bazel-registry
+      pkg:bazel/rules_java@7.8.0
+      pkg:bazel/rules_java@7.8.0#toolchains:singlejar
 
 bitbucket
 ---------
--- a/PURL-TYPES.rst
+++ b/PURL-TYPES.rst
@@ -65,16 +65,14 @@ bazel
 
 - The default repository ("registry") is ``https://bcr.bazel.build``, the
   Bazel Central Registry (BCR).
-- The ``name`` is the module name. It is case sensitive, but a single registry
-  will never contain two different module names that compare equal
-  case-insensitively.
+- The ``name`` is the module name. It must be lowercased.
 - The ``version`` is the module version in `Bazel's relaxed semver format
   <https://bazel.build/external/module#version_format>`_.
 - The optional ``repository_url`` can be used to specify the URL of an
-  alternative registry.
+  alternative registry, with any trailing forward slashes removed.
 - The optional ``subpath`` can name a particular Bazel target in the module via
   a label with the leading double slash (``//``) removed and canonicalized by
-  omitting the target name if it is equal to the name of the containing package. 
+  omitting the target name if it is equal to the name of the containing package.
 - Examples::
 
       pkg:bazel/rules_java@7.8.0
--- a/PURL-TYPES.rst
+++ b/PURL-TYPES.rst
@@ -82,6 +82,7 @@ bazel
       pkg:bazel/curl@8.8.0?repository_url=https://example.org/bazel-registry
       pkg:bazel/rules_java@7.8.0
       pkg:bazel/rules_java@7.8.0#toolchains:singlejar
+      pkg:bazel/rules_go@0.48.0#go
 
 bitbucket
 ---------
--- a/test-suite-data.json
+++ b/test-suite-data.json
@@ -550,5 +550,89 @@
     "qualifiers": null,
     "subpath": null,
     "is_invalid": false
+  },
+  {
+    "description": "bazel module with default registry",
+    "purl": "pkg:bazel/rules_java@7.8.0?registry_url=https://bcr.bazel.build/",
+    "canonical_purl": "pkg:bazel/rules_java@7.8.0",
+    "type": "bazel",
+    "namespace": null,
+    "name": "rules_java",
+    "version": "7.8.0",
+    "qualifiers": null,
+    "subpath": null,
+    "is_invalid": false
+  },
+  {
+    "description": "bazel module with custom registry",
+    "purl": "pkg:bazel/rules_java@7.8.0?registry_url=https://example.org/bazel-registry/",
+    "canonical_purl": "pkg:bazel/rules_java@7.8.0?registry_url=https://example.org/bazel-registry",
+    "type": "bazel",
+    "namespace": null,
+    "name": "rules_java",
+    "version": "7.8.0",
+    "qualifiers": null,
+    "subpath": null,
+    "is_invalid": false
+  },
+  {
+    "description": "bazel module with relaxed semver version",
+    "purl": "pkg:bazel/curl@8.8.0.bcr.1",
+    "canonical_purl": "pkg:bazel/curl@8.8.0.bcr.1",
+    "type": "bazel",
+    "namespace": null,
+    "name": "curl",
+    "version": "8.8.0.bcr.1",
+    "qualifiers": null,
+    "subpath": null,
+    "is_invalid": false
+  },
+  {
+    "description": "bazel module names are lowercased",
+    "purl": "pkg:bazel/Curl@8.8.0.bcr.1",
+    "canonical_purl": "pkg:bazel/Curl@8.8.0.bcr.1",
+    "type": "bazel",
+    "namespace": null,
+    "name": "curl",
+    "version": "8.8.0.bcr.1",
+    "qualifiers": null,
+    "subpath": null,
+    "is_invalid": true
+  },
+  {
+    "description": "bazel module with target",
+    "purl": "pkg:bazel/rules_java@7.8.0#//toolchains:singlejar",
+    "canonical_purl": "pkg:bazel/rules_java@7.8.0#toolchains:singlejar",
+    "type": "bazel",
+    "namespace": null,
+    "name": "rules_java",
+    "version": "7.8.0",
+    "qualifiers": null,
+    "subpath": "toolchains:singlejar",
+    "is_invalid": false
+  },
+  {
+    "description": "bazel module with target in short-hand form",
+    "purl": "pkg:bazel/rules_go@0.48.0#//go:go",
+    "canonical_purl": "pkg:bazel/rules_go@0.48.0#go",
+    "type": "bazel",
+    "namespace": null,
+    "name": "rules_go",
+    "version": "0.48.0",
+    "qualifiers": null,
+    "subpath": "go",
+    "is_invalid": false
+  },
+  {
+    "description": "bazel module with top-level target",
+    "purl": "pkg:bazel/rules_go@0.48.0#//:stdlib",
+    "canonical_purl": "pkg:bazel/rules_go@0.48.0#:stdlib",
+    "type": "bazel",
+    "namespace": null,
+    "name": "rules_go",
+    "version": "0.48.0",
+    "qualifiers": null,
+    "subpath": ":stdlib",
+    "is_invalid": false
   }
 ]


fmeum:
CC @mzeren-vmw

oej:
Any status on the feedback you waited for?

fmeum:
@oej Yes, this has been approved and is ready for review!

fmeum:
@pombredanne Not sure who to ask for a review, could you take a look?

sschuberth:
@fmeum please rebase to resolve conflicts.

fmeum:
@sschuberth Done

PURL-TYPES.rst
```diff
@@ -59,6 +59,24 @@ apk
       pkg:apk/alpine/curl@7.83.0-r0?arch=x86
       pkg:apk/alpine/apk@2.12.9-r3?arch=x86
 
+bazel
+-----
+``bazel`` for Bazel modules:
+
+- The default repository ("registry") is ``https://bcr.bazel.build``, the
```
sschuberth, line 66: This forwards to https://registry.bazel.build/. So should the URL be changed here?

PURL-TYPES.rst
```diff
@@ -59,6 +59,24 @@ apk
...
+  Bazel Central Registry (BCR).
+- The ``namespace``, if specified, is the host and path of the URL of an
+  alternative registry. It is not case sensitive and must be lowercased.
```
sschuberth: The "must be lowercased" part does not match the generic instructions [how to build a purl](https://github.com/package-url/purl-spec/blob/master/PURL-SPECIFICATION.rst#how-to-build-purl-string-from-its-components), so I think it should be dropped here. IMO it's not feasible for implementers to add special handling for Bazel here.

PURL-TYPES.rst
```diff
@@ -59,6 +59,24 @@ apk
...
+- The ``name`` is the module name. It is case sensitive, but a single registry
+  will never contain two different module names that compare equal
+  case-insensitively.
+- The ``version`` is the module version.
+- Examples::
+
+      pkg:bazel/rules_java@7.8.0
+      pkg:bazel/curl@8.8.0.bcr.1
+      pkg:bazel/example.org/bazel-registry/curl@8.8.0
```
sschuberth: Please add an example with a non-default registry.

PURL-TYPES.rst
fmeum: https://bcr.bazel.build is indeed the default registry (see https://github.com/bazelbuild/bazel/blob/a298190ff22435f32dc3a3ddc6507ddc72e0e503/src/main/java/com/google/devtools/build/lib/bazel/BazelRepositoryModule.java#L139-L140) whereas https://registry.bazel.build offers a UI for the hosted modules. We recently added redirects from bcr.bazel.build to registry.bazel.build for those requests that a registry consumer wouldn't make.

For example, https://bcr.bazel.build/modules/rules_go/0.50.1/MODULE.bazel is what Bazel would download and doesn't redirect, but https://bcr.bazel.build/modules/rules_go/0.50.1 redirects to https://registry.bazel.build/modules/rules_go/0.50.1 as a convenience for humans.

PURL-TYPES.rst
fmeum: See above, should now be fixed in the remaining single commit.

PURL-TYPES.rst
sschuberth: Ah, my bad, for some reason I seem to have stopped reviewing at the first commit. Anyway, good to have the fixups squashed now.

PURL-TYPES.rst
```diff
@@ -59,6 +59,29 @@ apk
...
+-----
+``bazel`` for Bazel modules:
+
+- The default repository ("registry") is ``https://bcr.bazel.build``, the
+  Bazel Central Registry (BCR).
+- The ``name`` is the module name. It must be lowercased.
+- The ``version`` is the module version in `Bazel's relaxed semver format
+  <https://bazel.build/external/module#version_format>`_.
+- The optional ``repository_url`` can be used to specify the URL of an
+  alternative registry, with any trailing forward slashes removed.
```
sschuberth, line 72: > with any trailing forward slashes removed.

I'm not sure about that bit. This is not one of the "type-specific normalizations" that are allowed for the namespace segments and name. And semantically, having a trailing slash or not does not make a difference for the URL.

PURL-TYPES.rst
```diff
@@ -59,6 +59,29 @@ apk
...
+- The optional ``subpath`` can name a particular Bazel target in the module via
+  a label with the leading double slash (``//``) removed and canonicalized by
+  omitting the target name if it is equal to the name of the containing package.
```
sschuberth, line 75: Hmm. The [spec](https://github.com/package-url/purl-spec/blob/master/PURL-SPECIFICATION.rst#package-url-specification-v10x) describes the `subpath` as

> extra subpath within a package, relative to the package root.

How is what you describe "relative to the package root"? Would a custom qualifier maybe make more sense to store the Bazel target?

fmeum: I simplified this part of the spec in a new commit so that it's always a regular file path, corresponding to a package (not target or file) in the Bazel module. This is analogous to the usage of `subpath` for `golang`.

sschuberth: Looks good to me now, but I'd like others to share their opinion as well.

Yannic (approved):
LGTM

What's needed to get this merged?

sschuberth:
> What's needed to get this merged?

Two approvals are needed since recently. Maybe @pombredanne can also review?

jkowalleck (approved):

Yannic:
@sschuberth I think this now has the second approval?

sschuberth:
Unfortunately, @Yannic has read-only permissions, so the approval does not count towards mergability:

image

sschuberth:
> @sschuberth Do you happen to have write permission and could add the second approval yourself?

@fmeum As mentioned over [here](https://github.com/package-url/packageurl-go/pull/76#issuecomment-2507737408) I'm currently pausing reviews to this project until some process question are clarified with @pombredanne.

Yannic:
> Unfortunately, @Yannic you have read-only permissions, so the approval does not count towards mergability:
> 
> image
> 
> image

Oh! Sorry, I was under the impression that you already approved. I'm aware that my LGTM doesn't count towards the two required approvals.

@sschuberth do you have a rough estimate on when you expect the process questions to be resolved? A few days? A few weeks? I'm asking mostly to understand when Fabian or me should follow-up here so we don't unnecessarily ping :)

sschuberth:
> @sschuberth do you have a rough estimate on when you expect the process questions to be resolved? A few days? A few weeks?

I simply have no idea. I've already reached out the respective people, but I'm not getting any answer. But I'll keep on pushing 😉

fmeum:
@stevespringett @shibumi @johnmhoran @pombredanne Apologies for the multi-ping, but it's hard to tell who would be able to move this forward. Could you add a second review?

shibumi:
@fmeum Hi, I am "only" the packageurl-go maintainer and have not much to say when it comes to purl specification issues. I am afraid you will have to wait for @pombredanne or anyone else of the "steering committee".

PURL-TYPES.rst
```diff
@@ -59,6 +59,28 @@ apk
       pkg:apk/alpine/curl@7.83.0-r0?arch=x86
       pkg:apk/alpine/apk@2.12.9-r3?arch=x86
 
+bazel
+-----
+``bazel`` for Bazel modules:
+
+- The default repository ("registry") is ``https://bcr.bazel.build``, the
+  Bazel Central Registry (BCR).
```
pombredanne, line 67:
```suggestion
```

test-suite-data.json
```diff
@@ -622,5 +622,77 @@
...
+    "name": "curl",
+    "version": "8.8.0.bcr.1",
+    "qualifiers": null,
+    "subpath": null,
+    "is_invalid": false
+  },
+  {
+    "description": "bazel module names are lowercased",
+    "purl": "pkg:bazel/Curl@8.8.0.bcr.1",
+    "canonical_purl": "pkg:bazel/Curl@8.8.0.bcr.1",
```
pombredanne, line 665:
```suggestion
    "canonical_purl": "pkg:bazel/curl@8.8.0.bcr.1",
```

test-suite-data.json
```diff
@@ -622,5 +622,77 @@
...
+    "type": "bazel",
+    "namespace": null,
+    "name": "curl",
+    "version": "8.8.0.bcr.1",
+    "qualifiers": null,
+    "subpath": null,
+    "is_invalid": true
```
pombredanne, line 672: I am  not sure we want this as invalid. I think that the canonical purl name should be lowercased as part canonicalization.

PURL-TYPES.rst
```diff
@@ -59,6 +59,28 @@ apk
...
+- The ``name`` is the module name. It must be lowercased.
```
pombredanne, line 68: Could it make sense to integrate the whole name spec at https://bazel.build/rules/lib/globals/module#module.name ?
> A valid module name must: 1) only contain lowercase letters (a-z), digits (0-9), dots (.), hyphens (-), and underscores (_); 2) begin with a lowercase letter; 3) end with a lowercase letter or digit. 

 Also, about names, I guess that when there is name as possible in the spec, there is no purl possible, right?
> The name of the module. Can be omitted only if this module is the root module (as in, if it's not going to be depended on by another module).

pombredanne (changes requested):
Thanks. Looking good overall! :heart:  It would be great to add at a link to the spec for modules at https://bazel.build/external/module and I have posted a few small nits, suggestions and questions for you consideration.

test-suite-data.json
jkowalleck: good catch.
according to proposed spec, "name"-part must be lowercased.

PURL-TYPES.rst
matt-phylum: Validating package names via the PURL spec makes PURL and its implementations more complicated and leads to a PURL spec crisis if the real validation rules are relaxed in the future (eg `@a/b` was invalid before NPM 2.0) where people will argue that it's a breaking change to allow something that was previously forbidden in PURL. I understand the value of canonicalizing the package name so multiple distinct PURLs don't refer to the same package, and that should be safe because changing canonicalization rules would break builds, but I don't think there's a need for PURL implementations in general to validate the name of the package to prevent PURLs that cannot refer to a package.

Are Bazel module names case insensitive, or are uppercase characters currently invalid? There's a difference, and some existing PURL types get this wrong (eg NPM).

fmeum: Bazel module names are enforced to be of the form @pombredanne linked. I agree that keeping validation out of the spec makes sense. We could just allow PURLs to reference module names that aren't valid as per the current rules enforced by Bazel, they just won't be useful.

test-suite-data.json
fmeum: Should I also keep casing out of canonicalization? `pkg:bazel/Curl@8.8.0.bcr.1` could be a valid canonical PURL, just not one for a module Bazel would ever accept.

PURL-TYPES.rst
fmeum: I dropped validation from the type spec and the test suite, PTAL.

jkowalleck (approved):

Yannic (approved):
Thanks!

Can we add a test case for valid purls that Bazel will reject (e.g., `pkg:bazel/Curl@1234`) to have coverage for the non-validation case?

fmeum:
> Can we add a test case for valid purls that Bazel will reject (e.g., `pkg:bazel/Curl@1234`) to have coverage for the non-validation case?

Added

Yannic (approved):

test-suite-data.json
```diff
@@ -622,5 +622,77 @@
...
+    "type": "bazel",
+    "namespace": null,
+    "name": "curl",
+    "version": "8.8.0.bcr.1",
+    "qualifiers": null,
+    "subpath": null,
+    "is_invalid": false
+  },
+  {
+    "description": "bazel module names are currently all lowercased, but this is not enforced by the PURL type spec",
```
jkowalleck, line 663: Can we express, that the type-spec does not enforce lowercasing, not by accident but for reasons?

```suggestion
    "description": "bazel module names are actually all lowercased, but this is not enforced by the PURL type spec for reasons",
```

jkowalleck (changes requested):

jkowalleck (approved):

Yannic (approved):

PURL-TYPES.rst
```diff
@@ -59,6 +59,27 @@ apk
...
       pkg:apk/alpine/apk@2.12.9-r3?arch=x86
 
+bazel
+-----
+``bazel`` for Bazel modules as specified at ``https://bazel.build/external/module``:
+
+- The default repository is the Bazel Central Registry (BCR) ``https://bcr.bazel.build``.
+- The ``name`` is the module name.
+- The ``version`` is the module version in `Bazel's relaxed semver format
+  <https://bazel.build/external/module#version_format>`_.
```
pombredanne, line 69:
```suggestion
```

jkowalleck (dismissed):
LGTM

I really like the way it is set up now.
all the "external" bazel rules are left out, and we stay with just the bare mapping of purl components.
This way the type definition is functional, but not over-specific and free to evolve with bazel's needs.

PURL-TYPES.rst
```diff
@@ -59,6 +59,26 @@ apk
...
       pkg:apk/alpine/apk@2.12.9-r3?arch=x86
 
+bazel
+-----
+``bazel`` for Bazel modules as specified at ``https://bazel.build/external/module``:
+
+- The default repository is the Bazel Central Registry (BCR) ``https://bcr.bazel.build``.
+- The ``name`` is the module name.
+- The ``version`` is the module version.
+- The optional ``repository_url`` can be used to specify the URL of an
```
pombredanne, line 69: This is a standard qualifier and option, so there is no need to repeat this here

PURL-TYPES.rst
```diff
@@ -59,6 +59,26 @@ apk
...
+  alternative registry.
+- The optional ``subpath`` can refer to a particular Bazel package in the
```
pombredanne, line 71: Can you clarify for me what a package means here? From https://bazel.build/concepts/build-ref#packages I get that this is the path to a directory relative to the root of a module, where the directory must contain a BUILD.bazel file and the root of the module a MODULE.bazel file. Is this correct? Is there something we should add to the definition?

PURL-TYPES.rst
```diff
@@ -59,6 +59,26 @@ apk
       pkg:apk/alpine/curl@7.83.0-r0?arch=x86
       pkg:apk/alpine/apk@2.12.9-r3?arch=x86
 
+bazel
+-----
+``bazel`` for Bazel modules as specified at ``https://bazel.build/external/module``:
+
+- The default repository is the Bazel Central Registry (BCR) ``https://bcr.bazel.build``.
+- The ``name`` is the module name.
```
pombredanne, line 67: @fmeum In most case-insensitive types, we state this:
`The name is not case sensitive and must be lowercased.` 
@Yannic I reckon the (minor) hardship on type-specific cae normalization, but we already do it everywhere else, and this help using a PURL consistently as a key in databases.

```suggestion
- The ``name`` is the module name and must be lowercased.
```

pombredanne (changes requested):
@fmeum Thanks for the updates. Here are some final nits for your consideration!

PURL-TYPES.rst
fmeum: That's pretty much it (`BUILD` files are also allowed). It's one of the top two core concepts of Bazel but is also indistinguishable from a "path from the root", so I'm not sure whether additional details should be added to the spec. What would you be looking for?

PURL-TYPES.rst
fmeum: I pushed an update, is that what you had in mind? Please note that Bazel module names are not case insensitive in the literal sense, it's an error to request a module with a name that isn't lowercased. But I guess that applies to other schemes as well and normalizing them is simpler than marking them invalid?

jkowalleck: > @fmeum In most case-insensitive types, we state this: `The name is not case sensitive and must be lowercased.` @Yannic I reckon the (minor) hardship on type-specific cae normalization, but we already do it everywhere else, and this help using a PURL consistently as a key in databases.

I disagree. 
we'vediscussesd this already, andlearned, that the name's case is highlyimportant.
therefore, i woudl rather go with the completely opposite declaration:
> - The ``name`` is case sensitive and MUST NOT be lowercased.

fmeum: > The name is case sensitive and MUST NOT be modified.

I like that this explicitly doesn't mention what the casing should be (that's left to the domain), just that it is important.

fmeum: @jkowalleck I updated spec and test case, PTAL.

jkowalleck: > > The name is case sensitive and MUST NOT be modified.
> 
> I like that this explicitly doesn't mention what the casing should be (that's left to the domain), just that it is important.

exactly.



In the end, I would really prefer to not mention any normalization or rule here.
this pins purl in a place where it has no authority.
the package ecosystemsmayevolve and chaneg their rules, and wewould (again) end upwith incompatible purl types and endless discussions.

jkowalleck (approved):

PURL-TYPES.rst
pombredanne: I needed to understand. I want to make sure that `can refer to a particular Bazel package in the module` is not ambiguous.

pombredanne (changes requested):

PURL-TYPES.rst
```diff
@@ -59,6 +59,24 @@ apk
       pkg:apk/alpine/curl@7.83.0-r0?arch=x86
       pkg:apk/alpine/apk@2.12.9-r3?arch=x86
 
+bazel
+-----
+``bazel`` for Bazel modules as specified at ``https://bazel.build/external/module``:
+
+- The default repository is the Bazel Central Registry (BCR) ``https://bcr.bazel.build``.
+- The ``name`` is case sensitive and must not be modified.
```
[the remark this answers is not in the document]
fmeum: Let me start by saying that I don't particularly mind how this is formulated. I'm just interested in getting the purl type registered.

The name isn't case sensitive in the sense that for each string, there is only ever at most one Bazel module name that is equal to it ignoring case.
The name is case sensitive in the sense that specifying a module name with any casing other than lower case will result in an error shown by Bazel.

PURL-TYPES.rst
jkowalleck: bazil is excplic it about caseing: it is case-sensitive, and no normalization is implied.


"name must be lowercased" -> meaning name is case-sensitive and Uppercase is forbidden today. bazil does not know any normalization, and does at no point imply that it will never evolve.
bazil is the authority, not purl. dont make the same mistakes again and start normalization where none is planned.

to contrast this: pypi clearly states how normalization work.

PURL-TYPES.rst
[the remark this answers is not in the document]
matt-phylum: I agree that it is case sensitive in that case.

If you can use Bazel to install SHOUTING_MODULE and it installs shouting_module, then it is case insensitive. However, if installing SHOUTING_MODULE is an error, it's possible that in the behavior will change or has changed, at which point PURL implementations will be automatically converting one package name into another package name like what happens with [`pkg:npm/jQuery`](https://www.npmjs.com/package/jQuery) which gets replaced by [`pkg:npm/jquery`](https://www.npmjs.com/package/jquery) because the _current_ rules forbid uppercase and past rules did not.

Besides the possibility that rule changes cause aliasing problems, it just doesn't seem like a good or useful idea to specify that if the user provides an invalid module name the PURL implementation must automatically convert it into a valid module name.

PURL-TYPES.rst
Yannic: I agree with what @jkowalleck and @fmeum said above:

Bazel module names are case sensitive. `foo`, `FOO`, and `FoO` are all separate modules that could theoretically exist side by side. The fact that there are currently no such modules does not matter.

I agree with the part that `foo` and `FOO` are unlikely to ever exist at the same time, but I could very well see `foo` and `BAR` both existing with exactly this casing.

IMO, `purl` should uniquely identify a package, and I should be able to take the `name`, go to the package manager, and get to the right package. Normalization breaks this invariant, since `bar` from the example above is going to break this.

I like `The name is case sensitive and MUST NOT be modified.` as phrasing

Yannic: Note that different casings can also happen if private registries are involved: public BCR may happen to contain `foo`, which the private registry at $company mirrors, but there can also be an additional module `FOO` in the private registry

PURL-TYPES.rst
fmeum: I can clarify this, but my understanding is that subpath should be a valid file path for all purls and a label with prefix `//` is not. What do you think?

matt-phylum: `//foo/bar` is not a valid subpath. PURL implementations are required to read `//foo/bar` as `foo/bar`.

Yannic: Meh...

Should we at least add a link to https://bazel.build/concepts/labels#package-names then? Even if we don't enforce the restrictions

fmeum: Sure, added the link

jkowalleck (dismissed):

PURL-TYPES.rst
fmeum: @pombredanne @jkowalleck Can we move this forward somehow? The Bazel ecosystem keeps building on this scheme out of necessity and it would be really helpful to make it official.

PURL-TYPES.rst
```diff
@@ -59,6 +59,24 @@ apk
...
+- The ``version`` is the module version.
+- The optional ``subpath`` can refer to a particular Bazel package in the
+  module (``https://bazel.build/concepts/labels#package-names``).
```
Yannic: I've been working with `pkg:bazel` `purl`s quite a bit over the last few months and am wondering whether the package name is too restrictive. In particular, I have a couple of cases where there would be a benefit to specifying a target here (e.g., in case a specific target has a different license).

@fmeum WDYT about allowing labels as `subpath` here? Something like `The optional "subpath" **MAY** refer to a Label of a particular package or target in the module. When referring to targets, the Label **MUST** include the name of the target, separated from the package by ":". If there is no target name, "subpath" is assumed to refer to the whole package. The leading "//" of the Label **MUST** be omitted.`

It's not something we need to solve in this PR, but I think it's something we'll definitely need at some point. Given how long this has been pending in rounds of reviews, I'm leaning towards doing this in a follow-up.

Yannic (approved):
Having worked on https://github.com/bazel-contrib/supply-chain for the last few month, I'd really like to see this land sooner than later.

PURL-TYPES.rst
fmeum: I like the idea and the phrasing. Since it seems pretty uncontentious (the syntax exactly matches the Bazel syntax for a repo-relative label specified on the command line), we can also add it right away. I'll wait for reviewers to chime in first though, maybe they prefer to merge this as is.

pombredanne:
- See  #514

... therefore with the new approach... this PR would need to be updated. 


Do you think you can update this PR to the new format?

Sorry for the churn. :heart:

fmeum:
@pombredanne PTAL

fmeum:
@jkowalleck Friendly ping, what is needed to get this merged?

jkowalleck (approved):

fmeum:
@pombredanne Friendly ping

fmeum:
@mjherzog Could you take a look?

pombredanne (approved):
@fmeum Thanks for the refinements and efforts. This looks great. Let's merge this!

PR #318, opened 2024-08-07
Add new PURL types: LabVIEW packages 'nipkg' and 'vip'

### LINKED ISSUE #202: Reserve LabVIEW-related package formats vip / nipm
I request to add nipm and vip as PURL types to support the generation of SBOM for LabVIEW-related package formats.

nip = NI Package Manager packages (https://www.ni.com/en-gb/support/downloads/ni-package-manager.html)
vip = VI Packages (https://www.vipm.io/)

These are the two main package formats used by LabVIEW developers.

samsharp99:
* Added pURL types for NIPKG (NI Package Format) and VIP (VIPM VI Package) for common LabVIEW package formats
#202

### DIFF
--- a/PURL-TYPES.rst
+++ b/PURL-TYPES.rst
@@ -428,7 +428,19 @@ mlflow
       pkg:mlflow/creditfraud@3?repository_url=https://westus2.api.azureml.ms/mlflow/v1.0/subscriptions/a50f2011-fab8-4164-af23-c62881ef8c95/resourceGroups/TestResourceGroup/providers/Microsoft.MachineLearningServices/workspaces/TestWorkspace
       pkg:mlflow/trafficsigns@10?model_uuid=36233173b22f4c89b451f1228d700d49&run_id=410a3121-2709-4f88-98dd-dba0ef056b0a&repository_url=https://adb-5245952564735461.0.azuredatabricks.net/api/2.0/mlflow
 
+nipkg
+-----
+``nipkg`` for NI Package format packages:
 
+- The ``namespace`` is the maintainer from the package attributes (optional).
+- The ``name`` is the package name from from the package attributes.
+- The ``version`` is the full version from the package attributes - not the display version.
+- Examples::
+
+      pkg:nipkg/foobar@12.3.1
+      pkg:nipkg/ni-visa-17.5-dotnet-runtime
+	  pkg:nipkg/National%20Instruments/ni-labview-2019-runtime-engine-x86@19.1.4.49152-0+f0
+	  
 npm
 ---
 ``npm`` for Node NPM packages:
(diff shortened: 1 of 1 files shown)


PURL-TYPES.rst
```diff
@@ -428,7 +428,19 @@ mlflow
       pkg:mlflow/creditfraud@3?repository_url=https://westus2.api.azureml.ms/mlflow/v1.0/subscriptions/a50f2011-fab8-4164-af23-c62881ef8c95/resourceGroups/TestResourceGroup/providers/Microsoft.MachineLearningServices/workspaces/TestWorkspace
       pkg:mlflow/trafficsigns@10?model_uuid=36233173b22f4c89b451f1228d700d49&run_id=410a3121-2709-4f88-98dd-dba0ef056b0a&repository_url=https://adb-5245952564735461.0.azuredatabricks.net/api/2.0/mlflow
 
+nipkg
+-----
+``nipkg`` for NI Package format packages:
+
+- The ``namespace`` is the maintainer from the package attributes (optional).
```
matt-phylum, line 435: What does it mean for the maintainer to be optional? Is it optional because the maintainer has a default?

If it's optional because it doesn't matter, it shouldn't be included because that creates problems for code that needs to match PURLs.

PURL-TYPES.rst
```diff
@@ -577,6 +589,20 @@ swift
 
       pkg:swift/github.com/Alamofire/Alamofire@5.4.3
       pkg:swift/github.com/RxSwiftCommunity/RxFlow@2.12.4
+	  
+vip
+---
+``vip`` for VI Package Manager (VIPM) format packages (LabVIEW):
+
+- The ``namespace`` is the company name from the package attributes (optional).
```
matt-phylum, line 597: It doesn't seem like the namespace is used by the native vipm tools (the website makes links like `vipm://oglib_numeric`), so it probably should not be included here.

PURL-TYPES.rst
samsharp99: The maintainer is an optional field in the nipkg attributes, so it could be empty. In the pURL spec it should be provided, if present in the nipkg attributes, but otherwise ommitted - hence being optional.

PURL-TYPES.rst
samsharp99: Updated to remove namespace in commit:
https://github.com/package-url/purl-spec/pull/318/commits/c193d21da453507a4d9f189765623778035aacc1

samsharp99:
Clarified nipkg namespace attribute inclusion and removed vip namespace, updated PR from master.

pombredanne:
- See  #514

... therefore with the new approach... this PR would need to be updated. 


Do you think you can update this PR to the new format?

Sorry for the churn. :heart:

jimkring:
Hi @samsharp99, @pombredanne. I'm on the [VI Package Manager (VIPM)](https://docs.vipm.io/) team (along with @Tom-McQuillan) and we're working on a plan for better SBOM support for VIPM's packages (.vip and legacy .ogp file extensions) and also NIPM's packages (.nipkg file extension) -- we have a configuration management solution for this VIPM+NIPM packaging space.  Internally, we've been discussing the `purl` type/namespace for these two package management protocols.

From the [README.md](https://github.com/package-url/purl-spec/blob/main/README.md#purl)
> type: the package "type" or package "protocol" such as maven, npm, nuget, gem, pypi, etc. Required.

Based on that, and also checking out [the purl examples](https://github.com/package-url/purl-spec/blob/main/README.md#some-purl-examples), I would suggest going with the package "protocol" (`vipm` and `nipm`) moreso than the package file "type" (`vip`, `ogp` and `nipkg`). 

Here are some of those examples, I'm referring to:

```
pkg:pypi/django@1.11.1
pkg:rpm/opensuse/curl@7.56.1-1.1.?arch=i386&distro=opensuse-tumbleweed
```

Here is what I'm proposing/thinking for VIPM and NIPM packages:

```
pkg:vipm/oglib_array@6.0.0.9
pkg:nipm/ni-daqmx@23.8.0.49349-0+f197?repository_url=download.ni.com/support/nipkg/products/ni-d/ni-daqmx/23.8/released
```

Let me know how I and @Tom-McQuillan can help support this effort. I appreciate all the great thought and efforts you've put into this, @samsharp99.  Cheers.

Bas-vE:
Whats the status on this?

mjherzog:
@Bas-vE The status is that this PR needs to be updated to follow the new JSON schema format and provide at least some base tests. The information provided to-date seems to be pretty straightforward. It also appears that there is no public code download - only via login at https://www.ni.com/en/support/downloads/software-products/download.labview.html#585933. Is that correct?

Bas-vE:
LabVIEW can only be downloaded via login, it's a proprietary programming language and IDE.

@Tom-McQuillan Did you implement this the way Jim has proposed in LV2026Q3?

PR #321, opened 2024-08-09
Add new PURL type:  'mendix'

MxMarcel:
Please consider adding the package type we use in our SBOM to be added to the list of known types.

The Mendix platform is a low-code platform with its own package system. Mendix packages are imported into Mendix Apps, or used as IDE extensions and hosted on the Mendix Marketplace , see https://marketplace.mendix.com/. A SBOM is created for the App which used package urls to list the components in the App.

We will add more details about the other elements like qualifiers that we use at a later point.

### DIFF
--- a/PURL-TYPES.rst
+++ b/PURL-TYPES.rst
@@ -607,6 +607,7 @@ Other candidate types to define:
 - ``helm`` for Kubernetes packages
 - ``julia`` for Julia packages:
 - ``melpa`` for Emacs packages
+- ``mendix`` for Mendix packages
 - ``meteor`` for Meteor JavaScript packages:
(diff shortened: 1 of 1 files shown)


matglas:
What is the further process to get types fully included? If there is still something missing it would be great to know about it.

pombredanne:
@matglas  I am sorry for the time it takes, we are working to improve ... but after the merge of PR #514, PURL tests and defs are now defined in new JSON schema :innocent: :grin: :
- See  #514

With the new approach... this PR needs to be updated. Sorry for the churn.  :heart:

matglas:
@pombredanne thanks for the update! We will pick it up!

PR #343, opened 2024-11-13
Introduce ecosystem-independent all and none schemes

immqu:
This PR introduces the possibility to specify _all_ versions, or _none_, as suggested in #267 and implemented in https://github.com/aboutcode-org/univers/pull/146

### DIFF
--- a/VERSION-RANGE-SPEC.rst
+++ b/VERSION-RANGE-SPEC.rst
@@ -240,10 +240,11 @@ A list of versions, enumerated:
 
 - ``vers:pypi/0.0.0|0.0.1|0.0.2|0.0.3|1.0|2.0pre1``
 
-Additionally, all versions (or none) SHOULD be specified:
+Also, it is possible to specify a range that contains all versions or none:
 
 - ``vers:all/*``
 - ``vers:none/*``
+
 
 A complex statement about a vulnerability in a "maven" package that affects
 multiple branches each with their own fixed versions at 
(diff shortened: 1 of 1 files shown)


VERSION-RANGE-SPEC.rst
```diff
@@ -240,6 +240,10 @@ A list of versions, enumerated:
 
 - ``vers:pypi/0.0.0|0.0.1|0.0.2|0.0.3|1.0|2.0pre1``
 
+Additionally, all versions (or none) can be specified:
```
tschmidtb51, line 243: I think we should be more strict.
```suggestion
Additionally, all versions (or none) SHOULD be specified:

Comment (tschmidtb51) [reply VERSION-RANGE-SPEC.rst:243]: @pombredanne Thoughts?

File: VERSION-RANGE-SPEC.rst [hunk 2]
```diff
@@ -637,6 +641,8 @@ These are a few known versioning schemes for some common Package URL
 - **alpine**: Alpine linux https://gitlab.alpinelinux.org/alpine/apk-tools/-/blob/master/src/version.c
   which is using Gentoo-like conventions.
 
+- **ecosystem-independent versioning schemes**: Several versioning schemes that are independent from any specific ecosystem are being developed. A very simple one is the generic *all* and *none* scheme which allows to specify that all (or none) of the versions are affected.
```
tschmidtb51, line 644: That shouldn't be part of the list of "some common Package URL `types`".

Moreover, we are not talking about `affected` or `not affected`. `vers` is solely about "is it in the range or not". 
```suggestion

Moreover, there are several versioning schemes that are independent from any specific ecosystem.

- **all**: a generic versioning scheme where every version is inside the range, i.e. all versions match. This is equivalent to `true`.
- **none**: a generic versioning scheme where no version is inside the range, i.e. no versions match. This is equivalent to `false`.

Comment (tschmidtb51) [reply VERSION-RANGE-SPEC.rst:644]: Also, we should sort this entries...

**Review by tschmidtb51 — CHANGES_REQUESTED** [review]
Please see my comments

File: VERSION-RANGE-SPEC.rst [hunk 3]
```diff
@@ -240,6 +240,10 @@ A list of versions, enumerated:
 
 - ``vers:pypi/0.0.0|0.0.1|0.0.2|0.0.3|1.0|2.0pre1``
 
+Additionally, all versions (or none) can be specified:
+
+- ``vers:all/*``
+- ``vers:none/*``
```
tschmidtb51, line 246: Also, we need to state somewhere that `*` is the only valid version constraint these versioning schemes can have and that no other construct is valid (even not a space in between).

immqu: Currently, the document says ```A separate document will provide details for each versioning scheme```. Is someone already taking care of this separate document?

tschmidtb51: I'm not aware of that effort.
@pombredanne Any insights in this?

VERSION-RANGE-SPEC.rst
```diff
@@ -624,31 +638,22 @@ These are a few known versioning schemes for some common Package URL
...
+TODO: add Rust, composer and archlinux, nginx, tomcat, apache.
 
-- **gentoo**: Gentoo https://wiki.gentoo.org/wiki/Version_specifier
 
-- **alpine**: Alpine linux https://gitlab.alpinelinux.org/alpine/apk-tools/-/blob/master/src/version.c
-  which is using Gentoo-like conventions.
+Moreover, there are several versioning schemes that are independent from any specific ecosystem.
 
-- **ecosystem-independent versioning schemes**: Several versioning schemes that are independent from any specific ecosystem are being developed. A very simple one is the generic *all* and *none* scheme which allows to specify that all (or none) of the versions are affected.
+- **all**: a generic versioning scheme where every version is inside the range, i.e., all versions match. This is equivalent to `true`.
```
tschmidtb51, line 649: Just saw: This repo uses double backticks instead of single ones.

immqu:
@pombredanne can this PR be merged or are there any open points to discuss?

pombredanne (changes requested):
Thanks, please remove the unrelated changes to another PR for simplicity.

immqu:
I have removed the sorting and will add it in another PR. It should be easily reviewable now.

PR #351, opened 2024-11-22
Replace the `LICENSE` link with `mit.LICENSE`

sschuberth:
This makes the license text properly show up at [1]. Apparently GitHub's logic is not able to follow symbolic links here.

[1]: https://github.com/package-url/purl-spec?tab=License-1-ov-file#readme


johnmhoran:
Why do we need the `LICENSE` file at all?  Isn't the `mit.LICENSE` file sufficient by itself?

ppkarwasz:
> Why do we need the `LICENSE` file at all? Isn't the `mit.LICENSE` file sufficient by itself?

For humans yes, but GitHub automation is not able to determine the kind of license used in this repository. If you look at the "About" box in the [main page](https://github.com/package-url/purl-spec), the license is classified as a custom license ("View license") instead of MIT.

sschuberth:
> Why do we need the `LICENSE` file at all? Isn't the `mit.LICENSE` file sufficient by itself?

IMO it should rather be the opposite: *If* only keeping one file, it should be named `LICENSE`, as that's more common.

johnmhoran (approved):
@ppkarwasz @sschuberth Thank you for your replies -- LGTM!

pombredanne:
The tyranny of github on license file names is troubling, especially when their detection is so poor. GitHub may have been  able to follow symlinks? .. but no longer.
Let me merge this

pombredanne:
Merged with https://github.com/package-url/purl-spec/commit/ea4bc0c02940ea51dc1468da1abf9780b0654eb6

Thanks!

PR #361, opened 2024-12-05
Clarify encoding of colon between scheme and type

pombredanne:
The colon separator between scheme and type must not be encoded.

This is also using, a new, and not yet adopted wording for MUST according to its definition in RFC2119 as clarified in RFC8174

Suggested-by: @gernot-h
Reference: https://github.com/package-url/purl-spec/issues/39#issuecomment-481422744

JMH: Reference: https://github.com/package-url/purl-spec/issues/376

### DIFF
--- a/PURL-SPECIFICATION.rst
+++ b/PURL-SPECIFICATION.rst
@@ -247,8 +247,9 @@ Use these rules for percent-encoding and decoding ``purl`` components:
 - the '#', '?', '@' and ':' characters must NOT be encoded when used as
   separators. They may need to be encoded elsewhere
 
-- the ':' ``scheme`` and ``type`` separator does not need to and must NOT be encoded.
-  It is unambiguous unencoded everywhere
+- The colon ':' separator between ``scheme`` and ``type`` MUST NOT be encoded.
+  For example, in the PURL snippet ``pkg:npm`` the colon ':' MUST NOT be encoded,
+  and the PURL snippet ``pkg%3Anpm`` is invalid.
 
 - the '/' used as ``type``/``namespace``/``name`` and ``subpath`` segments separator
   does not need to and must NOT be percent-encoded. It is unambiguous unencoded
--- a/PURL-SPECIFICATION.rst
+++ b/PURL-SPECIFICATION.rst
@@ -123,8 +123,8 @@ The rules for each component are:
 - **scheme**:
 
   - The ``scheme`` is a constant with the value "pkg".
-  - The scheme and type MUST be separated by a colon ':'.
-  - ``purl`` parsers MUST accept URLs in which the scheme and colon ':' are
+  - The ``scheme`` and ``type`` MUST be separated by a colon ':'.
+  - ``purl`` parsers MUST accept URLs in which the ``scheme`` and colon ':' are
     followed by one or more slash '/' characters, such as 'pkg://', and MUST
     ignore -- i.e., normalize by removing -- all such '/' characters.
 
--- a/PURL-SPECIFICATION.rst
+++ b/PURL-SPECIFICATION.rst
@@ -220,7 +220,7 @@ Character encoding
 
 For clarity and simplicity a ``purl`` is always an ASCII string. To ensure that
 there is no ambiguity when parsing a ``purl``, separator characters and non-ASCII
-characters MUST be UTF-encoded and then percent-encoded as defined at::
+characters must be UTF-encoded and then percent-encoded as defined at::
 
     https://en.wikipedia.org/wiki/Percent-encoding
 
--- a/PURL-SPECIFICATION.rst
+++ b/PURL-SPECIFICATION.rst
@@ -123,10 +123,10 @@ The rules for each component are:
 - **scheme**:
 
   - The ``scheme`` is a constant with the value "pkg".
-  - The ``scheme`` and ``type`` MUST be separated by a colon ':'.
-  - ``purl`` parsers MUST accept URLs in which the ``scheme`` and colon ':' are
+  - The ``scheme`` MUST be followed by an unencoded colon ':'.
+  - ``purl`` parsers MUST accept URLs where the ``scheme`` and colon ':' are
     followed by one or more slash '/' characters, such as 'pkg://', and MUST
-    ignore -- i.e., normalize by removing -- all such '/' characters.
+    ignore and remove all such '/' characters.
 
 
 - **type**:
@@ -218,10 +218,9 @@ The rules for each component are:
 Character encoding
 ~~~~~~~~~~~~~~~~~~
 
-For clarity and simplicity, a ``purl`` is always an ASCII string. To ensure that
-there is no ambiguity when parsing a ``purl``, unless otherwise provided in
-this specification, separator characters and non-ASCII characters MUST be
-UTF-encoded and then percent-encoded as defined at::
+For clarity and simplicity a ``purl`` is always an ASCII string. To ensure that
+there is no ambiguity when parsing a ``purl``, separator characters and non-ASCII
+characters MUST be UTF-encoded and then percent-encoded as defined at::
 
     https://en.wikipedia.org/wiki/Percent-encoding
 
@@ -229,13 +228,11 @@ Use these rules for percent-encoding and decoding ``purl`` components:
 
 - the ``type`` must NOT be encoded and must NOT contain separators
 
-- the '#', '?', '@' and ':' characters MUST remain unencoded and displayed
-  as-is when used as separators. They may need to be encoded elsewhere.
+- the '#', '?', '@' and ':' characters must NOT be encoded when used as
+  separators. They may need to be encoded elsewhere
 
-- the colon ':' separator between ``scheme`` and ``type`` MUST remain unencoded.
-  For example, in the PURL snippet ``pkg:npm`` the colon ':' MUST remain
-  unencoded and displayed as-is, i.e., ``pkg:npm``, and the PURL snippet
-  ``pkg%3Anpm`` is invalid.
+- the ':' ``scheme`` and ``type`` separator does not need to and must NOT be encoded.
+  It is unambiguous unencoded everywhere
 
 - the '/' used as ``type``/``namespace``/``name`` and ``subpath`` segments separator
   does not need to and must NOT be percent-encoded. It is unambiguous unencoded
@@ -246,7 +243,7 @@ Use these rules for percent-encoding and decoding ``purl`` components:
 - the '=' ``qualifiers`` key/value separator must NOT be encoded
 - the '#' ``subpath`` separator must be encoded as ``%23`` elsewhere
 
-- All non-ASCII characters MUST be encoded as UTF-8 and then percent-encoded.
+- All non-ASCII characters must be encoded as UTF-8 and then percent-encoded
 
 It is OK to percent-encode ``purl`` components otherwise except for the ``type``.
 Parsers and builders must always percent-decode and percent-encode ``purl``
--- a/PURL-SPECIFICATION.rst
+++ b/PURL-SPECIFICATION.rst
@@ -55,7 +55,7 @@ sometimes look like a ``host`` but its interpretation is specific to a ``type``.
 
 
 Some ``purl`` examples
-~~~~~~~~~~~~~~~~~~~~
+~~~~~~~~~~~~~~~~~~~~~~
 
 ::
 
@@ -72,7 +72,7 @@ Some ``purl`` examples
 
 
 A ``purl`` is a URL
-~~~~~~~~~~~~~~~~~
+~~~~~~~~~~~~~~~~~~~
 
 - A ``purl`` is a valid URL and URI that conforms to the URL definitions or
   specifications at:
@@ -110,7 +110,7 @@ A ``purl`` is a URL
 
 
 Rules for each ``purl`` component
-~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
 A ``purl`` string is an ASCII URL string composed of seven components.
 
@@ -122,27 +122,11 @@ The rules for each component are:
 
 - **scheme**:
 
-  - The ``scheme`` is a constant with the value "pkg"
-  - Since a ``purl`` never contains a URL Authority, its ``scheme`` must not be
-    suffixed with double slash as in 'pkg://' and should use instead
-    'pkg:'. Otherwise this would be an invalid URI per rfc3986 at
-    https://tools.ietf.org/html/rfc3986#section-3.3::
-
-        If a URI does not contain an authority component, then the path
-        cannot begin with two slash characters ("//").
-
-    It is therefore incorrect to use such '://' scheme suffix as the URL would
-    no longer be valid otherwise. In its canonical form, a ``purl`` must
-    NOT use such '://' ``scheme`` suffix but only ':' as a ``scheme`` suffix.
-  - ``purl`` parsers must accept URLs such as 'pkg://' and must ignore the '//'.
-  - ``purl`` builders must not create invalid URLs with such double slash '//'.
-  - The ``scheme`` is followed by a ':' separator
-  - For example these two purls are strictly equivalent and the first is in
-    canonical form. The second ``purl`` with a '//' is an acceptable ``purl`` but is
-    an invalid URI/URL per rfc3986::
-
-            pkg:gem/ruby-advisory-db-check@0.12.4
-            pkg://gem/ruby-advisory-db-check@0.12.4
+  - The ``scheme`` is a constant with the value "pkg".
+  - The scheme and type MUST be separated by a colon ':'.
+  - ``purl`` parsers MUST accept URLs in which the scheme and colon ':' are
+    followed by one or more slash '/' characters, such as 'pkg://', and MUST
+    ignore -- i.e., normalize by removing -- all such '/' characters.
 
 
 - **type**:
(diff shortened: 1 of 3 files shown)


gernot-h:
Great, thx for addressing this! I think this makes very clear that special characters must not be encoded as separators.

PURL-SPECIFICATION.rst
```diff
@@ -247,8 +247,9 @@ Use these rules for percent-encoding and decoding ``purl`` components:
 - the '#', '?', '@' and ':' characters must NOT be encoded when used as
   separators. They may need to be encoded elsewhere
```
gernot-h, line 248: To fully address the confusion of #39, I would probably also change the last sentence here:

```suggestion
  separators. Some of them need to be encoded elsewhere as specified in the rules below.
```

PURL-SPECIFICATION.rst
```diff
@@ -247,8 +247,9 @@ Use these rules for percent-encoding and decoding ``purl`` components:
...
 
-- the ':' ``scheme`` and ``type`` separator does not need to and must NOT be encoded.
-  It is unambiguous unencoded everywhere
+- The colon ':' separator between ``scheme`` and ``type`` MUST NOT be encoded.
+  For example, in the PURL snippet ``pkg:npm`` the colon ':' MUST NOT be encoded,
+  and the PURL snippet ``pkg%3Anpm`` is invalid.
```
johnmhoran, line 253: @gernot-h   @pombredanne   Consider adding at the top of the file, perhaps as a new one-line paragraph following the current first paragraph, something along the lines of the following:

> This specification uses RFC 2119 (https://datatracker.ietf.org/doc/html/rfc2119), as clarified in RFC 8174 (https://datatracker.ietf.org/doc/html/rfc8174), for the interpretation of certain terms, e.g., ``MUST NOT``.

Or perhaps a slight modification to the example provided by RFC 2119:

> The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED",  "MAY", and "OPTIONAL" in this document are to be interpreted as described in [RFC 2119](https://datatracker.ietf.org/doc/html/rfc2119), as clarified in [RFC 8174](https://datatracker.ietf.org/doc/html/rfc8174).

(Note that the core spec currently contains a great deal of language that will need to be modified to implement RFC 2119/8174.)

PURL-SPECIFICATION.rst
```diff
@@ -247,8 +247,9 @@ Use these rules for percent-encoding and decoding ``purl`` components:
 - the '#', '?', '@' and ':' characters must NOT be encoded when used as
   separators. They may need to be encoded elsewhere
 
-- the ':' ``scheme`` and ``type`` separator does not need to and must NOT be encoded.
-  It is unambiguous unencoded everywhere
+- The colon ':' separator between ``scheme`` and ``type`` MUST NOT be encoded.
```
pombredanne, line 250: I guess that following @JimFuller-RedHat suggestion in today's triage call https://github.com/package-url/purl-spec/discussions/360 ... this could reworded as:
```suggestion
- The colon ':' separator between ``scheme`` and ``type`` MUST be used as-is, unencoded.
```

@JimFuller-RedHat ?

JimFuller-RedHat: Hehe, I was (badly) trying say that specs typically define grammar, schema and behavior (ex. processing rules) ... in this case I think how you had it was fine. It is easier to define a 'thing' precisely which has the side effect of defining all the things it cannot be ;)

From grammar pov we could define (in EBNF) as a literal colon ... any parser built with such an EBNF would bork on encoded input ... we might suggest defining pre processing rules to normalise odd input into canonical purl. or define a looser grammar allowing encoding. 

I would propose:

_"The``scheme`` and ``type`` MUST be separated by a colon ':' "_

johnmhoran:
I added a test for the `:` between scheme and type.  More to follow tomorrow.

PURL-SPECIFICATION.rst
pombredanne: @JimFuller-RedHat 
> I would propose:
> 
> _"The`scheme` and `type` MUST be separated by a colon ':' "_

clean and neat.

Yet someone would likely immediately bet a nickel that someone else will open an issue asking whether the colon can or must or must not be encoded. ;)

Here is the way out:
We can keep these negative assertions as comments in a new FAQ document. I think that we can create a leaner spec this way.

I think it makes sense for the spec to avoid specifying all the things that it does not support as you rightly suggest, because there is an infinity of these. And we can push "negative" support questions in that FAQ.

So here:

- In the spec:
  - `The scheme and type MUST be separated by a colon ':'`

- In the FAQ:
  - Q: Is the colon between scheme and type encoded? Or can it be encoded? If yes, how?
  - A: The spec has no mention of encoding there, so the colon should be used as-is, never encoded and never requiring any decoding since this is not encoded. It is a parsing error if the colon does not come directly after `pkg`; Lenient parsing tools are welcome to recover from this error to help process and sanitize damaged purls, but that's not a requirement, and not part of the spec.

That approach can become the standard. The spec will be leaner and crisper and a the FAQ comprehensive to address the questions that _will_ come up for sure.

JimFuller-RedHat: cool ! fantastic suggestion - brevity and conciseness enables implementators to focus ... a good example is the json spec - https://ecma-international.org/wp-content/uploads/ECMA-404_2nd_edition_december_2017.pdf , which only contains normative text ;)

johnmhoran:
I've just pushed a commit that (1) updates the scheme-related sections of the core spec (PURL-SPECIFICATION.rst) inter alia to remove the use of "not" and to address questions that have been raised in the open issues/PRs and (2) provides added details and clarifications (including the use of "not" where helpful) in a new faq.rst.

I look forward to comments and questions, with the goal of finalizing the changes (including added modifications, tests etc. as needed) as part of the 1.0 release.  This PR is intended to address only the scheme-related updates, and other updates will be addressed in a similar narrowly-focused manner.

johnmhoran:
@matt-phylum @JimFuller-RedHat @pombredanne I've pushed my updates to this `scheme`-focused PR -- ready for your (and others') reviews and comments.  `scheme`-focused changes made to `PURL-SPECIFICATION.rst` and `test-suite-data.json` and a new `faq.rst` created.

PURL-SPECIFICATION.rst
```diff
@@ -122,27 +122,11 @@ The rules for each component are:
...
-  - ``purl`` builders must not create invalid URLs with such double slash '//'.
-  - The ``scheme`` is followed by a ':' separator
-  - For example these two purls are strictly equivalent and the first is in
-    canonical form. The second ``purl`` with a '//' is an acceptable ``purl`` but is
-    an invalid URI/URL per rfc3986::
-
-            pkg:gem/ruby-advisory-db-check@0.12.4
-            pkg://gem/ruby-advisory-db-check@0.12.4
+  - The ``scheme`` is a constant with the value "pkg".
+  - The ``scheme`` and ``type`` MUST be separated by a colon ':'.
```
pombredanne, line 126:
```suggestion
  - The ``scheme`` MUST be followed by a unencoded colon ':'.
```

PURL-SPECIFICATION.rst
```diff
@@ -122,27 +122,11 @@ The rules for each component are:
...
+  - ``purl`` parsers MUST accept URLs in which the ``scheme`` and colon ':' are
```
pombredanne, line 127:
```suggestion
  - ``purl`` parsers MUST accept URLs where the ``scheme`` and colon ':' are
```

PURL-SPECIFICATION.rst
```diff
@@ -122,27 +122,11 @@ The rules for each component are:
...
+    followed by one or more slash '/' characters, such as 'pkg://', and MUST
+    ignore -- i.e., normalize by removing -- all such '/' characters.
```
pombredanne, line 129:
```suggestion
    ignore and remove all such '/' characters.
```

PURL-SPECIFICATION.rst
```diff
@@ -234,21 +218,24 @@ The rules for each component are:
 Character encoding
 ~~~~~~~~~~~~~~~~~~
 
-For clarity and simplicity a ``purl`` is always an ASCII string. To ensure that
-there is no ambiguity when parsing a ``purl``, separator characters and non-ASCII
-characters must be UTF-encoded and then percent-encoded as defined at::
+For clarity and simplicity, a ``purl`` is always an ASCII string. To ensure that
```
pombredanne, line 221: We need to refactor this section separately. @mprpic volunteered to make this PR.
Let's remove all changes to this Character encoding section from this PR.

pombredanne (commented):
Here is some feedback

PURL-SPECIFICATION.rst
johnmhoran: Good suggestion -- change made.

PURL-SPECIFICATION.rst
johnmhoran: Good clarification, change made, and I'm making a corresponding change to related language in the FAQ.

PURL-SPECIFICATION.rst
johnmhoran: @pombredanne I've backed out all of my prior changes to the "Character encoding" section as requested.  In light of @mprpic 's PR #389, I've also deleted a reference to the "Character encoding" section in the FAQ and made a few other adjustments.

johnmhoran:
@pombredanne @mprpic I've made the requested changes and some other needed updates, including deleting a reference in the FAQ to the "Character encoding" section in light of @mprpic 's PR #389 (not yet merged as of right now), which deletes the "Character encoding" section.

One thought -- we might want to defer review of this PR until PR #389 has been merged and I can merge the updated `master` into this branch and push again.

@pombredanne I tried to request a review by you, but given the unusual parentage of this PR, which you opened, GH won't allow me to request your approving review.  🙂

johnmhoran:
^ looks like spam or a bot?

faq.rst
```diff
@@ -0,0 +1,46 @@
...
+
+- a "Components" section that includes each of the seven PURL components
+  (``scheme``, ``type``, ``namespace``, ``name``, ``version``, ``qualifiers``
+  and ``subpath``), and
+
+- a "General" section containing a mix of questions and answers that don't fit
+  neatly into a component-focused category.
+
+If you have a question about the PURL specification and don't find an answer
+below, you can open an issue `here <https://github.com/package-url/purl-spec/issues/new?template=Blank+issue>`_.
```
mprpic, line 14: I would probably ditch this paragraph since it's stating something that is hopefully obvious to anyone interacting with this repository.

faq.rst
```diff
@@ -0,0 +1,46 @@
...
+
+Components
+~~~~~~~~~~
+
+Scheme
+------
+
+**QUESTION**: Can the ``scheme`` component be followed by a colon and two slashes, like a URI?
+
+No.  Since a Package-URL, or PURL, never contains a URL Authority, its ``scheme`` should not be suffixed with double slash as in 'pkg://' and should use 'pkg:' instead. Otherwise this would be an invalid URI per RFC 3986 at https://tools.ietf.org/html/rfc3986#section-3.3::
```
mprpic, line 24:
```suggestion
No. Since a ``purl`` never contains a URL Authority, its ``scheme`` should not be suffixed with double slash as in 'pkg://' and should use 'pkg:' instead. Otherwise this would be an invalid URI per RFC 3986 at https://tools.ietf.org/html/rfc3986#section-3.3::
```

faq.rst
```diff
@@ -0,0 +1,46 @@
...
+
+    If a URI does not contain an authority component, then the path
+    cannot begin with two slash characters ("//").
+
+This rule applies to all slash '/' characters between the ``scheme``'s colon separator and the ``type`` component, e.g., ':/', '://', ':///' et al.
+
+In its canonical form, a PURL must not use any such ':/' ``scheme`` suffix and may only use ':' as a ``scheme`` suffix.  This means that:
```
mprpic, line 31:
```suggestion
In its canonical form, a purl must not use any such ':/' ``scheme`` suffix and may only use ':' as a ``scheme`` suffix.  This means that:
```

faq.rst
```diff
@@ -0,0 +1,46 @@
...
+
+- PURL parsers must accept URLs such as 'pkg://' and must ignore and remove all such '/' characters.
```
mprpic, line 33:
```suggestion
- purl parsers must accept URLs such as 'pkg://' and must ignore and remove all such '/' characters.
```

faq.rst
```diff
@@ -0,0 +1,46 @@
...
+- PURL builders should not create invalid URLs with one or more slash '/' characters between 'pkg:' and the ``type`` component.
```
mprpic, line 34:
```suggestion
- purl builders should not create invalid URLs with one or more slash '/' characters between 'pkg:' and the ``type`` component.
```

faq.rst
```diff
@@ -0,0 +1,46 @@
...
+
+For example, although these two PURLs are strictly equivalent, the first is in canonical form, while the second -- with a '//' between 'pkg:' and the ``type`` 'gem' -- is an acceptable PURL but is an invalid URI/URL per RFC 3986::
```
mprpic, line 36:
```suggestion
For example, although these two purls are strictly equivalent, the first is in canonical form, while the second -- with a '//' between 'pkg:' and the ``type`` 'gem' -- is an acceptable purl but is an invalid URI/URL per RFC 3986::
```

faq.rst
```diff
@@ -0,0 +1,46 @@
...
+
+    pkg:gem/ruby-advisory-db-check@0.12.4
+
+    pkg://gem/ruby-advisory-db-check@0.12.4
+
+**QUESTION**: Is the colon between ``scheme`` and ``type`` encoded? Can it be encoded? If yes, how?
+
+The "Rules for each ``purl`` component" section provides that "[t]he ``scheme`` MUST be followed by an unencoded colon ':'.
+
+In this case, the colon ':' between ``scheme`` and ``type`` is being used as a separator, and consequently should be used as-is, never encoded and never requiring any decoding. Moreover, it should be a parsing error if the colon ':' does not come directly after 'pkg'.  Tools are welcome to recover from this error to help with damaged PURLs, but that's not a requirement.
```
mprpic, line 46:
```suggestion
In this case, the colon ':' between ``scheme`` and ``type`` is being used as a separator, and consequently should be used as-is, never encoded and never requiring any decoding. Moreover, it should be a parsing error if the colon ':' does not come directly after 'pkg'.  Tools are welcome to recover from this error to help with malformed purls, but that's not a requirement.
```

mprpic (dismissed):
LGTM, just some minor nitpicks around consistency.

johnmhoran:
@mprpic Thanks for your comments.  While I'm happy to incorporate them, before I do I want to check with you and @pombredanne to make sure that despite Ecma TC54-TG2's 2024-10-11 approval of `Package-URL` and `PURL` as the proper capitalization, we want nonetheless to use lowercase `purl` in the spec.  See https://github.com/Ecma-TC54/tg2/blob/main/meetings/2024-10-11.md.

```
- Adam, Steve, Philippe discussed capitalization (or not) of the word “purl”.
  - Adam: for non-technical writing, we propose Package-URL and PURL.
  - Steve: in favor of this. This group can lead the standardization of the name.
  - Decided: Package-URL and PURL
```

faq.rst
johnmhoran: @mprpic   I wholeheartedly agree.  🙂  Will do as soon as we clarify whether to retain (temporarily or otherwise) the current lowercase `purl` usage in the core spec.  (See my [comment further below](https://github.com/package-url/purl-spec/pull/361#issuecomment-2640836146) to you and @pombredanne re Ecma TC54-TG2's approval of 'PURL'/'Package-URL'.)

johnmhoran: Removed para re opening an issue.

faq.rst
johnmhoran: Changed to 'purl'.

faq.rst
johnmhoran: Changed to 'purl'.

faq.rst
johnmhoran: Changed to 'purl'.

faq.rst
johnmhoran: Changed to 'purl'.

faq.rst
johnmhoran: Changed to 'purl'.

faq.rst
johnmhoran: Good suggestion -- done.

johnmhoran:
Thanks for sharing that link @mprpic .  I've adopted your suggestions and made a few other changes, will make corresponding changes to my 'type' PR as well.

mprpic (approved):

jkowalleck (approved):

PURL-SPECIFICATION.rst
```diff
@@ -236,7 +220,7 @@ Character encoding
 
 For clarity and simplicity a ``purl`` is always an ASCII string. To ensure that
 there is no ambiguity when parsing a ``purl``, separator characters and non-ASCII
-characters must be UTF-encoded and then percent-encoded as defined at::
+characters MUST be UTF-encoded and then percent-encoded as defined at::
```
pombredanne, line 223:
```suggestion
characters must be UTF-encoded and then percent-encoded as defined at::
```

pombredanne (commented):
Looking ready to merge IMHO.

pombredanne:
I only reverted minor change to the encoding section, so I consider this still approved and will merge.

PR #364, opened 2024-12-10
test: character encoding

jeremylong:
Correct versions that included a purl with an un-encoded colon in the version (test-suite-data.json:88). Per [purl-spec#character-encoding](https://github.com/package-url/purl-spec#character-encoding)
    
    > the '#', '?', '@' and ':' characters must NOT be encoded when used as separators. They may need to be encoded elsewhere

See #39

### DIFF
--- a/test-suite-data.json
+++ b/test-suite-data.json
@@ -85,8 +85,8 @@
   },
   {
     "description": "docker uses qualifiers and hash image id as versions",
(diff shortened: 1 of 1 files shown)


mprpic (approved):

pombredanne (approved):
Thanks! Looking good.

PR #365, opened 2024-12-10
test: correct test description for leading slashes

jeremylong:
Corrected copy-paste issue in the description of the test for /// (test-suite-data.json:255).

### DIFF
--- a/test-suite-data.json
+++ b/test-suite-data.json
@@ -252,7 +252,7 @@
     "is_invalid": false
   },
   {
-    "description": "slash /// after type  is not significant",
+    "description": "leading slash /// in front of type  is not significant",
     "purl": "pkg:///maven/org.apache.commons/io",
     "canonical_purl": "pkg:maven/org.apache.commons/io",
     "type": "maven",


test-suite-data.json
```diff
@@ -252,7 +252,7 @@
     "is_invalid": false
   },
   {
-    "description": "slash /// after type  is not significant",
+    "description": "leading slash /// in front of type  is not significant",
```
mprpic, line 255: I think it would be slightly more accurate to say that "trailing slash after `scheme` should be removed" since it's more associated with that field than the schema.

johnmhoran: or perhaps "all trailing slashes after `scheme` should be removed" ?  BTW, do these descriptions need to refer to both the scheme and the colon ':' separator, since the slashes follow not just the scheme but the scheme and its colon separator?  Or is that unneeded detail?

jeremylong: I used `leading` (or I suppose we could use preceding) because of the colon separator. If we said "trailing slash after schema" I think that would apply to `pkg////:`

johnmhoran: That's how I'd interpret it as well.  :+1:

mprpic: Ha, in `pkg////:` I understand `pkg` is the scheme field, `////` are trailing slashes, and `:` is the separator between scheme and type :-) The separators, in my view, don't really belong to any of the components.

johnmhoran:
`"description": "leading slash /// in front of type  is not significant",` was replaced by ` "description": "slash /// after scheme is not significant",` in [PR 383](https://github.com/package-url/purl-spec/pull/383), merged on 2025-02-19.  Closing this PR.

PR #372, opened 2024-12-19
TYPES: Add `yocto` type

jloehel:
- Add a first draft for the description of the `yocto` type


PURL-TYPES.rst
```diff
@@ -586,6 +586,50 @@ swift
...
+
+Examples
+^^^^^^^^
+
+::
+
+  pkg:yocto/poky/python-3dbus@f1ad013?PV=1.2.18#meta/recipes-devtools/python
+  pkg:yocto/openembedded-core/glibc@9400e1e9208b0f9075dfdfce0a3d1318a7fe6bf4?PV=2.35&repository=git.openembedded.org#meta/recipes-core/
+  pkg:yocto/meta-xilinx-core/u-boot-zynq-uenv@06e35a4#recipes-bsp/u-boot/
+  pkg:yocto/meta-odroid/emmc@4e07fab?PV=1.0.0&repository=github.com/akuster#recipes-bsp/
```
mprpic, line 630: While qualifier keys are case insensitive, the canonical form of a purl lowercases all keys so I think this should just contain `pv=1.0.0`.

PURL-TYPES.rst
```diff
@@ -586,6 +586,50 @@ swift
...
+The subpath is mandatory if the recipe is not in the root direcotry of the
+repository.
+
+Examples
+^^^^^^^^
+
+::
+
+  pkg:yocto/poky/python-3dbus@f1ad013?PV=1.2.18#meta/recipes-devtools/python
+  pkg:yocto/openembedded-core/glibc@9400e1e9208b0f9075dfdfce0a3d1318a7fe6bf4?PV=2.35&repository=git.openembedded.org#meta/recipes-core/
```
mprpic, line 628: The repository qualifier should be percent-encoded, so `repository=git.openembedded.org%23meta%2Frecipes-core%2F`.

PURL-TYPES.rst
```diff
@@ -586,6 +586,50 @@ swift
...
+
+The default repository is: ``git.yoctoproject.org``
+
+:namespace: The namespace is the name of the layer which provides the recipe.
+:name: The name is the name of the recipe.
+:version: The version is a commit or tag
+
+Qualifiers
+^^^^^^^^^^
+``repository``
```
mprpic, line 602: Other purl types use the qualifier `repository_url`, is there a specific reason to not use the same name here and have `repository` only?

PURL-TYPES.rst
jloehel: Makes sense. I will use lowercase here.

PURL-TYPES.rst
jloehel: I was not sure because of the missing protocol. But maybe it's better to rename it and the protocol is mandatory.

mprpic: +1 for making it mandatory (or at least suggesting its use), explicit is better than implicit :-)

jloehel:
Thanks for the long review @petermarko :-) I will try to answer some questions. 
 
> This PR seems to treat yocto layer as a distribution. But in fact, Yocto project is a tool to create your own distribution consisting of many layers and configuration options. As example, openembedded-core provides a recipe which would be linked in this purl. However, the recipe can be highly configurable (e.g. via packageconfig options). Recipe cannot be built without a distro which can modify it (e.g. poky) via bbappends and/or distro confiurations. As an example from poky reference distribution, I'd choose distro config influencing CVE_STATUS: [yoctoproject/poky@`yocto-5.1.4`/meta/recipes-graphics/xorg-xserver/xserver-xorg.inc#L179](https://github.com/yoctoproject/poky/blob/yocto-5.1.4/meta/recipes-graphics/xorg-xserver/xserver-xorg.inc?rgh-link-date=2025-04-09T14%3A31%3A07Z#L179) My idea would be to add additional qualifiers (e.g. `&distro=poky-altcfg&distro_url=https%3A%2F%2Fweb.git.yoctoproject.org%2Fpoky&distro_revision=...`). I'd also make these mandatory as the various layers where the recipes are stored need a distro to be built.

I was not aware of that. I thought https://web.git.yoctoproject.org/poky/ is also a layer. But yeah, we can do that. I guess `poky` would be the default, right? How would a distro_url look like. Is it always a git repo?

> 
> Another question is why the repository revision was chosen as "version". I think that with many configurations possibilities "version" should be the component version (currently stored in qualifier pv). Especially since the final configuration is given by distro and not layer containing the recipe. But I'm completely new to purl and thus I maybe be little off, so I'd like to hear some thoughts about these decisions.

Because the PURL locate in my view a recipe which specifies how to build the package and more important some meta information like license(s), CVEs, .... It's easier to resolve the PURL with the commit hash as version. I can parse, interpret and extract (with bitbake or an own interpreter for recipes) the PV from the recipe if necessary. That means the PV is not mandatory. Let me try to understand. I am not an expert in Building Yocto Images. How does the distro change the yocto recipe?

> 
> Also, hash is not so easy to decode on the first look. Could we add optional distro codename as e.g. Debian purl does? e.g. `&release=scarthgap`, as "distro" is already a "reserved word" in Yocto project.

If I remember correctly release names in Yocto are branches in the layer repos, right? I don't see a problem to use tags or branches as aliases for commit hashes. But I think there is no extra classifier necessary. For me the PURL does not need to be human readable. ^^ The application should be able to resolve the PURL, download the recipe, download the source code, analyze the source code/recipe and enrich a SBOM. Maybe that's the missing context you have mentioned. 

> 
> And finally, why is subpath mandatory. I think that it is rare to have two recipe versions in the same layer. And when it happens, they are usually in the same path. The link to repository should be enough as it’s easy to find recipes in a layer. My proposal would be to make it optional or remove completely (as distro will define what is used).
> 

It's not really mandatory but if the recipe(s) are not in the root folder or better if the repo includes more than one layer. In example https://git.openembedded.org/meta-openembedded/tree/ includes multiple layers. It should be mandatory to add the subpath to the right layer. It is not guaranteed that the repo contains always exact one layer.
For me it was easier to resolve the PURL with the subpath. Let's  rephrase this sentence. Do you have any suggestions Peter?

> Thanks a lot for your proposal and looking forward to get it finalized and merged, Peter

Thanks for your contribution. :-) It's always good to have feedback.

petermarko:
> I was not aware of that. I thought https://web.git.yoctoproject.org/poky/ is also a layer. But yeah, we can do that. I guess poky would be the default, right? How would a distro_url look like. Is it always a git repo?

Well, poky is a special case.
It is a virtual/combo repository which contains copies or 4 repositories (bitbake engine, oe-core layer, documentation and a reference distro - meta-yocto repository with meta-poky layer).
So I'd never link poky as package/distro source, it would create parallel world of versions (use openembedded-core for recipe source or meta-poky as distro source).

Defining yocto distribution in a short string is very difficult.
I think that yocto world is around git, so I'd say is a configuration which is available in a git repo.
Example can be (from reference distro): https://git.yoctoproject.org/meta-yocto/tree/meta-poky/conf/distro
Here we see that there are 4 reference distros which more or less define the possibilities of this layer.
It's of course not as simple as linking to the poky.conf, as each distro can have different build variants.
But I'd say it's impossible to put everything into in a simple string, so we need to abstract it.
My proposal would be to define distro as name of the conf file, e.g. `poky-altcfg` for `https://git.yoctoproject.org/meta-yocto/tree/meta-poky/conf/distro/poky-altcfg.conf`.

> Because the PURL locate in my view a recipe which specifies how to build the package and more important some meta information like license(s), CVEs, .... It's easier to resolve the PURL with the commit hash as version. I can parse, interpret and extract (with bitbake or an own interpreter for recipes) the PV from the recipe if necessary. That means the PV is not mandatory. Furthermore the PV does not need to change through releases but maybe the patches do. I would need to search in all git references for this PV and I will have multiple recipes to choose from. I think it's safer with a commit hash. Let me try to understand. I am not an expert in Building Yocto Images. How does the distro change the yocto recipe?

As I already mentioned, distro can significantly change the recipe.
I'd say that in the same way as the recipe itself modifies how the sources are built, the distro changes how the recipe is built.
In some cases the distro does not change anything, in some cases is changes almost everything.
So with this `sources -> recipe -> distro` changes, it really is a question which of these three should be chosen as the main version.
Is it component version (PV), or recipe layer or is it distro config?
My personal preference would be PV, but I guess any of the three items would have someone who would vote for it.

> If I remember correctly release names in Yocto are branches in the layer repos, right? I don't see a problem to use tags or branches as aliases for commit hashes. But I think there is no extra classifier necessary. For me the PURL does not need to be human readable. ^^ The application should be able to resolve the PURL, download the recipe, download the source code, analyze the source code/recipe and enrich a SBOM. Maybe that's the missing context you have mentioned.

Cool, having tag (or git describe string which includes tag and possible hash) looks like even better way than release codename.

> It's not really mandatory but if the recipe(s) are not in the root folder or better if the repo includes more than one layer. In example https://git.openembedded.org/meta-openembedded/tree/ includes multiple layers. It should be mandatory to add the subpath to the right layer. It is not guaranteed that the repo contains always exact one layer. For me it was easier to resolve the PURL with the subpath. Let's rephrase this sentence. Do you have any suggestions Peter?

OK, I didn't think about it like a path to layer, rather as path to recipe.
This is fine for me.

---

I have also sent link to this PR to yocto distribution list.
Maybe someone move suitable than me will also comment here.
Involving Yocto project is I think the right thing when we're defining what the project actually means in pURL.

https://lists.openembedded.org/g/openembedded-architecture/message/2141

JPEWdev:
Hello, I do most of the SBoM work for Yocto, so I'd like a chance to review this and make sure it aligns with our goals. Thanks

jloehel:
>  My personal preference would be PV, but I guess any of the three items would have someone who would vote for it.

Okay, can you please show me an example with the PV as version. I would like to resolve it and test it.

If we would have the example
#### Layer
* **repo**: "https://git.openembedded.org/meta-openembedded"
* **path**: meta-oe

#### Package
* **name**: nano
* **version**: 6.2

I will find 3 recipes with version 6.2:

~~~Python
[
	YoctoRecipe(
		name=nano_6.2.bb,
		src_uri=https://nano-editor.org/dist/v6/nano-6.2.tar.xz,
		layer_ref=origin/kirkstone,
		checksum=51e1942c,
	),
	YoctoRecipe(
		name=nano_6.2.bb,
		src_uri=https://nano-editor.org/dist/v6/nano-6.2.tar.xz,
		layer_ref=origin/kirkstone-next,
		checksum=51e1942c
	),
	YoctoRecipe(
		name=nano_6.2.bb,
		src_uri=https://nano-editor.org/dist/v6/nano-6.2.tar.xz,
		layer_ref=origin/stable/kirkstone-nut,
		checksum=ef77c52f
	)
]
~~~

2 recipes are the same but the third recipe differs. It would be necessary to specify additionally the `release` reference like `kirkstone`  or the target commit of kirkstone. Like you mentioned earlier. I think if I want to address a specific recipe, it's better to use the ref of the recipe.

I am still thinking about the distro. I guess I would consider also `formulation` in a cyclonedx SBOM but that would mean to use rpm, deb or ipk for the components. But the PURL should be independent of the SBOM. What classifiers do we need for the distro? A direct link to the distro_config should be enough right?

@JPEWdev I would love to here your opinion.

PURL-TYPES.rst
```diff
@@ -586,6 +586,52 @@ swift
...
+^^^^^^^
+The subpath is mandatory if the recipe is not in the root direcotry of the
+repository.
+
+Examples
+^^^^^^^^
+
+::
+
+  pkg:yocto/poky/python-3dbus@f1ad013?pv=1.2.18#meta/recipes-devtools/python
```
petermarko, line 629: Small typo (also in test-suite-data.json)
```suggestion
  pkg:yocto/poky/python3-dbus@f1ad013?pv=1.2.18#meta/recipes-devtools/python
```

JPEWdev:
Apologies, this is sort of all over the place, since I'm trying to get all my thoughts out :)

I'm pretty new to PURL, but I know a lot about yocto, so I'd need some context. I think perhaps it's important to keep in mind the Yocto build work flow. The way that it works is that a recipes (.bb file) is used to compile some source code (defined in the recipe) into binary packages. These packages can be in multiple formats (deb/rpm/ipk, currently), but even though they use the same format as other well known distros, that doesn't mean they are compatible (e.g. just because Yocto builds a .deb package doesn't mean it can be installed on Debian, and it doesn't mean Debian packages can be installed in a Yocto build; Yocto is just reusing the .deb file format). The produced binary packages are then installed using the correct packaging tool (e.g. `apt`, `dnf`, `opkg`) to produce the final image that would comprise the final product. However, some recipes (notably the Linux Kernel) don't work this way, and can directly produce binary artifacts from the recipe.

As such, there's a few levels that can be described; if you want the recipe (.bb) file used to build, that would be the first level. The second level would be binary packages that are produced.

I'm not exactly clear on what purpose of a recipe PURL would be in this context; is the intent to be able to match to PURL across other distros (e.g. Debian, Fedora, etc.) so that we can more or less tell if "my busybox is the same as yours?", and thus by extension to tell if a given CVE applies?

How do other Distros deal with multiple binary packages from the same "source"? For example, Ubuntu has `bash`, `bash-static`, `bash-doc` etc.? If all of these get the same PURL, than I think a recipe level PURL is probably the direction we need to go, if not.... I'm not really sure what we can do because Yocto would package the same build into completely different packages.

If we are then talking about recipe PURLs, then at the most basic, I think you'd probably want to do something like:
`pkg:yocto/python3-dbus@1.2.18` (or in Yocto parlance: `pkg:yocto/${PN}:${PV}`, although we could substitute `${PV}` in for something else if our PV name doesn't match the commonly accepted PURL name for the component)

IMHO, PURLs are only useful to Yocto if we can reasonably have the recipe actually write out the PURL in the SBoM (or elsewhere for that matter) at build time; it's unrealistic to assume that users will hand edit the PURL of everything. From that perspective, there's really only so much information we can realistically include in the PURL. I'm also unclear how much information in the PURL would be used to "match" (e.g. against a CVE). If the PURL is too precise and all the information is matched, it seems like that's a loosing proposition because no one is going to find every single Yocto PURL (especially ones they can't possibly know about) to list in their CVE. If you could shed some light on that for me it would be helpful

Distros can and do make changes to recipes, often changing configure time options, but I'm not really sure that matters too much; even if features are enabled/disabled, patches applied, etc. it's not like someone is transforming busybox in to python (at least, I hope not!), so this again goes to the purpose of the PURL; if it's correlation with other distros, the Yocto distro kinda doesn't matter (if you squint really hard anyway).

IMHO, `namespace` is pretty useless here as there no single work that can correctly describe it. We could leave it open to allow e.g. a company or organization to provide their domain here (e.g. `bobswidgets.com`) as an indication of the package supplier.

I suspect (again based on how matching is done), that our package version (`PV`) is better as the version than the hash. It's supposed to be the upstream version number and is certainly more useful (for matching), than the hash of the layer.

Also, the layer path is probably not useful in this context. PN and PV are sufficient to uniquely identify a recipe in a given layer. 

However, even if you have the layer URL, and SHA-1 you won't necessarily be able to find the exact recipe. This is because multiple layers can affect a single recipe, by providing .bbappend files that.... append.... the recipe and change it's behavior, or might have a machine config, or a distro conf, or many other thing that affect the recipe. The only way to come close to achieving this would be to provide the URL and SHA-1 of every single layer that contributed to the build, and even that's imperfect since the user can change the configuration of a recipe locally with files completely outside of source control.

Yocto _does_ have a way to distill all the complexity of every variable value, every line of a recipe, and every line of every bbappend that affects that recipe into a single entity; this is called the "taskhash" of the recipe, and it's integral to how Yocto track what need to rebuild when something change. Effectively, it's a hash of the entire state of the recipe, and Yocto will calculate it when a build is done; if it's change from the last know value the recipe has changed somehow and needs to be rebuilt. Even this though is imperfect for use in a PURL, because the taskhash is per recipe task, and each recipe has a dozen or more tasks, some of which affect the output of the recipe, and some of which don't. So it's not clear which taskhash would even make sense in a PURL, not to mention that the taskhash looses all information about how to _locate_ the recipe inputs; it would mostly just be useful from a matching perspective (if two PURLs have the same taskhash, they had the exact same input parameters to the recipe).

Sorry for the long ramble, I'm still trying to wrap my head around this. In short, I'm just not sure how much useful information we can actually include in a Yocto PURL. Any thoughts you have, or if you'd like me to elaborate more, please let me know.

Thanks.

jloehel:
@petermarko Hi :-)

> This example is probable not a good one, since kirkstone-next and kirkstone-nut are development branches.
So it's like you would pick branches branch of this PR...
But of course there are cases where the same recipe version (PV) is used on multiple release branches (including tags).
But is that any different from e.g. Debian when they have the same component version in Debian 11 and 12?

Maybe not, it was just the first one, I found. Maybe python3-dbus is a better example:

~~~
 YoctoRecipe(name=python3-dbus_1.2.18.bb, src_uri=http://dbus.freedesktop.org/releases/dbus-python/dbus-python-1.2.18.tar.gz, layer_ref=origin/honister checksum=27547f1e),
 YoctoRecipe(name=python3-dbus_1.2.18.bb, src_uri=http://dbus.freedesktop.org/releases/dbus-python/dbus-python-1.2.18.tar.gz, layer_ref=origin/kirkstone checksum=681121d2),
 YoctoRecipe(name=python3-dbus_1.2.18.bb, src_uri=http://dbus.freedesktop.org/releases/dbus-python/dbus-python-1.2.18.tar.gz, layer_ref=origin/langdale checksum=681121d2),
...
~~~

I mean if there is an additional information like a branch or commit, I can identify/locate it. Honestly I don't care so much if we write:

`@<commit|branch|tag>?PV=<PV>`  or `@<PV>?branch=<commit|branch|tag>`.

> As I said, it's a personal preference as I think that the base is oss component ...
Do you mean differ between the upstream tarball/repo and recipe? Can you please show me an SBOM example for this? Search for CVEs with `pkg:generic/python3-dbus@1.2.18` and apply the yocto layer info on top of it? Right now I am still using CPE for upstream and on top the infos from the recipe.

> One thing I still have is how to handle products which are not developed on public Github.
Imho most products contain proprietary code and recipes for those (but also configuration for public recipes) are in non-public proprietary layers.
When we're putting links to purl string, it won't have any meaning unless you work for that vendor and have access to the meta layer repository.

I made the experience that the most companies do not care. They care more about OSS licenses and vulns. However the non-public layers should be also in a git, private gitlab/github/cgit repo. The way how to construct the PURL should not differ. Granting access is nothing we can solve here.

@JPEWdev Thanks for the reply. 

> Apologies, this is sort of all over the place, since I'm trying to get all my thoughts out :)
> 
No need to apologies. :-)

> I'm pretty new to PURL, but I know a lot about yocto, so I'd need some context. I think perhaps it's important to keep in mind the Yocto build work flow. The way that it works is that a recipes (.bb file) is used to compile some source code (defined in the recipe) into binary packages. These packages can be in multiple formats (deb/rpm/ipk, currently), but even though they use the same format as other well known distros, that doesn't mean they are compatible (e.g. just because Yocto builds a .deb package doesn't mean it can be installed on Debian, and it doesn't mean Debian packages can be installed in a Yocto build; Yocto is just reusing the .deb file format). The produced binary packages are then installed using the correct packaging tool (e.g. `apt`, `dnf`, `opkg`) to produce the final image that would comprise the final product. However, some recipes (notably the Linux Kernel) don't work this way, and can directly produce binary artifacts from the recipe.

Thanks for the explanation. :-)

> 
> As such, there's a few levels that can be described; if you want the recipe (.bb) file used to build, that would be the first level. The second level would be binary packages that are produced.
> 
> I'm not exactly clear on what purpose of a recipe PURL would be in this context; is the intent to be able to match to PURL across other distros (e.g. Debian, Fedora, etc.) so that we can more or less tell if "my busybox is the same as yours?", and thus by extension to tell if a given CVE applies?

The purpose for me would be to download the recipe or better clone the corresponding layer. Checkout the correct version. Parse/Interpret the recipe with bitbake. Identify patches, CVEs, licenses and the source code (upstream tarball, configuration files, ...). Enrich a SBOM with the patches and the licenses. Enrich the VEX with the CVEs and add some basic analysis based on the patches. Download the source code and run further analysis on the source and enrich the SBOM/Vex. I do not care so much about bbappend files at this point. The objective is, to enrich an SBOM/VEX automatically as much as possible. If there are special bbappend files, these can be handled later manually by an analyst. ... or add them as file components to the SBOM.

> 
> How do other Distros deal with multiple binary packages from the same "source"? For example, Ubuntu has `bash`, `bash-static`, `bash-doc` etc.? If all of these get the same PURL, than I think a recipe level PURL is probably the direction we need to go, if not.... I'm not really sure what we can do because Yocto would package the same build into completely different packages.
>
I am not sure how to handle sub-packages. But probably not handled via PURL. 
 
> If we are then talking about recipe PURLs, then at the most basic, I think you'd probably want to do something like: `pkg:yocto/python3-dbus@1.2.18` (or in Yocto parlance: `pkg:yocto/${PN}:${PV}`, although we could substitute `${PV}` in for something else if our PV name doesn't match the commonly accepted PURL name for the component)

How would you resolve `pkg:yocto/python3-dbus@1.2.18` ? Which recipe is it:

* https://layers.openembedded.org/layerindex/recipe/325720/
* https://layers.openembedded.org/layerindex/recipe/300204/
* https://layers.openembedded.org/layerindex/recipe/190236/

To find this three recipes would involve to use openembedded.org or a own database of recipes, layers, distros, ... You need at least the branch/release `kirkstone` to identify the correct recipe considering that you can search already in a bunch of meta info.

> 
> IMHO, PURLs are only useful to Yocto if we can reasonably have the recipe actually write out the PURL in the SBoM (or elsewhere for that matter) at build time; it's unrealistic to assume that users will hand edit the PURL of everything. From that perspective, there's really only so much information we can realistically include in the PURL. I'm also unclear how much information in the PURL would be used to "match" (e.g. against a CVE). If the PURL is too precise and all the information is matched, it seems like that's a loosing proposition because no one is going to find every single Yocto PURL (especially ones they can't possibly know about) to list in their CVE. If you could shed some light on that for me it would be helpful

I agree. If there would be a way to include the PURL, Patches, CVEs, SRC already during the build to the SBOM would be awesome. This would make the enrichment afterwards not necessary anymore. I haven't thought so much about CVE matching. More about locating the recipe. But you are right matching would be a problem if it's too precise.

> 
> Distros can and do make changes to recipes, often changing configure time options, but I'm not really sure that matters too much; even if features are enabled/disabled, patches applied, etc. it's not like someone is transforming busybox in to python (at least, I hope not!), so this again goes to the purpose of the PURL; if it's correlation with other distros, the Yocto distro kinda doesn't matter (if you squint really hard anyway).

I guess, it's more about making the supply chain more visible. Where is the upstream tarball coming from. Who applies what patches. If the distro has really so much power that it would be theoretical possible to transform busybox into python, I would be also worried about that. Not what is likely, what is possible. But maybe the PURL is not the right location. I guess it's also something I would add to the SBOM. 

> 
> IMHO, `namespace` is pretty useless here as there no single work that can correctly describe it. We could leave it open to allow e.g. a company or organization to provide their domain here (e.g. `bobswidgets.com`) as an indication of the package supplier.
> 
> I suspect (again based on how matching is done), that our package version (`PV`) is better as the version than the hash. It's supposed to be the upstream version number and is certainly more useful (for matching), than the hash of the layer.
> 
> Also, the layer path is probably not useful in this context. PN and PV are sufficient to uniquely identify a recipe in a given layer.
> 
How would you add the layer info to the PURL? To find the recipe with PN and PV, I need to know where to search, right?

> However, even if you have the layer URL, and SHA-1 you won't necessarily be able to find the exact recipe. This is because multiple layers can affect a single recipe, by providing .bbappend files that.... append.... the recipe and change it's behavior, or might have a machine config, or a distro conf, or many other thing that affect the recipe. The only way to come close to achieving this would be to provide the URL and SHA-1 of every single layer that contributed to the build, and even that's imperfect since the user can change the configuration of a recipe locally with files completely outside of source control.

I know now that these kind of things are possible. It makes it really challenging. I guess it's only possible during build time and better to document in the SBOM instead of the PURL. But the SBOM should have all used recipes, bbappend files, local files, ... right? If I would combine all components from the SBOM, it should be possible to rebuild the "same" image, right?

> 
> Yocto _does_ have a way to distill all the complexity of every variable value, every line of a recipe, and every line of every bbappend that affects that recipe into a single entity; this is called the "taskhash" of the recipe, and it's integral to how Yocto track what need to rebuild when something change. Effectively, it's a hash of the entire state of the recipe, and Yocto will calculate it when a build is done; if it's change from the last know value the recipe has changed somehow and needs to be rebuilt. Even this though is imperfect for use in a PURL, because the taskhash is per recipe task, and each recipe has a dozen or more tasks, some of which affect the output of the recipe, and some of which don't. So it's not clear which taskhash would even make sense in a PURL, not to mention that the taskhash looses all information about how to _locate_ the recipe inputs; it would mostly just be useful from a matching perspective (if two PURLs have the same taskhash, they had the exact same input parameters to the recipe).

I think this is also more related to the build or assembling. I would document this in an SBOM not in the PURL.

> 
> Sorry for the long ramble, I'm still trying to wrap my head around this. In short, I'm just not sure how much useful information we can actually include in a Yocto PURL. Any thoughts you have, or if you'd like me to elaborate more, please let me know.
> 
> Thanks.

Thank you. These kind of deep knowledge about Yocto is really helpful.

JPEWdev:
> @petermarko Hi :-)
> 
> > This example is probable not a good one, since kirkstone-next and kirkstone-nut are development branches.
> > So it's like you would pick branches branch of this PR...
> > But of course there are cases where the same recipe version (PV) is used on multiple release branches (including tags).
> > But is that any different from e.g. Debian when they have the same component version in Debian 11 and 12?
> 
> Maybe not, it was just the first one I found. Maybe python3-dbus is a better example:
> 
> ```
>  YoctoRecipe(name=python3-dbus_1.2.18.bb, src_uri=http://dbus.freedesktop.org/releases/dbus-python/dbus-python-1.2.18.tar.gz, layer_ref=origin/honister checksum=27547f1e),
>  YoctoRecipe(name=python3-dbus_1.2.18.bb, src_uri=http://dbus.freedesktop.org/releases/dbus-python/dbus-python-1.2.18.tar.gz, layer_ref=origin/kirkstone checksum=681121d2),
>  YoctoRecipe(name=python3-dbus_1.2.18.bb, src_uri=http://dbus.freedesktop.org/releases/dbus-python/dbus-python-1.2.18.tar.gz, layer_ref=origin/langdale checksum=681121d2),
> ...
> ```
> 
> I mean if there is an additional information like a branch or commit, I can identify/locate it. Honestly I don't care so much if we write:
> 
> `@<commit|branch|tag>?PV=<PV>` or `@<PV>?branch=<commit|branch|tag>`.
> 
> > As I said, it's a personal preference as I think that the base is oss component ...
> > Do you mean differ between the upstream tarball/repo and recipe? Can you please show me an SBOM example for this? Search for CVEs with `pkg:generic/python3-dbus@1.2.18` and apply the yocto layer info on top of it? Right now I am still using CPE for upstream and on top the infos from the recipe.
> 
> > One thing I still have is how to handle products which are not developed on public Github.
> > Imho most products contain proprietary code and recipes for those (but also configuration for public recipes) are in non-public proprietary layers.
> > When we're putting links to purl string, it won't have any meaning unless you work for that vendor and have access to the meta layer repository.
> 
> I made the experience that the most companies do not care. They care more about OSS licenses and vulns. However the non-public layers should be also in a git, private gitlab/github/cgit repo. The way how to construct the PURL should not differ. Granting access is nothing we can solve here.
> 
> @JPEWdev Thanks for the reply.
> 
> > Apologies, this is sort of all over the place, since I'm trying to get all my thoughts out :)
> 
> No need to apologies. :-)
> 
> > I'm pretty new to PURL, but I know a lot about yocto, so I'd need some context. I think perhaps it's important to keep in mind the Yocto build work flow. The way that it works is that a recipes (.bb file) is used to compile some source code (defined in the recipe) into binary packages. These packages can be in multiple formats (deb/rpm/ipk, currently), but even though they use the same format as other well known distros, that doesn't mean they are compatible (e.g. just because Yocto builds a .deb package doesn't mean it can be installed on Debian, and it doesn't mean Debian packages can be installed in a Yocto build; Yocto is just reusing the .deb file format). The produced binary packages are then installed using the correct packaging tool (e.g. `apt`, `dnf`, `opkg`) to produce the final image that would comprise the final product. However, some recipes (notably the Linux Kernel) don't work this way, and can directly produce binary artifacts from the recipe.
> 
> Thanks for the explanation. :-)
> 
> > As such, there's a few levels that can be described; if you want the recipe (.bb) file used to build, that would be the first level. The second level would be binary packages that are produced.
> > I'm not exactly clear on what purpose of a recipe PURL would be in this context; is the intent to be able to match to PURL across other distros (e.g. Debian, Fedora, etc.) so that we can more or less tell if "my busybox is the same as yours?", and thus by extension to tell if a given CVE applies?
> 
> The purpose for me would be to download the recipe or better clone the corresponding layer. Checkout the correct version. Parse/Interpret the recipe with bitbake. Identify patches, CVEs, licenses and the source code (upstream tarball, configuration files, ...). Enrich a SBOM with the patches and the licenses. Enrich the VEX with the CVEs and add some basic analysis based on the patches. Download the source code and run further analysis on the source and enrich the SBOM/Vex. I do not care so much about bbappend files at this point. The objective is, to enrich an SBOM/VEX automatically as much as possible. If there are special bbappend files, these can be handled later manually by an analyst. ... or add them as file components to the SBOM.

Ok, so if I'm reading this correctly, the purpose of a PURL in this case (or maybe all cases?) is to identify a component in a succinct manner such that some one can come along later and perform some level of supply chain analysis on it using (basically) only the PURL (or more specifically, what the PURL points to)?

That didn't fit my conceptual model of what PURLs were supposed to be, but that's OK. I had thought they were an attempt to "correlate" things across different sources, e.g. fedora "busybox", debian "busybox" and yocto "busybox" might be configured differently, but they are all still busybox, so conclusions you make about (might) hold on all of them (or are at least worth looking at).

> 
> > How do other Distros deal with multiple binary packages from the same "source"? For example, Ubuntu has `bash`, `bash-static`, `bash-doc` etc.? If all of these get the same PURL, than I think a recipe level PURL is probably the direction we need to go, if not.... I'm not really sure what we can do because Yocto would package the same build into completely different packages.
> 
> I am not sure how to handle sub-packages. But probably not handled via PURL.

Ack. Recipes _probably_ make the most sense then. I think the weird thing here might be that recipes are a purely build system concept and there is indirection to the final thing (package) the end user gets, unlike some other PURLs which seem to focus on only the end user package?

I wonder how e.g. Gentoo would deal with PURLs, there are some similarities between how we do things.

> 
> > If we are then talking about recipe PURLs, then at the most basic, I think you'd probably want to do something like: `pkg:yocto/python3-dbus@1.2.18` (or in Yocto parlance: `pkg:yocto/${PN}:${PV}`, although we could substitute `${PV}` in for something else if our PV name doesn't match the commonly accepted PURL name for the component)
> 
> How would you resolve `pkg:yocto/python3-dbus@1.2.18` ? Which recipe is it:
> 
>     * https://layers.openembedded.org/layerindex/recipe/325720/
> 
>     * https://layers.openembedded.org/layerindex/recipe/300204/
> 
>     * https://layers.openembedded.org/layerindex/recipe/190236/
> 
> 
> To find this three recipes would involve to use openembedded.org or a own database of recipes, layers, distros, ... You need at least the branch/release `kirkstone` to identify the correct recipe considering that you can search already in a bunch of meta info.

Right. Then in that case, you probably want: The git repo URL (yocto repos aren't centralized on any particular git server, so there's no sane default here), the SHA-1 of the git repo, and the path to the .bb file in the repo.

> 
> > IMHO, PURLs are only useful to Yocto if we can reasonably have the recipe actually write out the PURL in the SBoM (or elsewhere for that matter) at build time; it's unrealistic to assume that users will hand edit the PURL of everything. From that perspective, there's really only so much information we can realistically include in the PURL. I'm also unclear how much information in the PURL would be used to "match" (e.g. against a CVE). If the PURL is too precise and all the information is matched, it seems like that's a loosing proposition because no one is going to find every single Yocto PURL (especially ones they can't possibly know about) to list in their CVE. If you could shed some light on that for me it would be helpful
> 
> I agree. If there would be a way to include the PURL, Patches, CVEs, SRC already during the build to the SBOM would be awesome. This would make the enrichment afterwards not necessary anymore. I haven't thought so much about CVE matching. More about locating the recipe. But you are right matching would be a problem if it's too precise.

We perhaps are in a unique situation that we already have really good SBoMs, and in fact they already (can) include all of this information (excepting the PURL of course). Since we are the build system, our data is very accurate also :)

I suspect most other cases this is the other way around where PURLs are being used to "reverse engineer" the SBoM from the end components because the build system isn't providing that information?

> 
> > Distros can and do make changes to recipes, often changing configure time options, but I'm not really sure that matters too much; even if features are enabled/disabled, patches applied, etc. it's not like someone is transforming busybox in to python (at least, I hope not!), so this again goes to the purpose of the PURL; if it's correlation with other distros, the Yocto distro kinda doesn't matter (if you squint really hard anyway).
> 
> I guess, it's more about making the supply chain more visible. Where is the upstream tarball coming from. Who applies what patches. If the distro has really so much power that it would be theoretical possible to transform busybox into python, I would be also worried about that. Not what is likely, what is possible. But maybe the PURL is not the right location. I guess it's also something I would add to the SBOM.

Ack. Its going to be quite hard to accurately capture in a single URL... people can and do patch CVEs in bbappend and what not, so without all the context it's difficult to draw accurate conclusions from just a PURL (or what the PURL points to). I'll give it some more thought though.

> 
> > IMHO, `namespace` is pretty useless here as there no single work that can correctly describe it. We could leave it open to allow e.g. a company or organization to provide their domain here (e.g. `bobswidgets.com`) as an indication of the package supplier.
> > I suspect (again based on how matching is done), that our package version (`PV`) is better as the version than the hash. It's supposed to be the upstream version number and is certainly more useful (for matching), than the hash of the layer.
> > Also, the layer path is probably not useful in this context. PN and PV are sufficient to uniquely identify a recipe in a given layer.
> 
> How would you add the layer info to the PURL? To find the recipe with PN and PV, I need to know where to search, right?

Correct. This comment was made under the (incorrect?) assumption PURLs were about correlating between distros (per above)

> 
> > However, even if you have the layer URL, and SHA-1 you won't necessarily be able to find the exact recipe. This is because multiple layers can affect a single recipe, by providing .bbappend files that.... append.... the recipe and change it's behavior, or might have a machine config, or a distro conf, or many other thing that affect the recipe. The only way to come close to achieving this would be to provide the URL and SHA-1 of every single layer that contributed to the build, and even that's imperfect since the user can change the configuration of a recipe locally with files completely outside of source control.
> 
> I know now that these kind of things are possible. It makes it really challenging. I guess it's only possible during build time and better to document in the SBOM instead of the PURL. But the SBOM should have all used recipes, bbappend files, local files, ... right? If I would combine all components from the SBOM, it should be possible to rebuild the "same" image, right?

The SBoM doesn't have layers et. al. today, but it's on the roadmap to be added (it's quite complicated, even in a well structured format like SPDX, if you can imagine), and yes if you have all the inputs, you should get the same output.

> 
> > Yocto _does_ have a way to distill all the complexity of every variable value, every line of a recipe, and every line of every bbappend that affects that recipe into a single entity; this is called the "taskhash" of the recipe, and it's integral to how Yocto track what need to rebuild when something change. Effectively, it's a hash of the entire state of the recipe, and Yocto will calculate it when a build is done; if it's change from the last know value the recipe has changed somehow and needs to be rebuilt. Even this though is imperfect for use in a PURL, because the taskhash is per recipe task, and each recipe has a dozen or more tasks, some of which affect the output of the recipe, and some of which don't. So it's not clear which taskhash would even make sense in a PURL, not to mention that the taskhash looses all information about how to _locate_ the recipe inputs; it would mostly just be useful from a matching perspective (if two PURLs have the same taskhash, they had the exact same input parameters to the recipe).
> 
> I think this is also more related to the build or assembling. I would document this in an SBOM not in the PURL.

And we do have that in the SBoM

> 
> > Sorry for the long ramble, I'm still trying to wrap my head around this. In short, I'm just not sure how much useful information we can actually include in a Yocto PURL. Any thoughts you have, or if you'd like me to elaborate more, please let me know.
> > Thanks.
> 
> Thank you. These kind of deep knowledge about Yocto is really helpful.

Wild idea: Could we link our PURLs to our SBoMs.... since they have way more context about what's going on than fits in a URL

jloehel:
> Wild idea: Could we link our PURLs to our SBoMs.... since they have way more context about what's going on than fits in a URL

Can you explain this a little bit? I thought it's the opposite. The component of the SBOM should have a PURL.

~~~
...
    {
      "name": "python3-dbus",
      "version": "1.2.18",
      "type": "library",
      "cpe": "cpe:2.3:*:*:python3-dbus:1.2.18:*:*:*:*:*:*:*",
      "purl": "pkg:yocto/python3-dbus@1.2.18",
      "bom-ref": "a27c313f-0c89-4f9b-ac0d-35e0e6b67a9c"
    },
...
~~~

Theoretically it should be possible to search for: `pkg:yocto/python3-dbus@1.2.18` and get all SBOMs which include this component.

JPEWdev:
> > Wild idea: Could we link our PURLs to our SBoMs.... since they have way more context about what's going on than fits in a URL
> 
> Can you explain this a little bit? I thought it's the opposite. The component of the SBOM should have a PURL.
> 
> ```
> ...
>     {
>       "name": "python3-dbus",
>       "version": "1.2.18",
>       "type": "library",
>       "cpe": "cpe:2.3:*:*:python3-dbus:1.2.18:*:*:*:*:*:*:*",
>       "purl": "pkg:yocto/python3-dbus@1.2.18",
>       "bom-ref": "a27c313f-0c89-4f9b-ac0d-35e0e6b67a9c"
>     },
> ...
> ```
> 
> Theoretically it should be possible to search for: `pkg:yocto/python3-dbus@1.2.18` and get all SBOMs which include this component.

Hmm... ya. I think you could do that, with something like:
```json
{
   "type": "software_Package",
   "spdxId": "", # NOTE: This is universally unique
   "name": "python3-dbus",
   "software_packageVersion": "1.2.18",
   "software_packageUrl": "spdx://software_Package/python3-dbus@1.2.18"
}
```

Externally, you could use the PURL `spdx://software_Package/python3-dbus@1.2.18` to find all SBoM references to *any* matching package, or you could reference the specific item with: `spdx://software_Package/python3-dbus@1.2.18?spdxId=https%3A//spdx.org/package/python3-dbus/537f01de-2a54-42fb-8494-64168a9eb9e6&url=https%3A//spdx.org/example.spdx.json`

Note that in SPDX, `spdxId` is universally unique, and always refers to that specific element (which ultimately means it's the minimum required to refer to an object); `url` would the parameter that refers to the location where the document containing that element can be found (`spdxId` is not required to resolve to an actual download location, although it can).

I don't know if this makes any sense in the PURL; my main goal is that our SBoMs are pretty complete and comprehensive, so if we can make PURLs reference that it can save us some trouble of trying to encode too much in the limited space that PURLs give us.

FYI, I'm heavily involved in the SPDX community also :)

gernot-h:
> > > 'm not exactly clear on what purpose of a recipe PURL would be in this context; is the intent to be able to match to PURL across other distros (e.g. Debian, Fedora, etc.) so that we can more or less tell if "my busybox is the same as yours?", and thus by extension to tell if a given CVE applies?
> > 
> > 
> > The purpose for me would be to download the recipe or better clone the corresponding layer. Checkout the correct version. Parse/Interpret the recipe with bitbake. Identify patches, CVEs, licenses and the source code (upstream tarball, configuration files, ...). Enrich a SBOM with the patches and the licenses. Enrich the VEX with the CVEs and add some basic analysis based on the patches. Download the source code and run further analysis on the source and enrich the SBOM/Vex. I do not care so much about bbappend files at this point. The objective is, to enrich an SBOM/VEX automatically as much as possible. If there are special bbappend files, these can be handled later manually by an analyst. ... or add them as file components to the SBOM.
> 
> Ok, so if I'm reading this correctly, the purpose of a PURL in this case (or maybe all cases?) is to identify a component in a succinct manner such that some one can come along later and perform some level of supply chain analysis on it using (basically) only the PURL (or more specifically, what the PURL points to)?
> 
> That didn't fit my conceptual model of what PURLs were supposed to be, but that's OK. I had thought they were an attempt to "correlate" things across different sources, e.g. fedora "busybox", debian "busybox" and yocto "busybox" might be configured differently, but they are all still busybox, so conclusions you make about (might) hold on all of them (or are at least worth looking at).

Identifying components is also our use case, hope I'm not boring you when describing it quickly. @petermarko and myself are developers in a large corporation. We have to track all integrated 3rd party components in advanced internal tools so that our security and legal colleagues can check the license compliance and monitor components for security incidents. In the last decade, PURL has established as *the* unique identifier to handle these topics for many Linux and language components like deb, rpm, pypi, npm etc. 

While we would love to correlate packages between those ecosystems as you mention, this is still a large manual effort anyways, given that distributions have own naming schemes for upstream components (foo vs. python-foo vs. libfoo-python vs...) and somtimes also bundle or split components in different ways. So main purpose is really to uniquely identify a component version (down to the patches) *in one ecosystem* and especially identify synergies to other products using the same package, so legal and sec teams only have to check license compliance and CVEs once for multiple products.

That said, speaking names and version numbers in a PURL would still help us greatly in half-automatic or manual matching processes. While we treat busybox patches individually for each Linux distribution, it's still very helpful to know it's busybox 1.35 in your whatever-image as 95% of license considerations and CVEs still apply. So I would really prefer common names and package versions in the PURL and would prefer recipe names, git hashes etc in the PURL qualifiers so humans have a chance to quickly deduce from the PURL what component they are seeing.

Also don't need to worry too much about adding superfluous PURL qualifiers. Given we can somehow sort them by relevance, we can easily decide which ones we consider when matching. We have to do this anyways already, e.g. for Debian with its shared package pool, a version is unique and we can safely ignore the `distro` the Syft scanner likes to add – while for Alpine, we have to respect the `distro` qualifier as same package version can mean different patches in different Alpine releases. 

> > > How do other Distros deal with multiple binary packages from the same "source"? For example, Ubuntu has `bash`, `bash-static`, `bash-doc` etc.? If all of these get the same PURL, than I think a recipe level PURL is probably the direction we need to go, if not.... I'm not really sure what we can do because Yocto would package the same build into completely different packages.
> > 
> > 
> > I am not sure how to handle sub-packages. But probably not handled via PURL.
> 
> Ack. Recipes _probably_ make the most sense then. I think the weird thing here might be that recipes are a purely build system concept and there is indirection to the final thing (package) the end user gets, unlike some other PURLs which seem to focus on only the end user package?

For Debian, we contributed the qualifier `arch=source` to PURL and I'm currently [trying to add it for other distros](https://github.com/package-url/purl-spec/pull/312). That allows us to express in a PURL whether we list source or binary package PURLs in our SBOM. Mapping between them is however out of scope of the PURL. 

I think one of the central ideas which made PURL a success is that it doesn't try to provide yet another global way of naming components and versions, but instead rely on established ecosystems. A PURL only tells you that this is the Debian binary package foo, and (only) a Debian-specific tool can then resolve what this actually means.

> Ack. Its going to be quite hard to accurately capture in a single URL... people can and do patch CVEs in bbappend and what not, so without all the context it's difficult to draw accurate conclusions from just a PURL (or what the PURL points to). I'll give it some more thought though.

I'm unfortunately no Yocto expert, but we only need to think about sane use cases. I guess, an educated developer would increase the package version or some git hash would change? For other Linux distros, we also have to rely that developers change a package version when they recompile a package with own patches.

gernot-h:
> Externally, you could use the PURL `spdx://software_Package/python3-dbus@1.2.18` to find all SBoM references to _any_ matching package, or you could reference the specific item with: `spdx://software_Package/python3-dbus@1.2.18?spdxId=https%3A//spdx.org/package/python3-dbus/537f01de-2a54-42fb-8494-64168a9eb9e6&url=https%3A//spdx.org/example.spdx.json`

Hmm, not sure about this. In our workflows, we look on a PURL, part by part. The type tells us which kind of plugin to call to decode the PURL (Debian, PyPI etc.) and then the plugin looks on the namespace, name, version and probably qualifiers to identify the used component and its source. Normally, > 80% of the components of our SBOMs are Open Source component versions, so the plugin can decode them without any project-specific knowledge and retrieve sources and CVEs from public resources. In your example, we would need to read project-specific SBOMs to identify what was used actually making sharing results between projects much harder. Or are you referring to publicly available SPDX SBOMs here which could be seen as unique identifier on their own?

JPEWdev:
@gernot-h 

Ok. Based on this conversation, I think a few things are apparent: 
1. You'll never get the precise level of detail you are looking for (e.g. down to the patches) for a Yocto recipe in a PURL. It's just too variable to neatly encode in a single URL. If you want precision, you _must_ look at the SBoM
2. We need something fairly fixed (for a given recipe) so that we can at least at a high enough level, find information (e.g. CVEs etc.) about the recipe. Because of 1, we just can't get too precise, but we need enought to figure it out in the first place.

Therefore, I think my suggestion would be something like this:
```
yocto://recipe/<LAYERNAME>/<PN>@<PV>?layerurl=<LAYERGITURL>&layersha1=<LAYERSHA1>
```
where:
- `LAYERNAME` is the "layer name" of the layer (as found in the layer.conf), which will help differentiate the same recipe in different layers
- `PN` is the recipe name
- `PV` is the recipe version
- `LAYERGITURL` is the git URL of the layer (optional)
- `LAYERSHA1` is the SHA1 of the layer git repo checked out (optional)

It's not perfect, and it can't give you everything you need, since other layers may affect the recipe in unexpected ways, _but_ it does give you a consistent base to start from.

If you need any more information than this, or if this isn't precise enough, than we need to link to the SBoM with additional qualifiers, for example, something like:

```
yocto://recipe/core/python3-dbus@1.2.18?spdxId=https%3A//spdx.org/package/python3-dbus/537f01de-2a54-42fb-8494-64168a9eb9e6&spdxurl=https%3A//spdx.org/example.spdx.json
```

And yes, this does require that the SBoM be published somewhere where it is publically accessible, but I think that's OK, because you've _always_ been at the mercy of whoever compiled the software to publish that information anyway. It's not like if you took a random Yocto .deb/.rpm/.ipk package off a package feed server it would magically tell you about all the layers used to compile it.

jloehel:
I have made a few changes. I think we can now all agree to use PV as version and not the commit hash. The "commit hash" of the layer repo is now a qualifier and optional. I am not sure yet about the naming of the qualifier. I chose `layer_version` but it is open for discussions. I am not convinced about `layersha1` because it should be also possible to use a branch or tag here. It would increase the readability.

Should be something like this allowed:
~~~
pkg:yocto/openembedded-core/glibc@2.35&layer_version=kirkstone
~~~

@JPEWdev I really like the idea for a back-reference to the SBOM. But I think it needs to be more general to support also cyclonedx and other well-known formats. Maybe just `sbom_id` and `sbom_url`.

I am not really sure about the subpath. In example `https://layers.openembedded.org` specifies also the subpath or subdirectory if necessary. But I have removed it for now.

PURL-TYPES.rst
```diff
@@ -582,6 +582,50 @@ swift
...
 
+
+YOCTO Linux
+-----------
+``yocto`` for Yocto Linux recipes:
+
+There is no default package repository. The classifier `repository_url` is optional.
+
+:namespace: The namespace is the name of the layer which provides the recipe. The layer name needs
+to match with the name used in the `meta-<namespace>/conf/layer.conf`.
```
JPEWdev, line 593: `meta-<namespace>` is a common convention to name layers when users download them, but is not mandatory. The better way to describe this would be:

```
The layer name as specified in the `BBFILE_COLLECTIONS` variable in `conf/layer.conf` of the layer.
```

PURL-TYPES.rst
```diff
@@ -582,6 +582,50 @@ swift
...
+  * ``https%3A%2F%2Fgit.openembedded.org%2Fopenembedded-core``
+  * ``https%3A%2F%2Fgithub.com%2Fakuster%2Fmeta-odroid``
+
+  The protocol can be ``https``, ``ssh``  or ``git`` and is mandatory.
+
+``layer_version``
+  layer_version is the version of the yocto layer which is a SHA1 commit, tag or branch
+  of the yocto layer GIT repository. It should support also [short commits](https://git-scm.com/book/en/v2/Git-Tools-Revision-Selection#_short_sha_1). (optional)
+``arch``
+  Optional for the architecture like ``mips64`` or ``arm64``
```
JPEWdev, line 614: Is this actually useful?

jloehel: I think we can get rid of it.

PURL-TYPES.rst
jloehel: Done!

gernot-h:
I really like the current state, thanks @jloehel, @JPEWdev! :dart:

PURL-TYPES.rst
```diff
@@ -582,6 +582,48 @@ swift
...
+All qualifiers should be percent-encoded.
+
+``repository_url``
+  The GIT URL of the layer.
+  In example:
+
+  * ``https%3A%2F%2Fgit.openembedded.org%2Fopenembedded-core``
+  * ``https%3A%2F%2Fgithub.com%2Fakuster%2Fmeta-odroid``
+
+  The protocol can be ``https``, ``ssh``  or ``git`` and is mandatory.
```
JPEWdev, line 608: Does this have to be mandatory? It seems like it _could_ be deduced from the namespace

PURL-TYPES.rst
```diff
@@ -582,6 +582,48 @@ swift
...
+
+YOCTO Linux
+-----------
+``yocto`` for Yocto Linux recipes:
+
+There is no default package repository. The classifier `repository_url` is optional.
+
+:namespace: The namespace is the name of the layer which provides the recipe. The layer name as
+specified in the `BBFILE_COLLECTIONS` variable in `conf/layer.conf` of the layer.
+:name: The name of the package also known as [PN](https://docs.yoctoproject.org/ref-manual/variables.html#term-PN) in a yocto recipe.
```
JPEWdev, line 594: `BPN` might be more appropriate since it covers `target`, `native` and `nativesdk` builds

PURL-TYPES.rst
```diff
@@ -582,6 +582,48 @@ swift
...
+``layer_version``
+  layer_version is the version of the yocto layer which is a SHA1 commit, tag or branch
+  of the yocto layer GIT repository. It should support also [short commits](https://git-scm.com/book/en/v2/Git-Tools-Revision-Selection#_short_sha_1). (optional)
+
+Examples
+^^^^^^^^
+
+::
+
+  pkg:yocto/openembedded-core/glibc@2.35
```
JPEWdev, line 619: openembedded-core's layername is just `core`

Also, this example doesn't have `repository_url` :)

test-suite-data.json
```diff
@@ -706,5 +706,20 @@
     "qualifiers": null,
     "subpath": null,
     "is_invalid": true
+  },
+  {
+    "purl": "pkg:yocto/openembedded-core/glibc@2.35&repository_url=https%3A%2F%2Fgit.openembedded.org%2Fopenembedded-core&layer_version=kirkstone",
```
JPEWdev, line 711: `core` instead of `openembedded-core`

petermarko: I'm not quite sure here...
`BBLAYERS_LAYERINDEX_NAME_core = "openembedded-core"`

PURL-TYPES.rst
jloehel: I thought the `repository_url ` should be optional. I wanted to show also an example without it to emphasize that this is also possible.

test-suite-data.json
petermarko: Also this says that the name should be `openembedded-core` and not `core`: https://github.com/yoctoproject/poky/blob/master/meta/classes/toaster.bbclass#L45

jloehel: He is right, because we use `BBFILE_COLLECTIONS`:  https://git.openembedded.org/openembedded-core/tree/meta/conf/layer.conf#n6

JPEWdev: > I'm not quite sure here... `BBLAYERS_LAYERINDEX_NAME_core = "openembedded-core"`

That's the name known to the layer index, which isn't necessarily the "layer name". Not all layers are published to the layer index, so using that name is probably unwise (unless it's completely optional to specify it....). Bitbake knows this layer by the `core` name (as is apparent in the suffix of the `BBLAYERS_LAYERINDEX_NAME` variable _et. al._), and that named is required (as much as anything) and needs to be unique among all layers. Finally, bitbake presents this name to the recipe code when it is parsed as `FILE_LAYERNAME`, so we can construct a valid PURL with a minimal: `pkg:yocto/${FILE_LAYERNAME}/${BPN}@${PV}`

PURL-TYPES.rst
JPEWdev: Ah, I see. I think I was misinterpreting the "is mandatory" part to apply to the whole `repository_url`, not just the protocol

PURL-TYPES.rst
JPEWdev: Ah, I misunderstood this, see below

jloehel: But this means that it can be misunderstood by someone else, too. What would be a better way to write it?

> The URL scheme is mandatory. The URL schema can be: `https`, `http`, `ssh`, `git` ...

PURL-TYPES.rst
```diff
@@ -582,6 +582,48 @@ swift
...
+specified in the `BBFILE_COLLECTIONS` variable in `conf/layer.conf` of the layer.
+:name: The name of the package also known as [PN](https://docs.yoctoproject.org/ref-manual/variables.html#term-PN) in a yocto recipe.
+:version: The version of the package also known as [PV](https://docs.yoctoproject.org/ref-manual/variables.html#term-PV) in a yocto recipe.
+
+Qualifiers
+^^^^^^^^^^
+All qualifiers should be percent-encoded.
+
+``repository_url``
+  The GIT URL of the layer.
```
JPEWdev, line 602: "The git URL of the layer. (optional)"

PURL-TYPES.rst
JPEWdev: Maybe the above to clarify `repository_url` is optional, and then:

"The URL scheme is mandatory and must be one of `https`, `http`, `ssh`, or `git`."

?

PURL-TYPES.rst
jloehel: I hope it's more clear now.

JPEWdev (approved):
LGTM, thanks!

petermarko:
Thank you all for great work on this specification.
I think it's now in a very good shape with one exception.

We have been discussing this internally in our project and we think that the current state is still not reflecting fact that Yocto is not a distribution but a tool to create your own distribution.
Therefore we'd love if additional qualifier(s) which would specify the distro could be introduced.
Basically it would allow vendors to say that they have modified the original Yocto package (similar as Yocto has modified original source package).
Note that also the definition of purl qualifiers contains distro as example what it should be used for.

The idea about link to SBOM is the most appealing as that will hold the accurate information about the component state, even if we would need dedicated servers to deliver our SBOMs.
But maybe it deviates from purl standards too much?

Listing the distro (name, url, version) would also be a step forward from current situation.

Currently I think there are two possibilities for vendors which are very sub-optimal, hopefully we won't have to use them:
* modify PV with a vendor specific suffix, but 1. without any possibility to add a reference to what that means 2. it will violate the spec because there is no package version with such vendor suffix in core
* change the namespace for all/many from e.g. `core` to `vendor-distro` which will violate the spec which says that it should be the layer where the recipe is located

JPEWdev:
> Thank you all for great work on this specification. I think it's now in a very good shape with one exception.
> 
> We have been discussing this internally in our project and we think that the current state is still not reflecting fact that Yocto is not a distribution but a tool to create your own distribution. Therefore we'd love if additional qualifier(s) which would specify the distro could be introduced. Basically it would allow vendors to say that they have modified the original Yocto package (similar as Yocto has modified original source package). Note that also the definition of purl qualifiers contains distro as example what it should be used for.
> 
> The idea about link to SBOM is the most appealing as that will hold the accurate information about the component state, even if we would need dedicated servers to deliver our SBOMs. But maybe it deviates from purl standards too much?
> 
> Listing the distro (name, url, version) would also be a step forward from current situation.
> 
> Currently I think there are two possibilities for vendors which are very sub-optimal, hopefully we won't have to use them:
> 
>     * modify PV with a vendor specific suffix, but 1. without any possibility to add a reference to what that means 2. it will violate the spec because there is no package version with such vendor suffix in core
> 
>     * change the namespace for all/many from e.g. `core` to `vendor-distro` which will violate the spec which says that it should be the layer where the recipe is located

I don't think we need to capture distro, for a number of reasons:
1) Distro are actually optional in Yocto, you don't _have_ to use one, so it gets really tricky
2) At the level we are specifying here (the recipe), distro isn't going to be relevant, and might actually be harmful for attempting to match a recipe to a purl.
3) I think what you really want is to attach a purl to your final binary package (as discussed above) using rpm/deb/etc. and that _should_ reference your distro

jloehel:
@JPEWdev Could you please share a minimal SBOM which gets generated by your PoC.

JPEWdev:
> @JPEWdev Could you please share a minimal SBOM which gets generated by your PoC.

Here is a snipet from an SPDX 3 SBoM that shows how the PURLs would look. Note that they appear under `externalIdentifier` and also a single one shows up in `software_packageUrl`

```json
    {
      "type": "software_Package",
      "spdxId": "http://spdx.org/spdxdocs/acl-3a7084c7-0f68-55ec-aaab-f9397756bc81/989a5099bc781050044d6e39055e45cc811d40c434e9d45274de1667f0f8e021/package/libacl1",
      "creationInfo": "_:CreationInfo244",
      "description": "ACL allows you to provide different levels of access to files and folders for different users.",
      "externalIdentifier": [
        {
          "type": "ExternalIdentifier",
          "externalIdentifierType": "packageUrl",
          "identifier": "pkg:yocto/core/acl@2.3.2"
        },
        {
          "type": "ExternalIdentifier",
          "externalIdentifierType": "packageUrl",
          "identifier": "pkg:yocto/core/acl@2.3.2?layer_version=91f31373efa7f6aee5dc327f6191ba1987f79944&repository_url=git%3A%2F%2Fgit.yoctoproject.org%2Fpoky-contrib"
        },
        {
          "type": "ExternalIdentifier",
          "externalIdentifierType": "packageUrl",
          "identifier": "pkg:yocto/core/acl@2.3.2?layer_version=jpew%2Fpurls&repository_url=git%3A%2F%2Fgit.yoctoproject.org%2Fpoky-contrib"
        },
        {
          "type": "ExternalIdentifier",
          "externalIdentifierType": "cpe23",
          "identifier": "cpe:2.3:*:*:acl:2.3.2:*:*:*:*:*:*:*"
        }
      ],
      "name": "libacl1",
      "summary": "Utilities for managing POSIX Access Control Lists",
      "software_primaryPurpose": "install",
      "software_homePage": "http://savannah.nongnu.org/projects/acl/",
      "software_packageUrl": "pkg:yocto/core/acl@2.3.2",
      "software_packageVersion": "2.3.2"
    }
```

petermarko:
@JPEWdev, some thought from my side

>  Distro are actually optional in Yocto, you don't have to use one, so it gets really tricky

I thought about the distro as an optional purl qualifier, so it doesn't need to be set in purl if it's not wanted.

But I could also quote Yocto landing page - "The Yocto Project - It’s not an embedded Linux distribution, it creates a custom one for you."
So by definition, purl string saying it's an unmodified Yocto distribution is not correct.

> At the level we are specifying here (the recipe), distro isn't going to be relevant, and might actually be harmful for attempting to match a recipe to a purl.

Maybe I'm seeing this from a wrong perspective, but I think that purl spec without distro would be only academical thing without any real use in products.
Basically all products based on Yocto need to do their bbappends and packageconfigs to get their features/behavior/security for several recipes.
And having purl referencing pure recipe without at least a flag that this was modified would be a lie when delivering to customers.

Of course we can say that it's not needed and SBOM is the thing the customer should look at, not purl identifier.
But then why bothering to define the purl which does not have a capability to say that the component is modified compared to what the modifiers say?

> I think what you really want is to attach a purl to your final binary package (as discussed above) using rpm/deb/etc. and that should reference your distro

Well, with our package backend of choice - opkg, there is no purl type available (and generic doesn't look like a good one and it says it's for packages without distro)
Submitting new PR for opkg would be a possibility, but we actually don't have a binary distro, we only produce images while opkg/deb/rpm package is just an intermediate internal build detail similar to choice of autotools/meson/cmake.

JPEWdev:
> @JPEWdev, some thought from my side
> 
> > Distro are actually optional in Yocto, you don't have to use one, so it gets really tricky
> 
> I thought about the distro as an optional purl qualifier, so it doesn't need to be set in purl if it's not wanted.
> 
> But I could also quote Yocto landing page - "The Yocto Project - It’s not an embedded Linux distribution, it creates a custom one for you." So by definition, purl string saying it's an unmodified Yocto distribution is not correct.
> 
> > At the level we are specifying here (the recipe), distro isn't going to be relevant, and might actually be harmful for attempting to match a recipe to a purl.
> 
> Maybe I'm seeing this from a wrong perspective, but I think that purl spec without distro would be only academical thing without any real use in products. Basically all products based on Yocto need to do their bbappends and packageconfigs to get their features/behavior/security for several recipes. And having purl referencing pure recipe without at least a flag that this was modified would be a lie when delivering to customers.

But the distro isn't any more meaningful to the end user IMHO, unless you are making your entire Yocto setup public. And, more or less, the distro as an effect on the _output_ of the recipe, and I think these PURLs should focus on the recipe itself. Back to the bbappend, it's simply not possible to capture everything about the recipe in the PURL, so I don't think distro belongs either.

More the the point, my fear is that including the distro is actually counter productive for the purpose of determining that a recipe is used. Most or all _recipes_ can be used with any distro, so specifying the distro only provides a way for a PURL to specify a specific way a recipe is used, which prevents it from being useful for actually discovering the recipe in the first place. As a thought experiment, if PURLs are used in the future to identify software components affected by CVEs, inclusion of a distro would be a disaster. There are probably 100s-1000's of Yocto distros out there, so you couldn't possibly list them all, and inclusion of _any_ distro in the CVE PURL would mean false negatives on CVE matching. I appreciate that this may mean false positives if any actual distro mitigates the CVE, but A) this is better than false negatives, and B) we have mechanisms in the project to deal with this that work quite well.

As an analogy, lets say you're making a Ubuntu derived embedded project (I mean, I don't recommend this, but it's a thought experiment :)). Would you re-write the PURLs for all packages you installed to change the distro? Or even on a tighter scope, would you re-write the PURL for any package you modified? It seems analogous to this discussion, and TBH I'm not sure what the answer is.

I just can't see the usecase where distro provides actual useful and actionable information to the PURL, but maybe you can provide one?

> 
> Of course we can say that it's not needed and SBOM is the thing the customer should look at, not purl identifier. But then why bothering to define the purl which does not have a capability to say that the component is modified compared to what the modifiers say?
> 
> > I think what you really want is to attach a purl to your final binary package (as discussed above) using rpm/deb/etc. and that should reference your distro
> 
> Well, with our package backend of choice - opkg, there is no purl type available (and generic doesn't look like a good one and it says it's for packages without distro) Submitting new PR for opkg would be a possibility, but we actually don't have a binary distro, we only produce images while opkg/deb/rpm package is just an intermediate internal build detail similar to choice of autotools/meson/cmake.

ipk should be added as a PURL type, which would solve that

jloehel:
@johnmhoran I think the PR is ready for a first review. Can you please review the PR. :-)

PURL-TYPES.rst
```diff
@@ -582,6 +582,48 @@ swift
...
+YOCTO Linux
+-----------
+``yocto`` for Yocto Linux recipes:
+
+.. note::
+  There is no default package repository. The classifier `repository_url` is optional.
+
+:namespace *(optional)*: The namespace is the name of the layer which provides the recipe. The layer name as specified in the ``BBFILE_COLLECTIONS`` variable in ``conf/layer.conf`` of the layer.
+:name *(mandatory)*: The name of the package with common prefixes and suffixes removed, also known as `BPN <https://docs.yoctoproject.org/ref-manual/variables.html#term-BPN>`_ in a yocto recipe.
+:version *(optional)*: The version of the package also known as `PV <https://docs.yoctoproject.org/ref-manual/variables.html#term-PV>`_ in a yocto recipe.
```
petermarko: Looking at how GitHub renders these three lines, it's the only table in the document.
I'm not sure how is this later rendered in the official docs, but this does not look intentional per how this looks in the source file.

PURL-TYPES.rst
```diff
@@ -582,6 +582,48 @@ swift
...
+
+Qualifiers
```
petermarko: I'm not sure why I was chosen as a reviewer (but thanks a lot for the trust).
I'm still unhappy with missing a qualifier useful for derived distros.
But if the Yocto community does not want it, let's go with it and I will have to use a version modifier (e.g. `${PV}~my-distro-1`).
If this (having a version which is not available in linked namespace) is OK from spec point of view, I'll approve this PR.

jloehel:
Converted the rst description into the new JSON schema.

afreof:
Would be appreciated if this spec would be merged and Yocto could come up with the corresponding implementation for the next LTS release. @mprpic any concerns from your side?

mjherzog:
@JPEWdev Do you have any final comments / concerns about adding the `yocto `PURL type as represented in the current PR?

JPEWdev:
@mjherzog I think this is good enough for now. Thanks

mjherzog:
@JPEWdev Thank you for the quick answer.

johnmhoran (changes requested):
The percent-encoded colon ':' in the `repository_url` value is incorrect -- `&repository_url=https%3A%2F%2F` -- as the "Character encoding" section clearly states, 

`The following characters must not be percent-encoded: . . . the colon ':', whether used as a Separator Character or otherwise . . . .` 

https://github.com/package-url/purl-spec/blob/main/docs/standard/characters-and-encoding.md?plain=1#L44-L50

johnmhoran:
@jloehel My understanding is that the percent-encoded colon in the `qualifiers` value in a `base` test must result in a failed test.

jloehel:
> The percent-encoded colon ':' in the `repository_url` value is incorrect -- `&repository_url=https%3A%2F%2F` -- as the "Character encoding" section clearly states,
> 
> `The following characters must not be percent-encoded: . . . the colon ':', whether used as a Separator Character or otherwise . . . .`
> 
> [`main`/docs/standard/characters-and-encoding.md?plain=1#L44-L50](https://github.com/package-url/purl-spec/blob/main/docs/standard/characters-and-encoding.md?plain=1&rgh-link-date=2025-11-10T22%3A56%3A10Z#L44-L50)

I changed it. If I missed something, pls ping me.

JPEWdev:
Do we need any other changes or can this be merged?

mjherzog:
@JPEWdev I have asked @johnmhoran to review today so that we can merge.
One secondary request - We are working on a Package-URL website where we have a grid of registered PURL types at https://package-url.github.io/packageurl.org/docs/purl/purl-spec-purl-types
Can you point us to an icon to use for Yocto there? I was not able to find a branding or similar section on the main Yocto website.

JPEWdev:
@mjherzog The Yocto Project Logo and guidelines can be found here: https://www.yoctoproject.org/brand-guidelines/ . If you have questions or concerns, let me know and I can get you in touch with someone who can help.

mjherzog (approved):
Looks good

mjherzog:
@JPEWdev Per the brand guidelines it looks like we will need to submit a request for logo usage.

PR #373, opened 2024-12-24
Add RFC 2119/8174 note and convert terms to uppercase as needed

johnmhoran:
As part of the core spec updating process, I've created a new branch and PR to address the proposed adoption and implementation of RFC 2119, as clarified by RFC 8174.

### DIFF
--- a/PURL-SPECIFICATION.rst
+++ b/PURL-SPECIFICATION.rst
@@ -4,6 +4,12 @@ Package URL specification v1.0.X
 The Package URL core specification defines a versioned and formalized format,
 syntax, and rules used to represent and validate ``purl``.
 
+The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL
+NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED",  "MAY", and
+"OPTIONAL" in this document are to be interpreted as described in
+`RFC 2119 <https://datatracker.ietf.org/doc/html/rfc2119>`_ and
+clarified in `RFC 8174 <https://datatracker.ietf.org/doc/html/rfc8174>`_.
+
 A ``purl`` or package URL is an attempt to standardize existing approaches to
 reliably identify and locate software packages.
 
@@ -32,24 +38,24 @@ The definition for each components is:
 
 - **scheme**: this is the URL scheme with the constant value of "pkg". One of
   the primary reason for this single scheme is to facilitate the future official
-  registration of the "pkg" scheme for package URLs. Required.
+  registration of the "pkg" scheme for package URLs. REQUIRED.
 - **type**: the package "type" or package "protocol" such as maven, npm, nuget,
-  gem, pypi, etc. Required.
+  gem, pypi, etc. REQUIRED.
 - **namespace**: some name prefix such as a Maven groupid, a Docker image owner,
-  a GitHub user or organization. Optional and type-specific.
-- **name**: the name of the package. Required.
-- **version**: the version of the package. Optional.
+  a GitHub user or organization. OPTIONAL and type-specific.
+- **name**: the name of the package. REQUIRED.
+- **version**: the version of the package. OPTIONAL.
 - **qualifiers**: extra qualifying data for a package such as an OS,
-  architecture, a distro, etc. Optional and type-specific.
+  architecture, a distro, etc. OPTIONAL and type-specific.
 - **subpath**: extra subpath within a package, relative to the package root.
-  Optional.
+  OPTIONAL.
 
 
 Components are designed such that they form a hierarchy from the most significant
 on the left to the least significant components on the right.
 
 
-A ``purl`` must NOT contain a URL Authority i.e. there is no support for
+A ``purl`` MUST NOT contain a URL Authority i.e. there is no support for
 ``username``, ``password``, ``host`` and ``port`` components. A ``namespace`` segment may
 sometimes look like a ``host`` but its interpretation is specific to a ``type``.
 
@@ -98,14 +104,14 @@ A ``purl`` is a URL
 - Special URL schemes as defined in https://url.spec.whatwg.org/ such as
   ``file://``, ``https://``, ``http://`` and ``ftp://`` are NOT valid ``purl`` types.
   They are valid URL or URI schemes but they are not ``purl``.
-  They may be used to reference URLs in separate attributes outside of a ``purl``
+  They MAY be used to reference URLs in separate attributes outside of a ``purl``
   or in a ``purl`` qualifier.
 
 - Version control system (VCS) URLs such ``git://``, ``svn://``, ``hg://`` or as
   defined in Python pip or SPDX download locations are NOT valid ``purl`` types.
   They are valid URL or URI schemes but they are not ``purl``.
   They are a closely related, compact and uniform way to reference VCS URLs.
-  They may be used as references in separate attributes outside of a ``purl`` or
+  They MAY be used as references in separate attributes outside of a ``purl`` or
   in a ``purl`` qualifier.
 
 
@@ -115,7 +121,7 @@ Rules for each ``purl`` component
 A ``purl`` string is an ASCII URL string composed of seven components.
 
 Some components are allowed to use other characters beyond ASCII: these
-components must then be UTF-8-encoded strings and percent-encoded as defined in
+components MUST then be UTF-8-encoded strings and percent-encoded as defined in
 the "Character encoding" section.
 
 The rules for each component are:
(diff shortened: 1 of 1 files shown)


PURL-SPECIFICATION.rst
```diff
@@ -219,50 +225,50 @@ The rules for each component are:
...
 - the '/' used as ``type``/``namespace``/``name`` and ``subpath`` segments separator
-  does not need to and must NOT be percent-encoded. It is unambiguous unencoded
+  does not need to and MUST NOT be percent-encoded. It is unambiguous unencoded
   everywhere
 
-- the '@' ``version`` separator must be encoded as ``%40`` elsewhere
-- the '?' ``qualifiers`` separator must be encoded as ``%3F`` elsewhere
-- the '=' ``qualifiers`` key/value separator must NOT be encoded
-- the '#' ``subpath`` separator must be encoded as ``%23`` elsewhere
+- the '@' ``version`` separator MUST be encoded as ``%40`` elsewhere
```
matt-phylum, line 263: Most of these are inaccurate when they say that encoding is required. It MUST be encoded in canonical PURLs, but in general it MUST be encoded if it is the right-most `@` and isn't the version separator and SHOULD be encoded otherwise. Parser implementations are required to handle the case where the PURL contains unencoded `@` characters as described in the parsing algorithm section.

PURL-SPECIFICATION.rst
```diff
@@ -273,7 +279,7 @@ How to build ``purl`` string from its components
 Building a ``purl`` ASCII string works from left to right, from ``type`` to
 ``subpath``.
 
-Note: some extra type-specific normalizations are required.
+Note: some extra type-specific normalizations are REQUIRED.
```
matt-phylum, line 282: This is an impossible requirement if the addition of new types is considered to be not a breaking change. More practically, normalization is REQUIRED during canonicalization and comparison, but otherwise implementations should do their best.

PURL-SPECIFICATION.rst
```diff
@@ -427,11 +433,11 @@ Known ``qualifiers`` key/value pairs
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
 Note: Do not abuse ``qualifiers``: it can be tempting to use many qualifier
-keys but their usage should be limited to the bare minimum for proper package
+keys but their usage SHOULD be limited to the bare minimum for proper package
```
matt-phylum, line 436: This is an instruction to PURL spec authors. PURL implementations shouldn't be removing qualifiers to make the PURL more compact.

PURL-SPECIFICATION.rst
```diff
@@ -482,22 +488,22 @@ pairs some of which may not be normalized:
 - **qualifiers**: the ``qualifiers`` corresponding to this ``purl`` as an object of
   {key: value} qualifier pairs.
 - **subpath**: the ``subpath`` corresponding to this ``purl``.
-- **is_invalid**: a boolean flag set to true if the test should report an
+- **is_invalid**: a boolean flag set to true if the test SHOULD report an
```
matt-phylum, line 491: This is incorrect. The test MUST report an error if the error is related to general PURL problems or if the error is related to a package type problem for a package type that is supported by the implementation being tested.

PURL-SPECIFICATION.rst
```diff
@@ -482,22 +488,22 @@ pairs some of which may not be normalized:
...
-- parsing the test ``purl`` should return the components parsed from the test
+- parsing the test ``purl`` SHOULD return the components parsed from the test
   canonical ``purl``
 
 - parsing the test ``purl`` then re-building a ``purl`` from these parsed components
-  should return the test canonical ``purl``
+  SHOULD return the test canonical ``purl``
 
-- building a ``purl`` from the test components should return the test canonical ``purl``
+- building a ``purl`` from the test components SHOULD return the test canonical ``purl``
```
matt-phylum, line 506: All of these are qualified MUSTs. The implementation MUST return the value specified by the test if the implementation supports the package type used by that test. The test suite could use some work to separate the behaviors that any PURL implementation must satisfy from the behaviors that PURL implementations supporting certain package types must satisfy.

PURL-SPECIFICATION.rst
```diff
@@ -32,24 +38,24 @@ The definition for each components is:
...
-  registration of the "pkg" scheme for package URLs. Required.
+  registration of the "pkg" scheme for package URLs. REQUIRED.
 - **type**: the package "type" or package "protocol" such as maven, npm, nuget,
-  gem, pypi, etc. Required.
+  gem, pypi, etc. REQUIRED.
 - **namespace**: some name prefix such as a Maven groupid, a Docker image owner,
-  a GitHub user or organization. Optional and type-specific.
-- **name**: the name of the package. Required.
-- **version**: the version of the package. Optional.
+  a GitHub user or organization. OPTIONAL and type-specific.
```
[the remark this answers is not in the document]
pombredanne: @matt-phylum your comment make sense, but they apply to package types. Here @johnmhoran is not trying to change the spec, but just adopt the well defined words of the referenced RFCs. 

> The namespace is part of the name, but for some reason documented separately from the name field, and it is always required, but sometimes must be empty. 

We should consider to merge namespace and name in one thing alright, but that's not the goal here. This will come later. And the namespace was never specified so far as being something possibly empty (with one exception which is a bug for the "mlflow" type https://github.com/package-url/purl-spec/blame/1951d217bde29590a73f075db4ab71cc00011459/PURL-TYPES.rst#L419 and I submitted a separate PR to remove this "empty" namespace https://github.com/package-url/purl-spec/pull/374  )

> For pkg:nuget, the namespace MUST be empty

I'd rather say that there is no namespace at all for nuget. 
> There is no namespace per se even if the common convention is to use dot-separated package names where the first segment is namespace-like.

An empty namespace could mean something like this with really an "empty" namespace `pkg:nuget//System.IO.FileSystem@4.3.0` which does not make sense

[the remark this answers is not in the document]
matt-phylum: PURL treats empty and missing namespaces the same. You can say that either all PURLs have a namespace, which may be empty, or you can say that the namespace cannot be empty and some PURLs do not have a namespace. It's not valid to encode an empty namespace as `pkg:nuget//System.IO.FileSystem@4.3.0` according to the "How to build" section.

It's not just a problem with the types spec. This file says here that the namespace is optional, but in other places it says "if the namespace is not empty."

But anyway I'm trying to say that here the namespace is documented as being optional, but in the current spec it's either required to exist and be non-empty (Maven), required to not exist or be empty (NuGet), required to match a property of the package which may not exist or may be empty for certain packages (NPM). I feel like saying here that it's OPTIONAL without any qualifiers implies that for Maven you could leave it out or for NuGet you could add one.

I guess it would make more sense if PURL was more like URL where there are URL parsers that don't understand eg ftp://, but with the way the PURL spec is currently written, the line between supporting PURL and supporting a particular package type is blurred by requiring PURL parsing and building implementations to understand type specific rules for validating and normalizing the PURL.

PURL-SPECIFICATION.rst
```diff
@@ -115,27 +121,27 @@ Rules for each ``purl`` component
...
 the "Character encoding" section.
 
 The rules for each component are:
 
 - **scheme**:
 
   - The ``scheme`` is a constant with the value "pkg"
-  - Since a ``purl`` never contains a URL Authority, its ``scheme`` must not be
-    suffixed with double slash as in 'pkg://' and should use instead
+  - Since a ``purl`` never contains a URL Authority, its ``scheme`` MUST NOT be
```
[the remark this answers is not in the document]
pombredanne: @matt-phylum You are making good points! I think we can make the spec crisper by separating:
- A strict core spec, one that defines what a purl is and for instance here stating that a purl must start with `pkg:`, without further details.
- And other (new) sections like an FAQ and recommendations to parser implementers that could explain here that a parser should be flexible and recover from some mistakes including handling `pkg://`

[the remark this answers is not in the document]
johnmhoran: @matt-phylum  The approach @pombredanne suggests is the approach we're taking in the first of our component-focused PRs, [PR 361](https://github.com/package-url/purl-spec/pull/361#issuecomment-2560283119), which is focused on the 'scheme' component and includes inter alia the initial draft of an FAQ (faq.rst).

pombredanne:
@johnmhoran is this stale and no longer relevant in the latest spec? If yes, please close and leave unmerged.

johnmhoran:
Closing this as no longer needed -- the application of RFC 2119/8174 is being handled in individual focused pull requests.

PR #420, opened 2025-03-13
Updated CPAN type spec

giterlizzi:
This PR is for the new version of the purl-type CPAN specification and includes new updated tests.

https://github.com/giterlizzi/perl-URI-PackageURL/issues/8

In brief, this PR confirms the rules for generating a PURL string for CPAN.

Previously it was possible to create PURL strings in two ways:
- author (namespace) and distribution (name)
- module (name component).

Using only the module could create ambiguity, because in CPAN a distribution can have several authors (and co-authors) over time and does not allow precise identification of the artifact.

It was clarified that the author of the "distribution" is mandatory, allowing an artifact and its URL to be identified along with its version.

```console
$ purl-tool pkg:cpan/GDT/URI-PackageURL@2.22 --download-url
https://www.cpan.org/authors/id/G/GD/GDT/URI-PackageURL-2.22.tar.gz
```

### DIFF
--- a/tests/types/cpan-test.json
+++ b/tests/types/cpan-test.json
@@ -2,7 +2,7 @@
   "$schema": "https://packageurl.org/schemas/purl-test.schema-1.0.json",
   "tests": [
     {
-      "description": "cpan distribution name are case sensitive. Rountrip an input purl to canonical.",
+      "description": "CPAN distribution name are case sensitive. Rountrip an input purl to canonical.",
       "test_group": "advanced",
       "test_type": "roundtrip",
       "input": "pkg:cpan/DROLSKY/DateTime@1.55",
@@ -11,7 +11,7 @@
       "expected_failure_reason": null
     },
     {
-      "description": "cpan distribution name are case sensitive",
+      "description": "CPAN distribution name are case sensitive",
       "test_group": "base",
       "test_type": "parse",
       "input": "pkg:cpan/DROLSKY/DateTime@1.55",
@@ -27,7 +27,7 @@
       "expected_failure_reason": null
     },
     {
-      "description": "cpan distribution name are case sensitive. Rountrip a canonical input to canonical output.",
+      "description": "CPAN distribution name are case sensitive. Rountrip a canonical input to canonical output.",
       "test_group": "base",
       "test_type": "roundtrip",
       "input": "pkg:cpan/DROLSKY/DateTime@1.55",
@@ -36,7 +36,7 @@
       "expected_failure_reason": null
     },
     {
-      "description": "cpan distribution name are case sensitive",
+      "description": "CPAN distribution name are case sensitive",
       "test_group": "base",
       "test_type": "build",
       "input": {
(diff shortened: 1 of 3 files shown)


test-suite-data.json
```diff
@@ -576,31 +576,31 @@
...
+    "version": "6.76",
     "qualifiers": null,
     "subpath": null,
     "is_invalid": false
   },
   {
-    "description": "cpan module name are case sensitive",
-    "purl": "pkg:cpan/URI::PackageURL@2.11",
-    "canonical_purl": "pkg:cpan/URI::PackageURL@2.11",
+    "description": "CPAN distribution name are case sensitive",
```
sjn, line 591: `names` in plural, or `is` in singular

test-suite-data.json
```diff
@@ -612,38 +612,14 @@
     "is_invalid": true
   },
   {
-    "description": "cpan distribution name like module name",
-    "purl": "pkg:cpan/GDT/URI::PackageURL@2.11",
-    "canonical_purl": "pkg:cpan/GDT/URI::PackageURL",
+    "description": "CPAN distribution in BackPan repository",
```
sjn, line 615: BackPAN

sjn (commented):
Looks good! Just a few tweaks, and one question.

Would it make sense to add some failing tests?

e.g. the previous format with `Module::Names`, and perhaps a couple more uses, with a short description on why it is invalid?

test-suite-data.json
giterlizzi: FIXED typo

test-suite-data.json
giterlizzi: FIXED typo

test-suite-data.json
```diff
@@ -623,6 +623,18 @@
...
+    "purl": "pkg:cpan/GDT/URI::PackageURL",
+    "canonical_purl": "pkg:cpan/GDT/URI::PackageURL",
+    "type": "cpan",
+    "namespace": "GDT",
+    "name": "URI::PackageURL",
+    "version": null,
+    "qualifiers": null,
+    "subpath": null,
+    "is_invalid": true
+  },
```
sjn, line 637: Would it be sensible to also add a failing test for the previously accepted pURLs? e.g.
```
  {
    "description": "invalid CPAN with just the module name",
    "purl": "pkg:cpan/URI::PackageURL",
    "canonical_purl": "pkg:cpan/URI::PackageURL",
    "type": "cpan",
    "namespace": null,
    "name": "URI::PackageURL",
    "version": null,
    "qualifiers": null,
    "subpath": null,
    "is_invalid": true
  },
```
 …and maybe the variants that include a version number?

giterlizzi: right, the test "invalid CPAN distribution without author" cover only the "distribution" naming. added ;)

pombredanne:
@giterlizzi Thanks. We need to make sure we are all clear first wrt. the encoding in the core spec and that is getting close to merge ready in #398

giterlizzi (commented):
the PR now includes these new tests for the new CPAN specification:
- valid CPAN purl
- CPAN distribution in BackPAN repository
- invalid CPAN distribution name as module name

test-suite-data.json
```diff
@@ -624,28 +636,28 @@
...
     "type": "cpan",
     "namespace": null,
-    "name": "URI",
+    "name": "DBD::SQLite",
     "version": null,
     "qualifiers": null,
     "subpath": null,
-    "is_invalid": false
+    "is_invalid": true
   },
```
sjn: Would it be sensible to add a few more tests for corner cases?

* Distribution name with lowercase characters
* Distribution name which contains `::`

sjn: Looks otherwise good! :-)

PURL-TYPES.rst
```diff
@@ -170,15 +170,12 @@ conda
...
 
 - The default repository is ``https://www.cpan.org/``.
-- The ``namespace``:
-  - To refer to a CPAN distribution name, the ``namespace`` MUST be present. In this case, the namespace is the CPAN id of the author/publisher. It MUST be written uppercase, followed by the distribution name in the ``name`` component. A distribution name MUST NOT contain the string ``::``.
-  - To refer to a CPAN module, the ``namespace`` MUST be absent. The module name MAY contain zero or more ``::`` strings, and the module name MUST NOT contain a ``-``
-
-- The ``name`` is the module or distribution name and is case sensitive.
-- The ``version`` is the module or distribution version.
+- The ``namespace`` is the CPAN id of the author/publisher. It MUST be written uppercase and is required.
+- The ``name`` is the distribution name and is case sensitive.
```
sjn: Not sure, but would it make sense to still point out that…

> A distribution name MUST NOT contain the string ``::``.

…as earlier? I get it that it's strictly speaking not necessary, but as a hint for new users to understand the difference between a module name and a distribution name, I think it can be helpful.

(I have no strong opinions on this idea, and leave it to you to decide if this goes in or not)

sjn (approved):
Looks good to me, with a few final optional suggestions.

PURL-TYPES.rst
giterlizzi: I agree, adding this sentence (`A distribution name MUST NOT contain the string ::`) reinforces the concept that only the distribution MUST be used.

giterlizzi: "name" component description updated

pombredanne (changes requested):
Looking good, and this is a big change... could it make sense to also add a FAQ entry and give hints to implementers what they can do when parsing older `::` PURLs?

pombredanne:
- See  #514

With the new approach... this PR needs to be updated. Could you look into this? Thanks for your understanding!

giterlizzi:
@sjn @pombredanne Hi, i have updated the CPAN type spec and tests.

giterlizzi:
> Looking good, and this is a big change... could it make sense to also add a FAQ entry and give hints to implementers what they can do when parsing older `::` PURLs?

If @sjn agrees, I will add the "notes" field with this sentence:

The previous CPAN PURL type specification allowed module names (e.g. URI::PackageURL) to be used as PURL "names" while also omitting the PURL "namespace". The parser MUST emit an error when a module is specified as a PURL "name" or detect "::" characters.

pombredanne (approved):
@giterlizzi this looks all good to me... is there more that you need to add to this?

giterlizzi:
Nothing to add. Thanks ;)

sjn:
LGTM!

PR #422, opened 2025-03-14
[VERS] Add ecosystem-independent all and none schemes

pombredanne:
This introduces the possibility to specify all versions, or no version.

This replaces this PR, with edits and squashing:
- https://github.com/package-url/purl-spec/pull/343

Reference: https://github.com/package-url/purl-spec/issues/267
Reference: https://github.com/aboutcode-org/univers/pull/146

### DIFF
--- a/VERSION-RANGE-SPEC.rst
+++ b/VERSION-RANGE-SPEC.rst
@@ -637,12 +637,18 @@ These are a few known versioning schemes for some common Package URL
 - **alpine**: Alpine linux https://gitlab.alpinelinux.org/alpine/apk-tools/-/blob/master/src/version.c
   which is using Gentoo-like conventions.
 
+
+These are generic schemes, to use sparingly for special cases:
+
 - **generic**: a generic version comparison algorithm (which will be specified
   later, likely based on a split on any wholly alpha or wholly numeric segments
   and dealing with digit and string comparisons, like is done in libversion)
 
+- **none**: a generic versioning scheme for a range containing no version.
+  ``vers:none/*`` is the only valid vers form for this scheme.
 
-TODO: add Rust, composer and archlinux, nginx, tomcat, apache.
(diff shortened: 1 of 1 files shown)


pombredanne:
@immqu @tschmidtb51 review welcome, this is good to merge for me otherwise.

tschmidtb51 (commented):
I wonder whether we should add the line 243-247 from the original PR

```
Also, it is possible to specify a range that contains all versions or none:

- ``vers:all/*``
- ``vers:none/*``
```

or whether you left them out intentionally (which would also be fine for me).

immqu:
> I wonder whether we should add the line 243-247 from the original PR
> 
> ```
> Also, it is possible to specify a range that contains all versions or none:
> 
> - ``vers:all/*``
> - ``vers:none/*``
> ```
> 
> or whether you left them out intentionally (which would also be fine for me).

After reviewing the wording again, I did leave them out intentionally as it seemed misleading to me. The wording was ```Additionally, all versions (or none) SHOULD be specified:``` which implies that this versioning should always be used.

immqu (approved):

tschmidtb51 (approved):
LGTM

prabhu:
`all` here represents "all schemes" right?. Are we going to allow `vers:all` to be used (mis-used) to specify even the ranges?

tschmidtb51:
> `all` here represents "all schemes" right?. Are we going to allow `vers:all` to be used (mis-used) to specify even the ranges?

The only version constraint allowed is `*` which results in `vers:all/*`. (cf. to line 651: "``vers:all/*`` is the only valid vers form for this scheme.") So theoretically, yes that is a range but it is the only one allowed and matches any version given. Given that `vers:all/*` is the only form allowed, I would argue that it is hard to misuse.

jkowalleck (approved):

pombredanne (commented):
Thanks. Merging!

PR #432, opened 2025-03-26
Clarification regarding the version in the oci type

p-rog:
Based on the https://github.com/package-url/purl-spec# the `version` is an optional filed:  
`version: the version of the package. Optional.`
When on the `oci` purl type description there is information that it's required parameter, which can cause a confusion. 

Version should stay as optional field because it covers specific use case when the version of the component is unknown.


mprpic (approved):
LGTM!

BradLugo:
Looks like there was a discussion about this exact wording when it was proposed: https://github.com/package-url/purl-spec/pull/123#discussion_r726177852

pombredanne (approved):
Thanks. Good catch! :+1:

Issue #381, opened 2025-01-22
Clarify spec for namespace and name


pombredanne:
We are treating namespace and name at once, as they could be merged and tightly related.

johnmhoran:
Beginning to flesh out the relevant issues, PRs and comments.  These include:

Issues

- https://github.com/package-url/purl-spec/issues/35
- https://github.com/package-url/purl-spec/issues/204

PRs

- https://github.com/package-url/purl-spec/pull/176
- https://github.com/package-url/purl-spec/pull/317
- https://github.com/package-url/purl-spec/pull/367
- https://github.com/package-url/purl-spec/pull/373

See also from package-url/packageurl-python:

- https://github.com/package-url/packageurl-python/pull/178

johnmhoran:
One more related item from packageurl-python (considered but otherwise not addressed):

- [x] https://github.com/package-url/packageurl-python/issues/131

PR #453, opened 2025-04-09
Clarify namespace and name components

johnmhoran:
- This initial commit begins to address https://github.com/package-url/purl-spec/issues/381, starting with a few [RFC 2119](https://datatracker.ietf.org/doc/html/rfc2119) updates.  (Turning next to a review of the other issues and PRs listed in issue 381.)

### DIFF
--- a/PURL-SPECIFICATION.rst
+++ b/PURL-SPECIFICATION.rst
@@ -143,7 +143,7 @@ The rules for each component are:
   - Unless required by the package's ``type`` definition in `<PURL-TYPES.rst>`_,
     the ``namespace`` is optional.
   - If present, the ``namespace`` MAY contain one or more segments, separated
-    by a single slash '/' character.
+    by a single unencoded slash '/' character.
   - All leading and trailing slashes '/' are not significant and SHOULD be
     stripped in the canonical form. They are not part of the ``namespace``.
   - Each ``namespace`` segment MUST be a percent-encoded string.
--- a/PURL-SPECIFICATION.rst
+++ b/PURL-SPECIFICATION.rst
@@ -140,8 +140,10 @@ The rules for each component are:
 
 - **namespace**:
 
-  - The optional ``namespace`` contains zero or more segments, separated by a
-    single slash '/' character.
+  - Unless required by the package's ``type`` definition in `<PURL-TYPES.rst>`_,
+    the ``namespace`` is optional.
+  - If present, the ``namespace`` MAY contain one or more segments, separated
+    by a single slash '/' character.
   - All leading and trailing slashes '/' are not significant and SHOULD be
     stripped in the canonical form. They are not part of the ``namespace``.
   - Each ``namespace`` segment MUST be a percent-encoded string.
@@ -149,7 +151,8 @@ The rules for each component are:
 
     - MUST NOT contain any slash '/' characters
     - MUST NOT be empty
-    - MAY contain any ASCII character other than '/'
+    - MAY contain any ASCII character other than '/' unless the package's
+      ``type`` definition provides otherwise.
 
   - A URL host or Authority MUST NOT be used as a ``namespace``. Use instead a
     ``repository_url`` qualifier. Note however that for some types, the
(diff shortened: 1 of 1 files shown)


jkowalleck (dismissed):

johnmhoran:
(Please keep this open -- more to come.  :-)

johnmhoran:
I've updated the `namespace` and `name` sections and related bits in the parsing section.  Among the proposed changes are clarifications re permitted and prohibited characters, which are included primarily to prompt feedback and determine whether such language would be beneficial.  In `namespace`, for example, the proposed language is 

```
When percent-decoded, a segment: . . . MAY contain any ASCII character other than '/' unless 
the package's ``type`` definition provides otherwise.
```

`name` has similar language:

```
When percent-decoded,a ``name`` MAY contain any ASCII character unless prohibited by the 
package's ``type`` definition in `<PURL-TYPES.rst>`_.
```

ppkarwasz:
@johnmhoran,

So we restrict the allowed names for segments and `name` to ASCII? There are no ecosystems that allow arbitrary Unicode letters?

matt-phylum:
There are supported ecosystems that allow Unicode characters. [`pkg:nuget/%E8%A5%BF%E9%83%A8%E6%9C%BA%E5%99%A8%E4%BA%BA%E5%90%8D%E7%A7%B0%E5%A4%8D%E7%94%A8%E7%BB%84%E4%BB%B6`](https://www.nuget.org/packages/%E8%A5%BF%E9%83%A8%E6%9C%BA%E5%99%A8%E4%BA%BA%E5%90%8D%E7%A7%B0%E5%A4%8D%E7%94%A8%E7%BB%84%E4%BB%B6)

PURL-SPECIFICATION.rst
```diff
@@ -405,13 +413,17 @@ To parse a ``purl`` string in its components:
 - Split the ``remainder`` once from right on '/'
 
   - The left side is the ``remainder``
+  - Strip all leading characters (e.g., '/', '//' and so on)
+    from the right side
```
ppkarwasz, line 417: If we split once from right, the right side will not have any `/` characters to strip.

PURL-SPECIFICATION.rst
```diff
@@ -405,13 +413,17 @@ To parse a ``purl`` string in its components:
...
   - Percent-decode the right side. This is the ``name``
   - UTF-8-decode this ``name`` if needed in your programming language
   - Apply type-specific normalization to the ``name`` if needed
   - This is the ``name``
 
 - Split the ``remainder`` on '/'
 
+  - Strip all leading '/' characters (e.g., '/', '//' and so on)
+    from that split
   - Discard any empty segment from that split
```
ppkarwasz, line 427: If we strip on `/`, the segments will not contain any (unencoded) slashes.

ppkarwasz (changes requested):
@johnmhoran,

My point was that after percent-decoding what we obtain is **Unicode** characters not ASCII characters.

johnmhoran:
Thank you @ppkarwasz .  

Re Unicode vs ASCII, an earlier (not pushed) PR I'd drafted referred to Unicode rather than ASCII but I changed it to ASCII fearing my suggestion was far too broad.  From @matt-phylum 's comment it sounds like he'd propose removing the changes completely, but we'll see what he (and other community members) have to say.

Re the parsing, I was trying to address the extra '/' between the `name` and the `version` separator '@' in a [2023 packageurl-python issue](https://github.com/package-url/packageurl-python/issues/131) @matt-phylum submitted.  In trying to walk step-by-step through the parsing instructions for name and namespace I thought I had encountered a step at which a split was intended to result in 2 strings, left and right, but because of the errant '/' there was nothing at all on the right.  I might well have mis-applied the parsing steps somewhere along the way.

ppkarwasz:
> Re the parsing, I was trying to address the extra '/' between the `name` and the `version` separator '@' in a [2023 packageurl-python issue](https://github.com/package-url/packageurl-python/issues/131) @matt-phylum submitted. In trying to walk step-by-step through the parsing instructions for name and namespace I thought I had encountered a step at which a split was intended to result in 2 strings, left and right, but because of the errant '/' there was nothing at all on the right. I might well have mis-applied the parsing steps somewhere along the way.

Ah, you were thinking about something like `pkg:type/namespace/name/@1.0.0`? In that case trailing slashes should be removed **before** `type/namespace/name/` is split once from the right to obtain `name`.

johnmhoran:
@ppkarwasz Exactly.  :-)  Thank you!  When I have the chance I'll revisit the parsing steps for both `name` and `namespace` (and eventually the other components) so that I can better understand (and follow) the process.  :+1:

PURL-SPECIFICATION.rst
pombredanne: But splitting once could be understood as splitting on one slash

pombredanne: Eventually this will be moved to a parse/unparse section.

PURL-SPECIFICATION.rst
pombredanne: correct, but no harm done. We will simplify this when updating the parsing section

pombredanne (approved):
All clear to me. Let's merge this,

ppkarwasz (approved):

mprpic (approved):

stevespringett (approved):

johnmhoran:
I'm happy to note that this is ready to be merged -- merging.  ;-)

PR #456, opened 2025-04-10
Subpath: simplify parsing

ppkarwasz:
This PR contains the part of #452 that is specific to `subpath`:

- Moves the check for `.` and `..` segments after UTF-8 decoding. This ensures that overlong encoding of `.` (like `%C0%AE`, `%E0%80%AE`, `%F0%80%80%AE`) are also discarded.
- Requires the parser to throw an error if an (percent-encoded) solidus `/` is encountered in any path segment.

----

* Closes the parser part of #448
* part of #379

### DIFF
--- a/docs/how-to-parse.md
+++ b/docs/how-to-parse.md
@@ -11,12 +11,11 @@ To parse a `purl` string in its components:
 - Split the `purl` string once from right on '#'
 
   - The left side is the `remainder`
-  - Strip the right side from leading and trailing '/'
-  - Split this on '/'
-  - Discard any empty string segment from that split
+  - Split the right side on `/`
   - Percent-decode each segment
-  - Discard any '.' or '..' segment from that split
   - UTF-8-decode each segment if needed in your programming language
+  - Discard any segment that is empty, or equal to `.` or `..`
+  - Report an error if any segment contains a slash `/`
   - Join segments back with a '/'
   - This is the `subpath`
 


jkowalleck (dismissed):

johnmhoran:
Why are we deleting ` Strip the right side from leading and trailing '/'` ?  The revised language does not seem to do anything with them, so they would remain -- and the subpath section still directs ` Leading and trailing slashes '/' are not significant and SHOULD be stripped in the canonical form`.  What happens to leading/trailing slashes if we do not strip them?  ;-)

Please replace "solidus" with "slash", which is already used throughout the core spec; "solidus" is not used anywhere there, or in RFC 3986.

ppkarwasz:
> Why are we deleting ` Strip the right side from leading and trailing '/'` ? The revised language does not seem to do anything with them, so they would remain -- and the subpath section still directs ` Leading and trailing slashes '/' are not significant and SHOULD be stripped in the canonical form`. What happens to leading/trailing slashes if we do not strip them? ;-)

Leading and trailing slashes result in empty segments that are later removed.

ppkarwasz:
> Please replace "solidus" with "slash", which is already used throughout the core spec; "solidus" is not used anywhere there, or in RFC 3986.

Done in https://github.com/package-url/purl-spec/pull/456/commits/b55100e20de2c3f7cabd8737573de7ba3729e048

johnmhoran (approved):
Thank you @ppkarwasz -- LGTM.  :+1:

jkowalleck (approved):

ppkarwasz:
@johnmhoran,

I rebased this on the current structure.

PR #461, opened 2025-04-22
Update "Character encoding" and related provisions #438

johnmhoran:
Reference: https://github.com/package-url/purl-spec/issues/438

### DIFF
--- a/PURL-SPECIFICATION.rst
+++ b/PURL-SPECIFICATION.rst
@@ -227,7 +227,7 @@ Character encoding
 Permitted characters
 --------------------
 
-A canonical ``purl`` is composed of these characters ("Permitted Characters"):
+A canonical ``purl`` is composed of these Permitted Characters:
 
 - alphanumeric characters ``A to Z``, ``a to z``, ``0 to 9``,
 - the ``purl`` separators ``:/@?=&#`` (colon ':', slash '/', at sign '@',
@@ -257,31 +257,26 @@ These ``purl`` separator characters MUST NOT be percent-encoded when used as
 Percent-encoding rules
 ----------------------
 
-Unless otherwise provided in this specification, when applying percent-encoding
-or decoding to a string, use the rules of RFC 3986 section 2
-(https://datatracker.ietf.org/doc/html/rfc3986#section-2).  In the event of any
-conflict between this specification and RFC 3986 section 2, this specification
-governs.
+- In the "Rules for each ``purl`` component" section above, each component
+  defines when and how to apply percent-encoding and decoding to its content,
+  including which characters to percent-encode and when percent-encoding is
+  required.
+- When percent-encoding is required by a component definition, each
+  codepoint MUST be replaced by the percent-encoded bytes of the codepoint's
+  UTF-8 encoding using the percent-encoding mechanism defined in RFC 3986
+  section 2.1 (https://datatracker.ietf.org/doc/html/rfc3986#section-2.1).
+- With the exception of the percent-encoding mechanism, the rules regarding
+  percent-encoding are defined by this specification alone.
+- Where the space ' ' is permitted, it MUST be percent-encoded as
+  '%20'.
+- The following characters do not need to be percent-encoded:
 
-In the "Rules for each ``purl`` component" section above, each component
-defines when and how to apply percent-encoding and decoding to its content.
-
-When percent-encoding is required, all Permitted Characters MUST be encoded as
-UTF-8 and then percent-encoded except for the following:
-
-- the alphanumeric characters,
-
-- the ASCII characters ``.-_~`` (period '.', dash '-', underscore
-  '_' and tilde '~'),
-
-- the percent sign '%' when used to represent a percent-encoded character,
-
-- a ``purl`` separator when being used as a ``purl`` separator, and
-
-- the colon ':', whether used as a ``purl`` separator or otherwise.
-
-In addition, where the space ' ' is permitted, it MUST be percent-encoded as
-'%20'.
+  - the alphanumeric characters ``A to Z``, ``a to z``, ``0 to 9``,
+  - the ASCII characters ``.-_~`` (period '.', dash '-', underscore
+    '_' and tilde '~'),
+  - the percent sign '%' when used to represent a percent-encoded character,
+  - a ``purl`` separator when being used as a ``purl`` separator, and
+  - the colon ':', whether used as a ``purl`` separator or otherwise.
 
 
 How to build ``purl`` string from its components
--- a/PURL-SPECIFICATION.rst
+++ b/PURL-SPECIFICATION.rst
@@ -114,9 +114,11 @@ Rules for each ``purl`` component
 
 A ``purl`` string is an ASCII URL string composed of seven components.
 
-Some components are allowed to use other characters beyond ASCII: these
-components must then be UTF-8-encoded strings and percent-encoded as defined in
-the "Character encoding" section.
+Except as expressly stated otherwise in this section, each component:
+
+- MAY be composed of any of the characters defined as "Permitted Characters" in
+  the "Character encoding" section
+- MUST be encoded as defined in the "Character encoding" section
 
 The rules for each component are:
 
@@ -225,17 +227,13 @@ Character encoding
 Permitted characters
 --------------------
 
-A canonical ``purl`` is an ASCII string composed of these characters:
+A canonical ``purl`` is composed of these characters ("Permitted Characters"):
 
 - alphanumeric characters ``A to Z``, ``a to z``, ``0 to 9``,
 - the ``purl`` separators ``:/@?=&#`` (colon ':', slash '/', at sign '@',
   question mark '?', equal sign '=', ampersand '&' and pound sign '#'), and
-- these punctuation marks ``%.-_~`` (percent sign '%', period '.', dash '-',
-  underscore '_' and tilde '~').
-
-All other characters MUST be encoded as UTF-8 and then percent-encoded.
-In addition, each component specifies its permitted characters and
-its percent-encoding rules.
+- the ASCII characters ``+%.-_~`` (plus '+', percent sign '%', period '.',
+  dash '-', underscore '_' and tilde '~').
 
 
 ``purl`` separators
@@ -259,14 +257,31 @@ These ``purl`` separator characters MUST NOT be percent-encoded when used as
 Percent-encoding rules
 ----------------------
 
-When applying percent-encoding or decoding to a string, use the rules of RFC
-3986 section 2 (https://datatracker.ietf.org/doc/html/rfc3986#section-2).
+Unless otherwise provided in this specification, when applying percent-encoding
+or decoding to a string, use the rules of RFC 3986 section 2
+(https://datatracker.ietf.org/doc/html/rfc3986#section-2).  In the event of any
+conflict between this specification and RFC 3986 section 2, this specification
+governs.
 
-Each component defines when and how to apply percent-encoding and decoding to
-its content.
+In the "Rules for each ``purl`` component" section above, each component
+defines when and how to apply percent-encoding and decoding to its content.
 
-When percent-encoding is required, all characters MUST be encoded except for
-the colon ':'.
+When percent-encoding is required, all Permitted Characters MUST be encoded as
+UTF-8 and then percent-encoded except for the following:
+
+- the alphanumeric characters,
+
+- the ASCII characters ``.-_~`` (period '.', dash '-', underscore
+  '_' and tilde '~'),
+
+- the percent sign '%' when used to represent a percent-encoded character,
+
+- a ``purl`` separator when being used as a ``purl`` separator, and
+
+- the colon ':', whether used as a ``purl`` separator or otherwise.
+
+In addition, where the space ' ' is permitted, it MUST be percent-encoded as
+'%20'.
 
 
 How to build ``purl`` string from its components
--- a/PURL-SPECIFICATION.rst
+++ b/PURL-SPECIFICATION.rst
@@ -116,8 +116,8 @@ A ``purl`` string is an ASCII URL string composed of seven components.
 
 Except as expressly stated otherwise in this section, each component:
 
-- MAY be composed of any of the characters defined as "Permitted Characters" in
-  the "Character encoding" section
+- MAY be composed of any of the characters defined in the "Permitted
+  characters" section
 - MUST be encoded as defined in the "Character encoding" section
 
 The rules for each component are:
@@ -221,26 +221,24 @@ The rules for each component are:
   - The ``subpath`` MUST be interpreted as relative to the root of the package
 
 
-Character encoding
-~~~~~~~~~~~~~~~~~~
-
 Permitted characters
---------------------
+~~~~~~~~~~~~~~~~~~~~
 
-A canonical ``purl`` is composed of these Permitted Characters:
+A canonical ``purl`` is composed of these permitted ASCII characters:
 
-- alphanumeric characters ``A to Z``, ``a to z``, ``0 to 9``,
-- the ``purl`` separators ``:/@?=&#`` (colon ':', slash '/', at sign '@',
-  question mark '?', equal sign '=', ampersand '&' and pound sign '#'), and
-- the ASCII characters ``+%.-_~`` (plus '+', percent sign '%', period '.',
-  dash '-', underscore '_' and tilde '~').
+- the Alphanumeric Characters: ``A to Z``, ``a to z``, ``0 to 9``,
+- the Punctuation Characters: ``.-_~`` (period '.',
+  dash '-', underscore '_' and tilde '~'),
+- the Plus Character: ``+`` (plus '+'),
+- the Percent Character: ``%`` (percent sign '%'), and
+- the Separator Characters ``:/@?=&#`` (colon ':', slash '/', at sign '@',
+  question mark '?', equal sign '=', ampersand '&' and pound sign '#').
 
 
 ``purl`` separators
--------------------
+~~~~~~~~~~~~~~~~~~~
 
-These ``purl`` separator characters MUST NOT be percent-encoded when used as
-``purl`` separators:
+This is how each of the Separator Characters is used:
 
 - ':' (colon) is the separator between ``scheme`` and ``type``
 - '/' (slash) is the separator between ``type``, ``namespace`` and ``name``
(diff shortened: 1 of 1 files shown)


johnmhoran:
@pombredanne @mprpic @jkowalleck @ppkarwasz @matt-phylum  (and the rest of the PURL community ;-)  This is an update to the encoding clarification made via PR #439.  I've tried to provide a few additional clear, concise statements addressing the various related issues/PRs as well as recent comments.  Please do not hesitate to correct, clarify and question as needed -- and please provide alternative language proposals whenever you can.

PURL-SPECIFICATION.rst
```diff
@@ -259,14 +257,31 @@ These ``purl`` separator characters MUST NOT be percent-encoded when used as
 Percent-encoding rules
 ----------------------
 
-When applying percent-encoding or decoding to a string, use the rules of RFC
-3986 section 2 (https://datatracker.ietf.org/doc/html/rfc3986#section-2).
+Unless otherwise provided in this specification, when applying percent-encoding
+or decoding to a string, use the rules of RFC 3986 section 2
+(https://datatracker.ietf.org/doc/html/rfc3986#section-2).  In the event of any
+conflict between this specification and RFC 3986 section 2, this specification
+governs.
```
matt-phylum, line 264: Are we going back to how it was before? The way I understood the current version was that it was a breaking change to intentionally break from RFC3986 section 2 because the RFC3986 encoding rules are more complicated to implement and most PURL implementations did not try to implement them, instead applying mostly the same encoding rules to all components. I've already had somebody ask why phylum-dev/purl doesn't encode plus signs in version numbers, which isn't required by RFC 3968 and is non-canonical according to the old PURL but is required by the rules in the current version of PURL, or at least we both thought so.

PURL-SPECIFICATION.rst
```diff
@@ -259,14 +257,31 @@ These ``purl`` separator characters MUST NOT be percent-encoded when used as
...
+
+In the "Rules for each ``purl`` component" section above, each component
+defines when and how to apply percent-encoding and decoding to its content.
+
+When percent-encoding is required, all Permitted Characters MUST be encoded as
+UTF-8 and then percent-encoded except for the following:
```
matt-phylum, line 270: This seems at odds with the first paragraph. "When percent-encoding" should be redundant because percent-encoding is always required (PURL has components that must not be encoded, but I would say it's more accurate that those components do not allow any characters that require encoding, especially for qualifier keys), but then because it says to use RFC 3986 rules and then "when percent-encoding is required, [...] characters must be [...] percent-encoded except," this could be taken to mean these are exceptions to the RFC 3986 rules instead of a replacement of the RFC 3986 rules (which don't map one-to-one with PURL).

PURL-SPECIFICATION.rst
```diff
@@ -259,14 +257,31 @@ These ``purl`` separator characters MUST NOT be percent-encoded when used as
...
+
+- the alphanumeric characters,
```
matt-phylum, line 272: This needs to specify which alphanumeric characters. Most people will probably understand that ß does not count, but it is a Unicode letter.

ppkarwasz:
I think that the main problems of the encoding sections are:

- the ambiguity between "character" used in the sense of byte/octet and "character" used to designate a Unicode codepoint.
- the ambiguity between "percent-encoding" as operation that transforms one octet into three octets and operation that encodes a sequence of Unicode characters into a sequence of octets.

Section 2 of RFC 3986 is all about encoding octets and it only defines an operation that takes one octet and represents it as three octets of the form `%[a-fA-F0-9]{2}` (using the mapping between characters and octets provided by US-ASCII).

Section 2.4 doesn't give hard rules on when this single-byte operation must be performed. It only says that:

- the octets in the _unreserved_ set SHOULD be represented as single octets (i.e. no "percent-encoding"). We want to change this to MUST, at least for the canonical form.
- each implementation should determine whether an octet in the _reserved_ set, that is not used as a delimiter, should be "percent-encoded" or can be safely copied directly to the output. Here we want a simple rule:
  - `:` can be used directly.
  - the other PURL separators MUST be "percent-encoded"
  - we didn't say anything about the remaining _reserved_ characters that are not used in PURL (e.g. `[`). IMHO these can be used directly.

johnmhoran:
As I just noted in a [comment in @mprcic's encoding discussion](https://github.com/package-url/purl-spec/discussions/459#discussioncomment-12929109), on reflection I'm leaning towards [@matt-phylum's suggestion](https://github.com/package-url/purl-spec/discussions/459#discussioncomment-12881111) that our RFC 3986 section 2 reference be limited to just section 2.1's percent-encoding process, which could reduce ambiguity and ease implementation.  Would we lose anything we can't afford (or don't want) to lose?

PURL-SPECIFICATION.rst
```diff
@@ -114,9 +114,11 @@ Rules for each ``purl`` component
 
 A ``purl`` string is an ASCII URL string composed of seven components.
 
-Some components are allowed to use other characters beyond ASCII: these
-components must then be UTF-8-encoded strings and percent-encoded as defined in
-the "Character encoding" section.
+Except as expressly stated otherwise in this section, each component:
+
+- MAY be composed of any of the characters defined as "Permitted Characters" in
+  the "Character encoding" section
```
ppkarwasz, line 120: Here we are talking about the **serialized** form of a component (e.g. a package with name `Pan%E8%AA%9E`).

I would rather use "Permitted Bytes" instead of "Permitted Characters" to stress out that we are in the codomain of the `percent-encode` function.

PURL-SPECIFICATION.rst
ppkarwasz, line 270: Here on the other hand we are talking about the **decoded** form of a component (e.g. a package named `Pan語`).

The domain of the `percent-encode` function is not restricted to the "Permitted Characters", but any Unicode character can be present (components can restrict this set).

pombredanne: @matt-phylum how would you phrase this?

PURL-SPECIFICATION.rst
```diff
@@ -225,17 +227,13 @@ Character encoding
 Permitted characters
 --------------------
 
-A canonical ``purl`` is an ASCII string composed of these characters:
+A canonical ``purl`` is composed of these characters ("Permitted Characters"):
```
pombredanne, line 230: Do we really need `("Permitted Characters")` here? What about this instead:
```suggestion
A canonical ``purl`` is composed of these Permitted Characters:
```

PURL-SPECIFICATION.rst
pombredanne: @mjherzog I think that Matts='s point may be that we re-spell out other characters below but not here. It does not hurt to list things here again IMHO, does it?
Something like:
 ```suggestion
- the alphanumeric characters ``A to Z``, ``a to z``, ``0 to 9``,
```

Matt, was this your point?

File: PURL-SPECIFICATION.rst [hunk 6]
```diff
@@ -259,14 +257,31 @@ These ``purl`` separator characters MUST NOT be percent-encoded when used as
...
+defines when and how to apply percent-encoding and decoding to its content.
+
+When percent-encoding is required, all Permitted Characters MUST be encoded as
+UTF-8 and then percent-encoded except for the following:
+
+- the alphanumeric characters,
+
+- the ASCII characters ``.-_~`` (period '.', dash '-', underscore
+  '_' and tilde '~'),
+
```
pombredanne, line 276: @johnmhoran we do not use much empty lines for single depth bulleted lists elsewhere. You may want to remove these lines in that list from:
```
- the alphanumeric characters,

- the ASCII characters ``.-_~`` (period '.', dash '-', underscore
  '_' and tilde '~'),

- the percent sign '%' when used to represent a percent-encoded character,

- a ``purl`` separator when being used as a ``purl`` separator, and

- the colon ':', whether used as a ``purl`` separator or otherwise.
```
to something like:
```
- the alphanumeric characters,
- the ASCII characters ``.-_~`` (period '.', dash '-', underscore
  '_' and tilde '~'),
- the percent sign '%' when used to represent a percent-encoded character,
- a ``purl`` separator when being used as a ``purl`` separator, and
- the colon ':', whether used as a ``purl`` separator or otherwise.
```

PURL-SPECIFICATION.rst
pombredanne: @matt-phylum can you elaborate? I am not sure I understand fully your point. I thing that what we are trying to convey is that:
1. this spec defines WHICH characters to encode and WHERE/WHEN (e.g., with specifics for separators and components)
2. we defer to RFC3986 to define HOW to encode characters we want encoded.

Would this be clearer?

pombredanne (commented):
@johnmhoran Thanks! See my comments for your consideration.

PURL-SPECIFICATION.rst
matt-phylum: It's unfortunate that the process of applying "percent-encoding" to a string does not necessarily change the string because it complicates talking about which characters need to be changed vs which characters do not need to be changed. RFC 3986 talks about percent-encoding a (byte) string, a process which may or may not alter the string, and percent-encoding an octet, a process which consistently converts one octet into three. WHATWG URL is similar, talking about percent-encoding a byte sequence, a process which may or may not alter the byte sequence, and percent-encoding a byte, a process which consistently converts one octet into three. PURL, at least in this section, is less specific about what "percent-encoding" means.

Merging in [@ppkarwasz's comment](https://github.com/package-url/purl-spec/pull/461/files#r2059140578), it could say something like this:

> When serializing a string, unless excluded by the following rules, every code point must be replaced by the percent-encoded bytes of the code point's UTF-8 encoding.

PURL-SPECIFICATION.rst
matt-phylum: By saying "In the event of any conflict between this specification and RFC 3986 section 2" it sounds to me like the implementer is supposed to combine RFC 3986 and PURL rules, eg merging RFC 3986 pchar with the PURL character rules when outputting a package name. If the PURL spec is clearly specifying WHICH characters WHERE/WHEN and RFC 3986 is specifying HOW then it's much easier to implement and there shouldn't be conflicts.

johnmhoran:
@pombredanne @matt-phylum @ppkarwasz @mjherzog  Thank you for your thoughtful comments.  I'll give them a closer read a bit later today and then draft an update (if I miss any of your points in that update, please advise and I'll correct).  Thanks as well for the language proposals -- super helpful -- keep them coming!  👍   

@jkowalleck and @mprpic (and others) -- please share your thoughts as well.

PURL-SPECIFICATION.rst
johnmhoran: @pombredanne Done.  Are you suggesting we keep using "Characters" in this title rather than "Bytes" as @ppkarwasz suggested?

johnmhoran:
I've just pushed an update that addresses many of the points and comments so far, though not all, which still need to be agreed upon.  That includes

- referring to serialization when defining the percent-encoding process -- this latest update takes a different approach, but perhaps serialization is better -- whatever we agree on works for me
- the proposed change from Permitted Characters to Permitted Bytes and related changes, including questions about the way we currently use the term "character"

PURL-SPECIFICATION.rst
```diff
@@ -257,31 +257,26 @@ These ``purl`` separator characters MUST NOT be percent-encoded when used as
...
-
-- the colon ':', whether used as a ``purl`` separator or otherwise.
-
-In addition, where the space ' ' is permitted, it MUST be percent-encoded as
-'%20'.
+- In the "Rules for each ``purl`` component" section above, each component
+  defines when and how to apply percent-encoding and decoding to its content,
+  including which characters to percent-encode and when percent-encoding is
+  required.
+- When percent-encoding is required by a component definition, each
```
ppkarwasz, line 264: For simplicity's sake I would say that percent encoding is always required. If a component only allows characters from the unreserved set, then percent-encoding simply becomes the identity function.

PURL-SPECIFICATION.rst
```diff
@@ -257,31 +257,26 @@ These ``purl`` separator characters MUST NOT be percent-encoded when used as
...
+  codepoint MUST be replaced by the percent-encoded bytes of the codepoint's
+  UTF-8 encoding using the percent-encoding mechanism defined in RFC 3986
+  section 2.1 (https://datatracker.ietf.org/doc/html/rfc3986#section-2.1).
+- With the exception of the percent-encoding mechanism, the rules regarding
+  percent-encoding are defined by this specification alone.
+- Where the space ' ' is permitted, it MUST be percent-encoded as
```
ppkarwasz, line 270: The rule about ` ` can be simply deduced from the other rules.

PURL-SPECIFICATION.rst
```diff
@@ -257,31 +257,26 @@ These ``purl`` separator characters MUST NOT be percent-encoded when used as
...
+  '%20'.
+- The following characters do not need to be percent-encoded:
```
ppkarwasz, line 272: Should we say "MUST NOT"?

PURL-SPECIFICATION.rst
```diff
@@ -257,31 +257,26 @@ These ``purl`` separator characters MUST NOT be percent-encoded when used as
...
+
+  - the alphanumeric characters ``A to Z``, ``a to z``, ``0 to 9``,
+  - the ASCII characters ``.-_~`` (period '.', dash '-', underscore
+    '_' and tilde '~'),
+  - the percent sign '%' when used to represent a percent-encoded character,
+  - a ``purl`` separator when being used as a ``purl`` separator, and
+  - the colon ':', whether used as a ``purl`` separator or otherwise.
```
ppkarwasz, line 279: Since each component specifies the arguments of the `percent-encode` method, I think this is not necessary.
The argument of `percent-encode` will never contain "characters used as `purl` separators"; those characters will be added afterwards.

For example:

- percent-encode each **segment** of the namespace.
- join the encoded segments with the `/` character.

ppkarwasz (commented):
Looks very nice! :100:

My 2 cents.

PURL-SPECIFICATION.rst
johnmhoran: Thanks for this comment @ppkarwasz .  Not sure if you're addressing all 3 lines you excerpt or just line 278, so just in case:

- line 277: that is definitely more than needed -- was just trying to be thorough ;-)  no objections at all to deleting if others agree
- line 278: believe it or not, there have been issues that this point addresses, e.g., does the colon between scheme and type need to be percent-encoded?  See, e.g., [Percent encoding spec and : and /](https://github.com/package-url/purl-spec/issues/39); imho this is needed to avoid such issues in the future and make the use of a PURL as clear as possible.
- line 279: perhaps I am misunderstanding your point here -- without line 279, how will users know that colons do not need to be percent-encoded?

PURL-SPECIFICATION.rst
johnmhoran: Thanks @ppkarwasz .  I have seen that point asserted elsewhere (that percent-encoding is always required) but that's contrary to my understanding that percent-encoding is only required under certain defined circumstances (both in RFC 3986 and in the PURL spec).  In the PURL spec, it's the component definitions that instruct WHICH characters and WHEN, and the `type` definition, for example, makes clear that `The type MUST NOT be percent-encoded.`

ppkarwasz: There are two concepts of "percent-encoding": the [percent encoding of a byte](https://url.spec.whatwg.org/#percent-encode) and the [percent encoding after encoding of a string](https://url.spec.whatwg.org/#string-percent-encode-after-encoding).
Some components like `type` will never need to use the former, but they can certainly used the latter. The "percent encoding after encoding" of `maven` will be simply `maven`.

PURL-SPECIFICATION.rst
ppkarwasz: > - line 278: believe it or not, there have been issues that this point addresses, e.g., does the colon between scheme and type need to be percent-encoded?

Line 242 says:

> These ``purl`` separator characters MUST NOT be percent-encoded when used as ``purl`` separators"

Do we need to repeat it here too?

> * line 279: perhaps I am misunderstanding your point here -- without line 279, how will users know that colons do not need to be percent-encoded?

Sure we need to say that colon `:` does not need to be percent-encoded, but I think we don't need to repeat that it also does not need to be encoded when used as a separator.

Maybe we could make this paragraph less descriptive and more imperative like:

```
To percent-encode a string of characters:

1. encode it using UTF-8,
2. for each byte of the encoded string:
    - if the byte corresponds to:
       - an alphanumeric ASCII character (``A to Z``, ``a to z``, ``0 to 9``)
       - or one of the ASCII characters `.`, `-`, `_`, `~` and `:`.
       
      copy the byte to the output.
    - otherwise, append the percent-encoding of the byte to the output, as defined in RFC 3986
      section 2.1 (https://datatracker.ietf.org/doc/html/rfc3986#section-2.1).
```

PURL-SPECIFICATION.rst
pombredanne: @ppkarwasz Thanks, but let's not introduce bytes just yet.

PURL-SPECIFICATION.rst
pombredanne: @ppkarwasz let's refine the wording alright. But we use RFC 3986 encoding rules for each "octet" and we define which character needs this treatment and when. Blanket encoding everything will be a source of confusion IMHO.

PURL-SPECIFICATION.rst
pombredanne: Agreed, but let's be explicit rather than implicit

PURL-SPECIFICATION.rst
pombredanne: agreed

PURL-SPECIFICATION.rst
pombredanne: Let's reference this as permitted ASCII characters.

johnmhoran:
@pombredanne I've committed and pushed the latest set of changes to the character-encoding section.  Looking forward to comments and suggestions.

ppkarwasz (approved):
Looks good to me!

mprpic:
LGTM as well!

pombredanne (approved):
Looking all fine now. Let's merge this puppy!

pombredanne:
@mprpic can you approve through the formal GH mechanism?

pombredanne:
Actually, nm, I consider this as approval!

> LGTM as well!

_Originally posted by @mprpic in https://github.com/package-url/purl-spec/issues/461#issuecomment-2877707132_

PR #474, opened 2025-05-30
Add libversion as generic versioning scheme

immqu:
This PR adds libversion to the list of generic versioning schemes. The comparison algorithm of libversion is described in https://github.com/repology/libversion/blob/3.0.3/doc/ALGORITHM.md and I have linked it in the version range spec (I have pinned it to the latest release).

Addresses package-url/vers-spec#15

### DIFF
--- a/VERSION-RANGE-SPEC.rst
+++ b/VERSION-RANGE-SPEC.rst
@@ -660,6 +660,8 @@ These are generic schemes, to use sparingly for special cases:
 
 - **semver**: a generic scheme that uses the same syntax as ``semver``. It follows the MAJOR.MINOR.PATCH format and is defined in the Semantic Versioning Specification 2.0.0, see https://semver.org/spec/v2.0.0.html. 
 
+- **libversion**: a generic versioning scheme that is designed to compare versions of different notations. It splits all numeric and alphabetic components and then performs the comparison per component. The algorithm is described in https://github.com/repology/libversion/blob/3.0.3/doc/ALGORITHM.md.
+
 A separate document will provide details for each versioning scheme and:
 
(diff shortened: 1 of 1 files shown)


VERSION-RANGE-SPEC.rst
```diff
@@ -660,6 +660,8 @@ These are generic schemes, to use sparingly for special cases:
 
 - **semver**: a generic scheme that uses the same syntax as ``semver``. It follows the MAJOR.MINOR.PATCH format and is defined in the Semantic Versioning Specification 2.0.0, see https://semver.org/spec/v2.0.0.html. 
 
+- **libversion**: a generic versioning scheme that is designed to compare versions of different notations. It splits all numeric and alphabetic components and then performs the comparison per component. The algorithm is described in https://github.com/repology/libversion/blob/3.0.3/doc/ALGORITHM.md.
```
pombredanne, line 663: It makes sense... yet at the same time, libversion is trying to provide a single algorithm to parse any version... I wonder if this would not be instead something to use in univers or a vers implementation as a last resort, fallback?

immqu: Good point. I think we can see libversion both as a comparison algorithm and a schema: It allows versions like ```1.0beta1``` -- which is a greater version than ```1.0alpha1```. So, implicitly it combines intdot and lexicographic.
I like the idea of adding the algorithm as a fallback in univers, though! I will look into this.

pombredanne:
Closing in favor of revisiting the topic as a possible fallback. See details in:
- https://github.com/package-url/vers-spec/issues/15

pombredanne:
@immqu FWIW, PRs cannot be moved across GitHub orgs.

PR #477, opened 2025-06-04
TYPES: Add `gitee` type

PrinterFranklin:
# Motivation

[Gitee](https://gitee.com/) is the biggest code hosting platform in Asia with many original projects. We hope Gitee can be accepted as a type so these projects can be tracked with PURL.

# Changes

- Add a first draft for the description of the gitee type


pombredanne:
After the merge of PR #514, PURL types and tests are now defined in JSON: :innocent: :grin: 
- See  #514

With the new approach... this PR needs to be updated. Could you look into this? Thanks for your understanding

PR #478, opened 2025-06-04
Add new PURL type: 'powershell'

alenmRL:
### Motivation
Add support for PowerShell modules distributed via the [PowerShell Gallery](https://www.powershellgallery.com/).
PowerShell is a widely used automation and configuration framework developed by Microsoft. Its module ecosystem is centralized in the PowerShell Gallery, which currently hosts over 450,000 packages, making it a significant and active package registry in the software ecosystem.

### DIFF
--- a/purl-types-index.json
+++ b/purl-types-index.json
@@ -25,6 +25,7 @@
   "npm",
   "nuget",
   "oci",
+  "psgallery",
   "pub",
   "pypi",
   "qpkg",
--- a/tests/types/powershell-test.json
+++ b/tests/types/powershell-test.json
@@ -1,0 +1,87 @@
+{
+  "$schema": "https://packageurl.org/schemas/purl-test.schema-1.0.json",
+  "tests": [
+    {
+      "description": "Powershell names are case insensitive. Roundtrip test for Powershell with version.",
+      "test_group": "base",
+      "test_type": "roundtrip",
+      "input": "pkg:Psgallery/EnterpriseLibrary.Common@6.0.1304",
+      "expected_output": "pkg:psgallery/enterpriselibrary.common@6.0.1304",
+      "expected_failure": false,
+      "expected_failure_reason": null
+    },
+    {
+      "description": "Powershell names are case insensitive. Parse test with version.",
+      "test_group": "base",
+      "test_type": "parse",
+      "input": "pkg:psgallery/Enterpriselibrary.Common@6.0.1304",
+      "expected_output": {
+        "type": "psgallery",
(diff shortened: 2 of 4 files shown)


pombredanne:
Thanks! Are these all nupkg?

alenmRL:
Yes, they’re all `.nupkg` files.

rhalar:
I work with @alenmRL and just wanted to jump in with an observation

Although it isn't explicitly stated in the specification NuGet (and PowerShell which is based on the same package format) treat package names as case-insensitive. This is the reason for the last change, and we'd advise the NuGet specification also reflect this and be updated, to allow exact unique identifiers to be constructed via PURLs.

Additionally, the NuGet specification also states
>NuGetVersion uses case insensitive string comparisons for pre-release components. This means that 1.0.0-alpha and 1.0.0-Alpha are equal

so perhaps enforcing a lowercase version string would also be beneficial, in lieu of full version normalization which technically isn't required in the PURL spec.
Though that is also something to consider, see for example
`pkg:psgallery/pswindowsupdate@2.2.1.0`
`pkg:psgallery/pswindowsupdate@2.2.1`

which are equivalent, but we can get 2 (or more) different PURLs in such cases.

pombredanne:
- See  #514

With the new approach... this PR needs to be updated. Could you look into this? Thanks for your understanding

ppkarwasz:
Is there an essential difference between `nuget.org` and `powershellgallery.com` to justify a different PackageURL type?

alenmRL:
> Is there an essential difference between `nuget.org` and `powershellgallery.com` to justify a different PackageURL type?

NuGet.org and PowerShellGallery.com serve different ecosystems even though they both use the NuGet package format. While both use .nupkg packages, there is also a difference in package manager: NuGet packages are installed into .NET projects using nuget or dotnet commands, whereas PowerShell modules are installed into the PowerShell environment using Install-Module.
The file format alone shouldn’t determine the purl type; what matters is the ecosystem and how the package is consumed. Even though both NuGet and PowerShellGallery packages use .nupkg archives, they belong to different ecosystems with distinct metadata, conventions, and tooling, so they deserve separate purl types.

A similar discussion took place in this [thread](https://github.com/package-url/purl-spec/pull/673#discussion_r2406165474), where it was decided to introduce two separate PURL types.

An important detail is that PowerShell packages include their own module manifest (.psd1), which defines metadata, exported commands, and dependencies.

rhalar:
@pombredanne I think everything was addressed, could we push this further?

Issue #380, opened 2025-01-22
Clarify spec for version


pombredanne:

pombredanne:
@mprpic This is yours since your volunteered. Do you want to join the project?

mprpic:
@pombredanne Yes please.

davidB:
Yes please clarify, because the doc says version is percent-encoded. But it's only for some characters => it's a pain, because if I always percent-encode the version before building the URL:

`sha256:adf450ad2a44e7cf94a9ba15378ad66f3b906ebd88d541f980d3b3e2b33f2399` => `sha256%3Aadf450ad2a44e7cf94a9ba15378ad66f3b906ebd88d541f980d3b3e2b33f2399` as expected

`0.3.0` => `0%2E3%2E0`  (ok for `tag` in query parameter by against version samples with `...@0.3.0`

matt-phylum:
The encoding is only supposed to be applied to characters that require encoding to be unambiguous in the position they are used in: https://github.com/package-url/purl-spec/blob/master/PURL-SPECIFICATION.rst#character-encoding It's suggested that plus signs are also encoded to avoid interoperability problems: https://github.com/package-url/purl-spec/pull/261

Unfortunately, there is a lot of confusion about percent encoding and x-www-form-urlencoded and RFC3986 reserved characters, all of which are distinct topics. Percent encoding is the mechanism of encoding, and does not specify which characters are to be encoded, although obviously you have to encode % itself or its meaning will change during decoding. Based on the results you're seeing, you're probably encoding all characters except ascii alphanumeric characters, which is valid and should be understood by any PURL implementation, but it's not the canonical form.

I think at some other time on another issue I suggested having explicit rules that clearly specify which characters are or are not expected to be encoded. WHATWG URL has a nice format where it defines different encode sets and then specifies the encode set to be used for the different parts of the URL: https://url.spec.whatwg.org/#percent-encoded-bytes

As far as I know, these sets are correct for PURL: https://github.com/phylum-dev/purl/blob/151168733f75a9802556e4b07eb577b9d99f7cea/purl/src/format.rs#L9-L27

matt-phylum:
The PURL `pkg:generic/a@b@c` does have a well defined expected parse result, but it may not be what you expect. `pkg:npm/@angular/whatever@1.0.0` and `pkg:npm/@angular/whatever` differ by more than their versions.

`java.net.URI` would not encode it because it only has a special meaning to PURL.

[...]

PR #479, opened 2025-06-05
Clarify version definition #380

johnmhoran:
This PR updates the `version` definition and adds an entry to the FAQs document.

Reference: https://github.com/package-url/purl-spec/issues/380

### DIFF
--- a/PURL-SPECIFICATION.rst
+++ b/PURL-SPECIFICATION.rst
@@ -166,14 +166,12 @@ The rules for each component are:
 
 - **version**:
 
-  - The ``version`` is prefixed by a '@' separator when not empty
-  - This '@' is not part of the ``version``
-  - A ``version`` must be a percent-encoded string
-
-  - A ``version`` is a plain and opaque string. Some package ``types`` use versioning
-    conventions such as SemVer for NPMs or NEVRA conventions for RPMS. A ``type``
-    may define a procedure to compare and sort versions, but there is no
-    reliable and uniform way to do such comparison consistently.
+  - The ``version`` is prefixed by a '@' separator when not empty.
+  - This '@' is not part of the ``version``.
+  - A ``version`` MUST be a percent-encoded string.
+  - When percent-decoded, a ``version`` MAY contain any Unicode character unless
+    the package's ``type`` definition provides otherwise.
+  - A ``version`` is a plain and opaque string.
 
 
 - **qualifiers**:
--- a/faq.rst
+++ b/faq.rst
@@ -46,3 +46,15 @@ package ``type``
 
 As a result, a purl spec implementation must return an error when encountering
 a ``type`` that contains a prohibited character.
+
+
+Version
+~~~~~~~
+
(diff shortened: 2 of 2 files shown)


johnmhoran:
@pombredanne I've taken the language from the `version` definition you pointed out yesterday and moved it to a new entry in the FAQs document.  That leaves behind the following bullet in the `version` definition:

```
- A ``version`` is a plain and opaque string.
```

I'm not sure we need it or what it actually means.  What do you think?

PURL-SPECIFICATION.rst
```diff
@@ -166,14 +166,12 @@ The rules for each component are:
...
-  - This '@' is not part of the ``version``
-  - A ``version`` must be a percent-encoded string
-
-  - A ``version`` is a plain and opaque string. Some package ``types`` use versioning
-    conventions such as SemVer for NPMs or NEVRA conventions for RPMS. A ``type``
-    may define a procedure to compare and sort versions, but there is no
-    reliable and uniform way to do such comparison consistently.
+  - The ``version`` is prefixed by a '@' separator when not empty.
+  - This '@' is not part of the ``version``.
+  - A ``version`` MUST be a percent-encoded string.
```
mprpic, line 171: Should this refer to the percent-encoding section for "how" to encode it?

johnmhoran: Hi @mprpic and thanks for your comment.  My thinking was that the language I added to the top of the `Rules for each 'purl' component` section (see below) would be sufficient to address the character-encoding process, including the "how".  What do you think?

```
Except as expressly stated otherwise in this section, each component:

- MAY be composed of any of the characters defined in the "Permitted
  characters" section
- MUST be encoded as defined in the "Character encoding" section
```

mprpic: Ah, right on! Looks good then! :-)

pombredanne:
> I've taken the language from the `version` definition you pointed out yesterday and moved it to a new entry in the FAQs document. That leaves behind the following bullet in the `version` definition:
> 
> ```
> - A ``version`` is a plain and opaque string.
> ```

This means that it carries no special implied meaning. In particular the version alone does not say how to compare two versions e.g., which ones come before another among two versions.

faq.rst
```diff
@@ -46,3 +46,15 @@ package ``type``
...
+
+Version
+~~~~~~~
+
+**QUESTION**: How do package ``types`` handle the comparison and sorting of
+versions?
+
+**ANSWER**: Some package ``types`` use versioning conventions such as SemVer
+for NPMs or NEVRA conventions for RPMS. A ``type`` may define a procedure to
+compare and sort versions, but there is no reliable and uniform way to do such
```
pombredanne, line 59: actually there is a way, it is defined in the VERS companion spec. We can add this later in all cases.

pombredanne (approved):
Good to go for me :heart:

faq.rst
johnmhoran: Thank you @pombredanne .

PR #490, opened 2025-06-24
Add datetime as generic versioning scheme

immqu:
In https://github.com/package-url/vers-spec/issues/14, the usage and options for specifying a datetime scheme were discussed.
[RFC3339, Section 5.6](www.rfc-editor.org/rfc/rfc3339#section-5.6) emerged as a suitable candidate, so this PR adds the datetime scheme based on the RFC specification.


pombredanne (approved):
Thanks!

PR #496, opened 2025-06-27
Add new PURL type: 'bsd'

s-kenji:
This PR adds support for BSD package types to the purl specification. 

BSD operating system variants (e.g. FreeBSD, OpenBSD, and NetBSD) have their own package management systems, but currently lack dedicated purl types despite their widespread use.

Key points:
- FreeBSD uses `pkg` for binary packages and `ports` for source-based packages.
- OpenBSD uses `pkg_add` for package management.
- NetBSD uses `pkgsrc` for package management.


PURL-TYPES.rst
```diff
@@ -89,6 +89,28 @@ bitnami
...
+  It is not case sensitive and must be lowercased.
+- The ``name`` is the package name.
+- The ``version`` is the package version.
+- Optional qualifiers may include:
+  - ``distro``: The BSD release version or distribution variant.
+  - ``epoch``: The package epoch number used for version comparison when normal version ordering is insufficient.
+  - ``arch``: The target architecture for the package.
+
+- Examples::
+      pkg:bsd/freebsd/emacs@30.1_2?distro=14.3&epoch=3
```
kaiorafael: For FreeBSD, another suggestion would include`arch=all` so it could have similar format as any other OS that includes  qualifier `arch`.

Example: `pkg:bsd/freebsd/emacs@30.1_2?distro=14.3&epoch=3&arch=all`

If that is not required, no as it would reduce file size as well.

pombredanne (commented):
Thanks! Do you know if you can invite folks in each of these communities to chime in on this? That would be awesome!

PURL-TYPES.rst
illuusio: Would there be more name space like `bsd/freebsd/ports/emacs@30.1_2?distro=14.3&epoch=3` as it would nice to be able to also say `bsd/freebsd/kernel/module@14.3` like thing or  `bsd/freebsd/core/app@14.3` or can they be handled other way?

kaiorafael: Probably `ports` would fit better as  additional qualifier, example:

`bsd/freebsd/emacs@30.1_2?distro=14.3&epoch=3&source=ports`

More educational question: 
For Kernel module, that is interesting point, how do you collect that? Since users can compile they own Kernel and possible enable/disable modules. 

I ask this as in FreeBSD one can query the sqlite3 database and packages installed can be found, but for Kernel modules I am not too sure :D

illuusio: If [pkgbase](https://wiki.freebsd.org/PkgBase) ever goes thru (and it seems to steadily do so) it'll bring kernel modules as `pkg(8)` packages [New packages repositories for kernel modules](https://lists.freebsd.org/archives/freebsd-announce/2025-May/000189.html). If one compiles own ones there is nothing to do but official ones should be able to point with purl?

illuusio: Does it need epoch? As there is this `_2` part in package version?

pombredanne:
@s-kenji After the merge of PR #514, PURL types are now defined in JSON: :innocent: :grin: 
- See  #514

With the new approach... this PR needs to be updated. Could you look into this? Thanks for your understanding !

PURL-TYPES.rst
s-kenji: `_2` is `PORTREVISION`, which is part of version and different from epoch. We use `PORTEPOCH` as epoch qualifier. 

Please refer to [FreeBSD Porter's Handbook](https://docs.freebsd.org/en/books/porters-handbook/book/#makefile-naming-revepoch) for more information about `PORTREVISION` and `PORTEPOCH`.

s-kenji:
@pombredanne  Updated the PR with the new approach. Thank you for letting me know!

We will try to reach out to BSD communities to gather feedback from them.

tests/types/bsd-test.json
```diff
@@ -0,0 +1,52 @@
+{
+  "$schema": "https://packageurl.org/schemas/purl-test.schema-1.0.json",
```
illuusio, line 2: Is this correct URL or is server down as this just returns HTTP server is too slow for me

types-doc/bsd-definition.md
```diff
@@ -0,0 +1,49 @@
+<!--  NOTE: Auto-generated from the JSON PURL type definition.
+Do not manually edit this file. Edit the JSON type definition instead. -->
+
+# PURL Type Definition: bsd
+
+- **Type Name:** BSD
+- **Description:** BSD Operating System variant packages
+- **Schema ID:** `https://packageurl.org/types/bitname-definition.json`
+
```
illuusio, line 9: Why choose Bitname for types?

types-doc/bsd-definition.md
```diff
@@ -0,0 +1,49 @@
...
+## Namespace definition
+
+- **Requirement:** Required
+- **Native Label:** vendor
+- **Note:** `The namespace is the "vendor" name such as "freebsd", "openbsd", or "netbsd". It is not case sensitive and must be lowercased.`
+
+## Name definition
+
+- **Native Label:** name
+- **Note:** `The name is the component name. It must be lowercased.`
```
illuusio: This requirement is bit complicated. How about KiCAD or FreeCAD for example whom likes that their name is written in strange way or X packages. There should be possibility to have also uppercase chars and how about is this alpha-numeric and how about packages like `py311-my-python-package3` or `package_name`

types-doc/bsd-definition.md
```diff
@@ -0,0 +1,49 @@
...
+- **Note:** `The version is the version of the package.`
+
+## Qualifiers Definition
+
+| Key  | Requirement | Native name | Default Value | Description |
+|------|-------------|-------------|---------------|-------------|
+| arch | Optional |  |  | The target architecture for the package. |
+| distro | Optional |  |  | The BSD release version or distribution variant. |
+| epoch | Optional |  |  | The package epoch number used for version comparison when normal version ordering is insufficient. |
+
```
illuusio, line 45: As FreeBSD is going thru [pkgbasecifcation](https://wiki.freebsd.org/PkgBase)  how one could write that package is from base FreeBSD 15.1 and not ports? I would suggest add keys like these (they can go under different name):

- branch - which would be is it unstable or stable version
- source - source of package like core, kernel or ports

With these one could write something like:
`pkg:bsd/freebsd/mkuzip@15.1?source=core&branch=stable`
`pkg:bsd/freebsd/mkuzip@15.1?source=core&branch=releng/15.1`

Then would URL would translate to `https://pkg.freebsd.org/15/base_latest` or `https://pkg.freebsd.org/15/base_release_1` not ports one.

illuusio (commented):
I've looked this PR thru FreeBSD developer glasses. It seems be solid and can be used as-is without applying this review. It would nice to have more eyes to look at this from NetBSD and OpenBSD.

types-doc/bsd-definition.md
s-kenji: We should use package names exactly as they are stored in the pkg database. Removed `It must be lowercased.`.

tests/types/bsd-test.json
s-kenji: Not sure. I just followed other purl type examples. 
@pombredanne Do you know anything about this?

types-doc/bsd-definition.md
s-kenji: Thanks, fixed.

types-doc/bsd-definition.md
s-kenji: @illuusio I agree to add something qualifier to distinguish how package is installed. But how can we get these information for each installed package? 
It appears that this information does not exist in the pkg database.

illuusio: In my opinion  we have to trust people how write these PURLs. In perfect world it would be nice to have such info straight from pkg out generated. I don't know is this irrelevant to this but there were discussion about bit similar to this in mailing list  [PKGBASE Removes FreeBSD Base System Feature](https://lists.freebsd.org/archives/freebsd-pkgbase/2025-July/000590.html) few weeks ago.

s-kenji: @illuusio Maybe we should leverage the existing [known qualifiers](https://github.com/package-url/purl-spec/blob/main/PURL-SPECIFICATION.rst#known-qualifiers-keyvalue-pairs) such as `repository_url`, `download_url`, and `vcs_url` to represent how packages are installed. What do you think about using the known qualifiers?

types-doc/bsd-definition.md
illuusio: Is there need to say anything about umlauts or similar stuff? How is UTF-8 things handled or is PURL similar to URL that one can decode inside with URL decoding if it's difficult character?

types-doc/bsd-definition.md
illuusio: I think they are needed at the beginning and hopefully magically things will be sorted out and they are not needed anymore as they make PURLs look gluttered to me. If they don't exist then just assume installation location is default one.

s-kenji: I agree. I think we should keep it simple at this point:

1. As per PURL specification, these qualifiers should be omitted when using default installation sources, which helps maintain cleaner and more readable PURLs.

2. Rather than defining new qualifiers now, we could revisit this discussion when FreeBSD's pkg database actually supports storing and retrieving this information. This approach would allow us to design qualifiers that align with FreeBSD's actual implementation.

This way, we can avoid premature specification while ensuring future extensibility when the underlying infrastructure supports it.

illuusio: I think path goes that as 1. suggest we follow KISS and after this one has landed to PURL then start 2. as it can be little tricky part. As said we can't assume what will be supported by `pkg(8)` but I think it's doable.

illuusio:
@s-kenji what do you think should this be narrowed down to FreeBSD if there is not interest at this point of time? As this is made modular rest of the BSDs could be added later on when time is ready for them.

s-kenji:
@illuusio Yes, we reached out to OpenBSD and NetBSD communities, but we haven't heard from them.
So narrowing down to FreeBSD for now, and updating the spec if necessary sounds reasonable to me. 

I'll update this PR accordingly.

s-kenji:
@illuusio  I've updated this PR to primarily focus on FreeBSD, but to be able to extend it to other BSD variants in the future.

Could you review this change and approve this PR if this looks good to you?

types-doc/bsd-definition.md
illuusio: This one needs to be probably added in some point but as I can't get anything out of that domain or URL it probably does not matter.

illuusio:
> @illuusio I've updated this PR to primarily focus on FreeBSD, but to be able to extend it to other BSD variants in the future.
> 
> Could you review this change and approve this PR if this looks good to you?

I approve latest version. As I don't have admin/commit rights I can't push the approve button but hopefully @pombredanne can.

tests/types/bsd-test.json
pombredanne: The server will be up in the new few days ;)

s-kenji:
@pombredanne @johnmhoran We have received approval from the FreeBSD community (through @illuusio) for this purl spec proposal.

We also reached out to OpenBSD and NetBSD communities but haven't received any responses. Therefore, we decided to proceed with the FreeBSD specification for now. Since we have defined a minimal spec for FreeBSD, we believe this can be extended to support other BSD variants in the future if needed.

If this looks good to you, could you approve and merge this PR?

pombredanne:
> FreeBSD

@s-kenji you are awesome! If I wanted to start a flame war, I could have said that there is only one BSD anyway. (But I will not tell you which one :angel: :smiling_imp: )

types/bsd-definition.json
```diff
@@ -0,0 +1,44 @@
...
+  "type_name": "BSD",
+  "description": "BSD Operating System variant packages, primarily focused on FreeBSD with extensibility for other BSD variants",
+  "repository": {
+    "use_repository": true,
+    "note": "The default repository is http://pkg.freebsd.org for FreeBSD. Other BSD variants may use different repositories as appropriate."
+  },
+  "namespace_definition": {
+    "native_name": "vendor",
+    "is_case_sensitve": false,
+    "note": "The namespace is the BSD variant name, primarily \"freebsd\" with support for other BSD variants. It is not case sensitive and must be lowercased.",
```
pombredanne: Could you spell out explicitly netbsd and openbsd at least?

iamleot: Maybe all `bsd` (both in the schema name and files) should be changed to `freebsd`. I think both OpenBSD ports and NetBSD - in that case pkgsrc, that can be used to most Unix-like operating systems - will need a dedicated type.

(I've followed a bit the discussion but I need to dig deeper and learn more about both this spec... I think I can help in a possible `pkgsrc` spec.)

s-kenji: @iamleot Thank you for your input. I've looked into this more carefully.

While pkgsrc is indeed designed for cross-platform use across many Unix-like systems (which justifies its own `pkg:pkgsrc` namespace), I found that OpenBSD ports is actually specific to OpenBSD, tightly integrated with the OS like other BSD package managers.

Looking at the various BSD package managers:
- FreeBSD ports/pkg: FreeBSD-specific
- OpenBSD ports/pkg_add: OpenBSD-specific
- DragonFly DPorts: DragonFly BSD-specific
- NetBSD: uses pkgsrc

Given that most BSD systems have their OS-specific package managers (except NetBSD which uses the cross-platform pkgsrc), I believe using `pkg:bsd/*` namespace still makes sense for these OS-specific package managers.

So I would suggest:
- Keep `pkg:bsd/*` for OS-specific BSD package managers
- Use `pkg:pkgsrc` for pkgsrc packages

Would you share your thoughts on this perspective?

iamleot: Sounds fine to me!

I'm unfamiliar with OpenBSD ports and Dragonfly DPorts though, so I have no idea regarding them.

s-kenji: @pombredanne I've updated the note to explicitly list the BSD variants (FreeBSD, OpenBSD, and DragonFlyBSD), while incorporating feedback from @iamleot (from the NetBSD community) about NetBSD's package manager needing its own type due to its cross-platform nature.

Could you take a look at the PR again?

kaiorafael:
The comments say that we should have something like as follows, right?

* FreeBSD
`pkg:bsd/freebsd/emacs@30.1_2?distro=14.3&epoch=3`
* NetBSD
`pkg:pkgsrc/netbsd/emacs@29.4?distro=10.0`
* OpenBSD
`pkg:bsd/openbsd/emacs@30.2-pgtk?distro=7.7`

s-kenji:
@pombredanne @johnmhoran Hi, I'm following up this PR regarding the BSD PURL type proposal.

Quick status update:
- We've received positive feedback and agreement from the FreeBSD community
- NetBSD may introduce their own type if needed, as they primarily use pkgsrc (which is cross-platform and not BSD-specific)
- While we haven't received responses from other BSD variants, having FreeBSD's agreement allows us to proceed with the implementation

We would like to move forward with this specification as FreeBSD is ready to adopt this PURL type. 
Could you please review the current specification when you have a chance? 

Thank you for your time!

illuusio:
Should topic be changes to `Add FreeBSD package type` to show that this has been narrowed down.

mjherzog:
Yes - I updated the title.

alice-sowerby:
@s-kenji thank you for advocating for this. @pombredanne and @johnmhoran please let us (me and @illuusio) know if you have any questions on this PR. 🙏

illuusio:
Should these update commits be merged together. It's bit bad habit to rebase but at least the updates as now there is 12 commits but only one correct..

PR #497, opened 2025-06-27
feat: Clarify case-folding behavior

### LINKED ISSUE #437: Define the "lowercase" operation in parser

ppkarwasz:
This change explicitly defines "lowercase" in terms of the **culture-invariant full case mapping** specified in the Unicode Standard. This avoids potential issues caused by locale-dependent casing operations and ensures consistent behavior across implementations.

Fixes #437

### DIFF
--- a/PURL-SPECIFICATION.rst
+++ b/PURL-SPECIFICATION.rst
@@ -295,7 +295,7 @@ full case mapping defined in
 `Section 3.13.2 of the Unicode Standard <https://www.unicode.org/versions/Unicode16.0.0/core-spec/chapter-3/#G34078>`_.
 
 When applied to the ASCII character set, this operation converts uppercase
-Latin letters (``A``–``Z``) to their corresponding lowercase forms (``a``–``z``).
+Latin letters (``A to Z``) to their corresponding lowercase forms (``a to z``).
 All other ASCII characters remain unchanged.
 
 How to build ``purl`` string from its components
--- a/PURL-SPECIFICATION.rst
+++ b/PURL-SPECIFICATION.rst
@@ -55,7 +55,7 @@ sometimes look like a ``host`` but its interpretation is specific to a ``type``.
 
 
 Some ``purl`` examples
-----------------------
+~~~~~~~~~~~~~~~~~~~~~~
 
 ::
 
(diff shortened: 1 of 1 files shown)


pombredanne (changes requested):
@ppkarwasz Thanks. I like that, but can you remove other unrelated cosmetic changes so that the PR is only about lowercase?

ppkarwasz:
> I like that, but can you remove other unrelated cosmetic changes so that the PR is only about lowercase?

I restored the previous headings in https://github.com/package-url/purl-spec/pull/497/commits/13db87e270f4d084cafdf4f71d9cdf283feec713. I thought heading levels in ReStructuredText had predefined characters (`-` for first level, `~` for second level), but apparently any character works as long as it is consistent.

PURL-SPECIFICATION.rst
```diff
@@ -286,6 +287,16 @@ Character encoding
...
 
+Case folding
+~~~~~~~~~~~~
+
+References to "lowercase" in this specification refer to the **culture-invariant**
+full case mapping defined in
+`Section 3.13.2 of the Unicode Standard <https://www.unicode.org/versions/Unicode16.0.0/core-spec/chapter-3/#G34078>`_.
+
+When applied to the ASCII character set, this operation converts uppercase
+Latin letters (``A``–``Z``) to their corresponding lowercase forms (``a``–``z``).
```
pombredanne: This would align the definition with Permitted Characters:
```suggestion
Latin letters (``A to Z``) to their corresponding lowercase forms (``a to z``).
```

Issue #369, opened 2024-12-11
PURLs for terraform providers as opposed to modules?


willmurphyscode:
At https://github.com/package-url/purl-spec/blob/1951d217bde29590a73f075db4ab71cc00011459/PURL-TYPES.rst?plain=1#L633

we see that the PURL type `terraform` might be used for terraform "modules" in the future. However, the terraform ecosystem has two types of reusable bundles of code: Modules and Providers (for example, if you go to https://registry.terraform.io/, you see "Browse Providers" and "Browse Modules" as two top level choices).

Is the use of "modules" here generic, and meant to include providers? Or will something else be used for packages of type terraform provider?

I think we want PURLs of Terraform providers and of Terraform modules to be distinct, but I'm not sure how to achieve this in the PURL spec.

Thanks!

dev-bio:
I was just investigating this as well, I'm using a subpath until this is sorted: `pkg:terraform/a/b/c@0.0.0#providers`

@pombredanne any thoughts?

pombredanne:
@dev-bio sorry for the late reply! can you or @willmurphyscode or @spiffcs or @ghouscht propose the type definition for terraform in a PR?

ghouscht:
Hey @pombredanne, @dev-bio, @willmurphyscode
I created an MR with a proposal for a terraform type definition. I'm curious to hear what you think, please let me know 🙏🏻

PR #498, opened 2025-06-27
Add new PURL type: 'terraform'

ghouscht:
See #369 for details.

Please let me know what you think. This proposal is for now solely based on my ideas and I'm open for feedback. It might even make sense to define two distinct types `terraform-provider` and `terraform-module` for more clarity.

### DIFF
--- a/PURL-TYPES.rst
+++ b/PURL-TYPES.rst
@@ -113,7 +113,6 @@ Other candidate types to define
 - ``puppet`` for Puppet Forge packages:
 - ``sourceforge`` for Sourceforge-based packages:
 - ``sublime`` for Sublime packages:
-- ``terraform`` for Terraform modules
 - ``vagrant`` for Vagrant boxes
 - ``vim`` for Vim scripts packages:
 - ``wordpress`` for Wordpress packages:
--- a/purl-types-index.json
+++ b/purl-types-index.json
@@ -30,5 +30,6 @@
   "qpkg",
   "rpm",
   "swid",
-  "swift"
+  "swift",
+  "terraform"
 ]
--- a/types-doc/terraform-definition.md
+++ b/types-doc/terraform-definition.md
@@ -1,0 +1,42 @@
+<!--  NOTE: Auto-generated from the JSON PURL type definition.
+Do not manually edit this file. Edit the JSON type definition instead. -->
+
+# PURL Type Definition: terraform
+
+- **Type Name:** Terraform providers and modules
+- **Description:** Terraform providers and modules
+- **Schema ID:** `https://packageurl.org/types/terraform-definition.json`
+
+## PURL Syntax
+
+The structure of a PURL for this package type is:
+
+    pkg:terraform/<namespace>/<name>@<version>?<qualifiers>#<subpath>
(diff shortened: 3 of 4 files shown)


PURL-TYPES.rst
```diff
@@ -582,6 +582,28 @@ swift
...
+- The ``namespace`` is the provider or module namespace, which is usually the
+  organization or vendor name.
+- The ``name`` is the name of the provider or module.
+- The ``version`` is the package version and is required.
+- Optional qualifiers may include:
+
+  - ``registry_url``: A registry URL where the provider/module may be found, but
+    not intended as the only location. This value is encouraged to identify a
+    location the content may be fetched in case the default registry is not used.
+- The ``subpath`` is used to distinguish between a provider and a module.
```
kzantow: Just one thought: it seems fairly abnormal to _require_ a subpath. Is this something that is actually required to identify the terraform components or could it be made optional? In other words: is there overlap between provider and module where `pkg:terraform/hashicorp/aws@version` is unable to identify software because there are both providers and modules available at `hashicorp/aws` (or other names)? The fact this is required makes me think there is some overlap and if the case, maybe it's better to add this as the namespace? e.g. 
```
      pkg:terraform/provider/hashicorp/aws@6.0.0
      pkg:terraform/module/terraform-aws-modules/vpc@6.0.1
      pkg:terraform/module/terraform-aws-modules/vpc@6.0.1?registry_url=https://registry.example.com
```
... or lean into the other suggestion having multiple types, of `terraform-provider`, `teraform-module`, etc..

ghouscht: Fair point. I think in practice there is no overlap between modules and providers, but afaik there could be overlap. I agree that it might make sense to add it to the namespace instead of using a subpath.

pombredanne:
- See  #514

With the new approach... this needs to be updated. Can you take a stab at it? Thanks for your understanding !

ghouscht:
> After the merge of PR #514, PURL types are now defined in JSON: 😇 😁
> 
>     * See  [Refine Purl type schema #514](https://github.com/package-url/purl-spec/pull/514)
> 
> 
> With the new approach... this needs to be updated. Can you take a stab at it? Thanks for your understanding !

PR is updated with the new JSON format. Let me know what you think

PURL-TYPES.rst
ghouscht: I did think a bit more about this and I belive in reality there is no overlap between module and provider names. Because I had to update the PR to use the new JSON schema I got rid of this distinction in the subpath. Let me know what you think.

kzantow: Seems simpler to me, thanks! 👍

pombredanne:
@ghouscht this is a good PR to have... the PR fork should not have been deleted IMHO, but we can always get back to the content in all cases.

ghouscht:
> @ghouscht this is a good PR to have... the PR fork should not have been deleted IMHO, but we can always get back to the content in all cases.

Oh, honestly I forgot that this PR was open. I did some cleanup of forked repos today so this is why it got closed, sorry about that.

PR #508, opened 2025-07-01
Streamline Unicode character references

pombredanne:
This PR is only stylistic and does not change the spec. Its goal is to:

- Ensure that we use parallel constructions for all Unicode character references in the spec.
- Also align allowed decoded content section in subpath with that of namespace.

Reference: https://github.com/package-url/purl-spec/pull/485


pombredanne:
Thanks for the swift approval. Merging as it is.

PR #509, opened 2025-07-01
Remove plus from allowed character in type

pombredanne:
The plus was not used in any type and was always to be encoded elsewhere.
No more plusses!

Reference: https://github.com/package-url/purl-spec/issues/436


pombredanne:
OK... thanks! merging

PR #522, opened 2025-07-10
Add new PURL type: 'chrome-extension'

mixmix:
This adds support for Chrome Browser Extensions

The examples I've given have been acquired sampling the chrome store through scraping + querying an updated "updatecheck" API that chrome itself uses to poll for updates


PURL-TYPES.rst
```diff
@@ -102,6 +102,20 @@ cargo
...
+chrome
+-----
+``chrome`` for Chrome Browser Extensions:
+
+- The default repository is ``https://chromewebstore.google.com/``.
+- The ``name`` is the extension id - the human readable name often depends on locale.
+- The ``version`` is the package version.
+- Examples::
+
+      pkg:chrome/hlepfoohegkhhmjieoechaddaejaokhf@25.7.1
```
pombredanne: Thanks! this name look fairly weird, is this really the thing? Given an installed extension, where do you find this?

PURL-TYPES.rst
```diff
@@ -102,6 +102,20 @@ cargo
       pkg:cargo/clap@2.33.0
       pkg:cargo/structopt@0.3.11
 
+chrome
+-----
+``chrome`` for Chrome Browser Extensions:
+
+- The default repository is ``https://chromewebstore.google.com/``.
+- The ``name`` is the extension id - the human readable name often depends on locale.
+- The ``version`` is the package version.
```
pombredanne: Is this something that is also tracked with another number internally, like the id above?

mixmix: The places I know if that reference the version number:
- chrome web store page for an extensions lists current version
- the chrome web store `updatecheck` API - returns the current version
- the chrome web store blob URL - files suffixed with a modified form of the version (camel case... plus more?)
- the `manifest.json` has a `version` field

mixmix: NOTE: the extension `id` is NOT the same as the `name` in the `manifest.json` - the name is whatever human-readable name. This allows them to have a primary key that is very collision-free (32 char a-z), and a flexible locale/translation mutable "name" which is presented on web pages / UI.

PURL-TYPES.rst
mixmix: The unique `name` is a 32 char a-z string. This is NOT the same as the `name` in the `manifest.json` - that name is whatever human-readable name. This allows them to have a primary key that is very collision-free, and a flexible locale/translation mutable "name" which is presented on web pages / UI.

If you want to know the `name` of an installed extension, one way is to look at your extensions, and click "View in Chrome Web Store". For 1Password, this will take you to:
- https://chromewebstore.google.com/detail/aeblfdkhhhdcdjpifhhbdiojplfjncoa , which quickly redirects you to 
- https://chromewebstore.google.com/detail/1password-%E2%80%93-password-mana/aeblfdkhhhdcdjpifhhbdiojplfjncoa

Note that human readable "name" field can be absent, OR any string at all. It's totally not relied upon for identifying the extension

pombredanne:
- See  #514

Thanks for your understanding !

mixmix:
Thanks @pombredanne ! I've had a go reformatting this PR
I added some tests, but was unclear how to run those. I ran `make conf && make check` and got some passes but then an Exit 1 with no error message so unclear how to improve

I also note `README.dev.rst` is perhaps out of date (there is no `make docs` or make `generate`)?
I ran `make gendocs`

tests/types/chrome-test.json
```diff
@@ -0,0 +1,41 @@
...
+    },
+    {
+      "description": "invalid chrome name",
+      "test_group": "base",
+      "test_type": "parse",
+      "input": "pkg:chrome/kanfjhdeebkfgk@0.6",
+      "expected_output": null,
+      "expected_failure": true,
+      "expected_failure_reason": "Should fail to parse a PURL from invalid purl input"
+    }
```
mixmix: I'd like to see this fail, but I'm not sure the schema currently support checks for `name` (extension id) that is too short?

mixmix:
Please feel free to take what's here and adjust it as you see fit.
I'll look back here every few days to answer any questions too

Thanks for your work 🍍 🐟

tests/types/chrome-test.json
matt-phylum: I'd argue that PURL should do the absolute minimum in terms of validating the names of packages because:
1. It makes PURL implementations more complicated
2. It increases the likelihood of problems arising from inconsistent implementations
3. PURL does not control the validation rules and there is no reason that eg Google couldn't add a different format for extension IDs tomorrow as long as the new format cannot be confused with the current format, which would be a real problem for users of software that is validating the extension ID matches a certain pattern

mixmix: Good points! I'll remove this test

mixmix:
```suggestion
    }
```

mixmix:
@pombredanne (or someone else), is there anything I can do to unblock merging this?

pombredanne:
> @pombredanne (or someone else), is there anything I can do to unblock merging this?

@mixmix Thanks! It would be awesome to have someone from Chrome to have a quick look on this. Could reach out there?
@di gentle ping... you may be able to help too?

mixmix:
I filed a  "developer feature request" in [this chrome_webstore/contact form](https://support.google.com/chrome_webstore/contact/one_stop_support?hl=en)

> Hi I've filed a PR to register a Package URL (PURL) spec for Chrome Webstore packages.
https://github.com/package-url/purl-spec/pull/522
I'd like to get input from the Chrome Webstore team to make sure the format I've suggested feels aligned.

I'm searched for other ways to contact that team but couldn't find any @pombredanne

mixmix:
Response TL;DR ~= _"message Chrome Extensions Google Group"_

<details>
<summary> (full reply) </summary>

>
> Hello Mix,
>
> Thank you for contacting Chrome Web Store Developer Support.
> 
> In regards to your inquiry, as much as I'd like to provide the answer you're looking for, unfortunately I cannot give you a hard answer due to legal related matters.
>
> I would stress that review is ultimately responsible for determining compliance and that we do not offer pre-submission assessments. It's generally best to implement the best solution that will make it work, and then if there's a problem with it after submitting then the review team will let you know.  Please review the Chrome Web Store [Program Policy](https://developer.chrome.com/webstore/program_policies) and [Developer Agreement](https://developer.chrome.com/webstore/terms) and see if it contradicts any policies. Please [see here](https://developer.chrome.com/webstore/user_data) regarding User Data.
> 
> In the meantime, I would highly recommend posting in our [Chromium Extensions group](https://groups.google.com/a/chromium.org/g/chromium-extensions). Members of Chrome's extensions development team and Product team also post there and answer developer and development related questions. They'll be able to offer some additional insights based on your use case.
We greatly appreciate your understanding.
>
> Kind regards,
> 
> Catherine
> 
> Chrome Web Store Developer Support
</details>

Message posted to group: https://groups.google.com/a/chromium.org/g/chromium-extensions/c/jR7mqtIwQ3A

PURL-TYPES.rst
patrickkettner: Hi! Patrick from the Chrome Extension team here - @mixmix is correct. This extension ID is 100% unique to the individual item, whereas the `name` field can be any string, and as a result there can easily be collisions.

patrickkettner:
Hiya! Patrick from the Chrome Extension team - happy to do whatever, but unsure what is being asked.

mixmix:
Hi @patrickkettner I'm essentially asking if you can see any problems with the following as standard ways to reference chrome extensions as purls (package URLs):

- `pkg:chrome/hlepfoohegkhhmjieoechaddaejaokhf@25.7.1`
- `pkg:chrome/dncgedbnidfkppmdgfgidcepclnokpkb@6.0.2.3611`
- `pkg:chrome/kanfjhdeebkfgkbmnfknhejpadhlmiab@0.6`
- `pkg:chrome/dlpngalgnefjeiefhmpklpfiohadpglk@1`
- `pkg:chrome/dlpngalgnefjeiefhmpklpfiohadpglk` < means latest

We have the chrome `extensionId` as the "name", because that's the UID. The version is non-semver, and I've chosen to support what is shown on the extension pages as a version, and what we see in the `manifest.json`. Internally I noticed you seem to normalize to something like `\d+(\.\d+){0,3` (i.e. 1 to 4  dot-seperated numbers), but I've not used those because purls are going to be something user-entered often

patrickkettner:
That all SGTM. The only thing that comes to mind is that similar to how nodejs can specify `engines`, extensions can set a `minimum_chrome_version`. Is it worthwhile to define qualifiers or anything like that to ensure the user isn't requesting a version that they can't use?

For what its worth, the Edge marketplace and mozilla add-on store would share identical logic to what you created

Regarding the note ```"There is currently no documented API for querying different versions of a package - there only seems to be responses on the latest version. To this end the version component of a chrome purl can be optional."```, that is correct as of today (but in theory _could_ change one day). Only the latest version of an item is publicly available, but specifying a URL may be a good way to ensure an update isn't unexpected?

mixmix:
I think the idea of an `engines` query in interesting. Is this what the `prodversion` query is used to resolve on the `updatecheck` API? (we're guessing from the outside!). I think if that's what you're doing then adding support for `engines` could cool. Would love to hear from PURL maintainers if that's a pattern that's supported elsewhere (cc @pombredanne)

To be concrete @patrickkettner , would you expect the purl would then be something like
- `pkg:chrome/dlpngalgnefjeiefhmpklpfiohadpglk?engine=107.0.01` => the latest version that supports this

mixmix:
>  Only the latest version of an item is publicly available, but specifying a URL may be a good way to ensure an update isn't unexpected?

@patrickkettner can you clarify more. Are you saying you think it would be good to NOT have `pkg:chrome/dlpngalgnefjeiefhmpklpfiohadpglk` as a generic "references latest version" PURL, because it may lead to unexpected updates? I think I would normally agree, but it also seems chrome extensions do un-announced updates as the norm, so _in this context_ discussing the latest is generally what you will be doing?

mixmix:
@patrickkettner , polite bump on this ❤️

patrickkettner:
Appreciate it! 

> think the idea of an engines query in interesting. Is this what the prodversion query is used to resolve on the updatecheck API? 

Sorry, I think my suggestion was not clear. I was making an analogy between how nodejs has the [engines](https://docs.npmjs.com/cli/v9/configuring-npm/package-json#engines) field in package.json. This allows a project to say "only run this code in node version N.NN". In a similar manor, a Chrome Extension have the [minimum-chrome-version](https://developer.chrome.com/docs/extensions/reference/manifest/minimum-chrome-version), which lets a project say "this cannot run on versions of Chrome lower than N.N.N.N". I was suggesting that if someone was using the PURL, they may want to say something like "because I am in a deployment that must support ${OLDER_VERSION_OF_CHROME}, can I make sure that the extension is NOT going to break because of my chrome version?". 

In practice this could look like 

`pkg:chrome/dlpngalgnefjeiefhmpklpfiohadpglk?minimum_chrome_version=107.0.01` 

where the chrome extension being downloaded would _only_ resolve if 
 - it has no `minimum_chrome_version` in it's manifest.json, or
 - the version set in it's manifest.json `minimum_chrome_version` is less than or equal to the value specified (107.0.01)

> Are you saying you think it would be good to NOT have pkg:chrome/dlpngalgnefjeiefhmpklpfiohadpglk as a generic "references latest version"

No, not saying that. I think for most people it works great. But for some folks that want extreme control, and they want to manually opt-in/review every update, that could be useful.

mixmix:
I think your suggestion of `minimum_chrome_version` is interesting. I'm hesitant to add it to this spec without clarity about how/ whether this actually works with any of the google APIs at the moment.

A little scouting + experimenting surfaced:
1. In `updatecheck` API, varying `prodversion` gets you from results to no result...
  - chose from [this list over version tags](https://chromium.googlesource.com/chromium/src/+refs)
  - very unclear what no results means, would need to test against a known long-running extension which has updated it's "minimum_chrome_version"?
2. Doesn't seems Chrome allows you to download old versions?
  - maybe I missed a link, but I only found blogs talking about building your own from a Chromium Release at a tag?

I'd like to propose we shelve this (put "minimum_chrome_version" on hold) unless we can clarify if there is any feasible way for users to use it currently. Once there is, then we should definitely consider adding this. This would also allow us to proceed with merging this I hope

patrickkettner:
Apologies for making this confusing. 

`minimum_chrome_version`  is not exposed as an API. It is a value in the manifest.json file of the CRX extension. If it was supported, the file would need to be parsed and that value extracted.
There is no public API to download old versions of an extension. However using the `minimum_chrome_version` can prevent an extension from being _updated_ if requested.


in short, your current existing proposal is good from my point of view.

mixmix:
@pombredanne I think all outstanding comments/ question have been resolved on this
@patrickkettner if you'd like to leave a review confirming your position on this that might help accelerate this getting merged

mjherzog:
@mixmix It would probably be more accurate to set the PURL type to "chrome-extension" to be more descriptive and to leave "space" just in case there are future PURL types related to chrome. What do you think?

mixmix:
I get the defensive stance. I cannot imagine what other things might inhabit `chrome:` as a type... I guess you could reference a version of chrome itself... but that feels more like `chromium:` or `electron:`. Can you come up with any example ideas which would possible justify this change? (Full disclosure, I've been building on this specification, so I'd rather not change thing unless there is a really good reason)

pombredanne:
re: https://github.com/package-url/purl-spec/pull/522#issuecomment-3795099961

> @mixmix It would probably be more accurate to set the PURL type to "chrome-extension" to be more descriptive and to leave "space" just in case there are future PURL types related to chrome. What do you think?

@patrickkettner what's your take? @oliverdunk do you have an opinion on the topic?

mjherzog:
@mixmix My concern about the name is more about the **type** being clear and descriptive.  I appreciate that you would like to avoid rework on your project, but our primary concern is to get PURL types right when they are registered.

mixmix:
If we're going for clarity, which other purls would this also effect? This change would suggest for example `pkg:vscode` ([already merged](https://github.com/package-url/purl-spec/pull/673)) should really be changed to `pkg:vscode-extension`... oh wait you proposed that change and it was merged - I missed it as I was travelling, which is kinda frustrating. I spent a decent amount of energy trying to clarify that over months, and it was changed quite quickly.

Are things merged into main immutable? If they are, this kinda forces the move to `chrome-extension`. Lets give it a moment to hear other responses before finalising that.

mixmix:
Have considered this more, and sought advice. I think your suggestion is a good one @mjherzog , have made the change

types/chrome-extension-definition.json
```diff
@@ -0,0 +1,34 @@
...
+    "default_repository_url": "https://https://chromewebstore.google.com/",
+    "note": "There is no documented API, only a sitemap and 'updatecheck' endpoints."
+  },
+  "namespace_definition": {
+    "requirement": "prohibited",
+    "note": "There is no namespace"
+  },
+  "name_definition": {
+    "native_name": "version",
+    "case_sensitive": false,
```
pombredanne, line 20: With this, what about adding a `permitted_characters` pattern? Something like:

```suggestion
    "case_sensitive": false,
    "permitted_characters": ".....", 
```

types/chrome-definition.json
```diff
@@ -0,0 +1,34 @@
...
+    "use_repository": true,
+    "default_repository_url": "https://https://chromewebstore.google.com/",
+    "note": "There is no documented API, only a sitemap and 'updatecheck' endpoints."
+  },
+  "namespace_definition": {
+    "requirement": "prohibited",
+    "note": "There is no namespace"
+  },
+  "name_definition": {
+    "native_name": "version",
```
pombredanne: This does not look right and my 1st reaction would be to use `"native_name": "name",` 

But then.... the name in the manifest may not be that name ... see https://developer.chrome.com/docs/extensions/reference/manifest#popup-with-permissions mentions: 

> name: A string that identifies the extension in the Chrome Web Store, the install dialog, and the user's Chrome Extensions page (chrome://extensions). The maximum length is 75 characters. For information on using locale-specific names, see Internationalization.

So this 32 char name is not the 75 char name... but something else... some hints about what this is would be great.

types/chrome-definition.json
```diff
@@ -0,0 +1,34 @@
+{
+  "$schema": "https://packageurl.org/schemas/purl-type-definition.schema-1.0.json",
+  "$id": "https://packageurl.org/types/chrome-definition.json",
+  "type": "chrome-extension",
+  "type_name": "Chrome Browser Extensions",
+  "description": "Chrome Browser Extensions (extensions from the official Chrome Webstore)",
```
pombredanne: Is there some more you can add there? Where are the data in the PURL found? in the manifest.json? or the store API? is there a URL for some reference docs? For instance I found this https://developer.chrome.com/docs/extensions/reference/manifest#popup-with-permissions .... but is this the proper doc?

types/chrome-definition.json
```diff
@@ -0,0 +1,34 @@
...
+  },
+  "name_definition": {
+    "native_name": "version",
+    "case_sensitive": false,
+    "note": "The name is a 32 characters a-z and is case insensitive"
+  },
+  "version_definition": {
+    "native_name": "version",
+    "requirement": "optional",
+    "note": "The chrome extension version is semver-like but 1-4 segments"
```
pombredanne: Could it help to also add the URL to https://developer.chrome.com/docs/extensions/reference/manifest/version in the notes ?

pombredanne (dismissed):
Thanks for the update. I have some nitpickings for your consideration.

types/chrome-extension-definition.json
mixmix: It *appears* to be a-z... but I've not seen a spec so I'm scared to lock that in?

types/chrome-definition.json
mixmix: id, extension_id?

types/chrome-definition.json
mixmix: There is no official documention that I know of. I can link to unofficial docs?

mixmix:
Thanks for the notes! I'll loop back next week

types/chrome-extension-definition.json
pombredanne: @mixmix fair enough!

types/chrome-definition.json
pombredanne: could there be a doc pointer?

types/chrome-definition.json
pombredanne: >  I can link to unofficial docs?

Any doc is better than none IMHO. You can at least link in a comment here for a start.

mjherzog:
@mixmix One more change request. The name for the PURL type definition file should be "chrome-extension-definition.json" instead of "chreome-definition.json". The name for the test file is "chrome-extension-test.json" so we are good there.
Thanks

types/chrome-definition.json
mixmix: The closest I've seen is e.g. https://developer.chrome.com/docs/extensions/how-to/distribute/install-extensions#prereq-cws
Calls it "Extension ID", so I'm going to go with that

patrickkettner: this is the correct terminology for chrome extensions

types/chrome-definition.json
mixmix: Have linked to the sources of info I've used

types/chrome-extension-definition.json
mixmix: I take it back, I've tested and less than and greater than 32 char is not supported, so I'll lean into this validation. Couldn't see good examples so hope this is correct

types/chrome-definition.json
mixmix: DONE

mixmix:
@pombredanne @mjherzog I've addressed all your concerns I think.
I've added tests for the patterns I've added, but have not figured out how one runs the tests to prove my patterns are correctly formatted to work....

Let me know if there is anything else

tests/types/chrome-extension-test.json
```diff
@@ -0,0 +1,69 @@
...
+    },
+
+    {
+      "description": "invalid chrome purl (name: invalid characters)",
+      "test_group": "base",
+      "test_type": "parse",
+      "input": "pkg:chrome-extension/44444algnefjeiefhmpklpfiohadpglk",
+      "expected_output": null,
+      "expected_failure": true,
+      "expected_failure_reason": "permitteds_characters"
```
mjherzog: for this and following test cases please fix the typo "s" in "permitteds_characters"
"expected_failure_reason": "permitted_characters"

tests/types/chrome-extension-test.json
```diff
@@ -0,0 +1,69 @@
...
+      "description": "invalid chrome purl (name: wrong length)",
+      "test_group": "base",
+      "test_type": "parse",
+      "input": "pkg:chrome-extension/dogs",
+      "expected_output": null,
+      "expected_failure": true,
+      "expected_failure_reason": "permitteds_characters"
+    },
+    {
+      "description": "invalid chrome purl (version: too many parts  )",
```
mjherzog: For description - probably better to say (version: too many segments) to match the terminology in the type definition for version: - **Note:** `The chrome extension version is semver-like but 1-4 segments. See https://developer.chrome.com/docs/extensions/reference/manifest/version`

mjherzog (changes requested):
@mixmix This looks very close. I found two typo-level errors to correct.
Regarding tests - it is ideal for you to use packageurl-python or one of the other tools to run your tests but that is not currently a requirement for adding a PURL type.

tests/types/chrome-extension-test.json
mixmix:
```suggestion
      "expected_failure_reason": "permitted_characters"
```

tests/types/chrome-extension-test.json
mixmix:
```suggestion
      "description": "invalid chrome purl (version: too many segments)",
```

tests/types/chrome-extension-test.json
```diff
@@ -0,0 +1,69 @@
...
+      "expected_failure_reason": "permitteds_characters"
+    },
+    {
+      "description": "invalid chrome purl (name: wrong length)",
+      "test_group": "base",
+      "test_type": "parse",
+      "input": "pkg:chrome-extension/dogs",
+      "expected_output": null,
+      "expected_failure": true,
+      "expected_failure_reason": "permitteds_characters"
```
mixmix:
```suggestion
      "expected_failure_reason": "permitted_characters"
```

tests/types/chrome-extension-test.json
```diff
@@ -0,0 +1,69 @@
...
+    },
+    {
+      "description": "invalid chrome purl (version: too many parts  )",
+      "test_group": "base",
+      "test_type": "parse",
+      "input": "pkg:chrome-extension/dlpngalgnefjeiefhmpklpfiohadpglk@1.2.3.4.5",
+      "expected_output": null,
+      "expected_failure": true,
+      "expected_failure_reason": "permitteds_characters"
```
mixmix:
```suggestion
      "expected_failure_reason": "permitted_characters"
```

tests/types/chrome-extension-test.json
```diff
@@ -0,0 +1,69 @@
...
+    },
+    {
+      "description": "invalid chrome purl (version: invalid characters)",
+      "test_group": "base",
+      "test_type": "parse",
+      "input": "pkg:chrome-extension/dlpngalgnefjeiefhmpklpfiohadpglk@1.2.3-beta",
+      "expected_output": null,
+      "expected_failure": true,
+      "expected_failure_reason": "permitteds_characters"
```
mixmix:
```suggestion
      "expected_failure_reason": "permitted_characters"
```

mixmix:
Thanks for the close eye @mjherzog , changes made

mjherzog (approved):
Confirmed that all open issues / feedback have been resolved

mjherzog:
@mixmix I have approved the changes and expect that we can merge by tomorrow.
A related question is for the www.packageurl.org website where we show the registered PURL types at:
https://www.packageurl.org/docs/purl/purl-spec-purl-types#registered-purl-types
Can you provide information for an icon to use in that display?

mixmix:
For Chrome, I'd recommend an icon like this svg

chrome_logo

mixmix:
Hey @pombredanne can you check in sign off on this PR? You're marked as "Requested Change".
Would love to see this merged

mjherzog (commented):
This type is ready to register. Good to see failing tests.

johnmhoran (approved):
LGTM, and I agree with @mjherzog -- it's good to see a group of expected-to-fail tests.  :-)

Issue #489, opened 2025-06-23
Confirm primary license for purl-spec and vers-spec


mjherzog:
The current license for purl-spec is MIT. MIT does not seem to be the optimal license for a specification because the primary definition is that it covers  "software and associated documentation files (the "Software")". Although PURL and VERS specification repos will include some software the primary content is the documentation of the specifications and standards, not the software. 
The license for the PURL and VERS standards will ultimately be the Ecma Copyright Notice - https://ecma-international.org/policies/by-ipr/ecma-text-copyright-policy/, but we will need a broader license for the overall specification. The Linux Foundation is actively promoting their Community Specification license - (https://github.com/CommunitySpecification/Community_Specification/blob/main/1._Community_Specification_License-v1.md), but this license seems to be focused on patents which do not seem relevant to PURLs.

I suggest that we change the base license to Apache 2.0 which is clearly a Permissive license and which offers definitions of Works, Derivative Works and Contributions to cover specification documentation beyond software and its documentation. Apache 2.0 is the current license for the CycloneDX specification.

mjherzog:
Further research from: https://ecma-international.org/policies/by-ipr/ecma-international-policy-on-submission-inclusion-and-licensing-of-software/ indicates that an Ecma software license applies to the inclusion of software in a standard in addition to descriptive text. Our plan is that Ecma standard will include only text (including text in JSON.Schema format) so it appears that we do not need a software license for the standard, but we will want a license to cover other code that we may include in the PURL Core Spec and utilities like those to generate documentation from schema files.

mjherzog:
Upon further review we seem to have 3 areas where we need to clarify the license:
- The content of the standard will be under the Ecma Copyright Notice - https://ecma-international.org/policies/by-ipr/ecma-text-copyright-policy/ - this is set.
- The code (scripts for running tests, generating documentation, etc.) is fine under MIT
- The specification/documentation that is not part of the standard should be licensed under a more documentation-friendly license than MIT. This should likely also be the license for the purl-type data.    
   - This could be the Community Specification License 1.0 or CC-BY-4.0 or ?
   - This could be the Apache 2.0 license which then could also be used for the code (instead of MIT) to simplify overall licensing.

pombredanne:
I like going to Apache, like with CycloneDX.
- It is not as complex as the Community Spec license
- It has patent protection
- It works OK for code and docs
- MIT is sublicenseable alright to Apache, but I want to make sure we get proper, wide community buy-in.

[...]

PR #542, opened 2025-07-27
Licensing improvements

sschuberth:
Please have a look at the individual commit messages for the details.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -180,4 +180,4 @@ See the `dedicated adopters list <ADOPTERS.rst>`_.
 License
 ~~~~~~~
 
-This document is licensed under the MIT license
+The `purl` specification is licensed under the `MIT license <LICENSE>`_.


pombredanne:
Thanks. This brings an interesting discussion with the Ecma standardization...

sschuberth:
> This brings an interesting discussion with the Ecma standardization...

Like what exactly? Do you have any link to an existing discussion?

LICENSE
mjherzog, line 1: We have a pending Issue to confirm the license for the specification versus code -see https://github.com/package-url/purl-spec/issues/489. MIT may make sense for code in the purl-spec repo, but it is not a good license for a specification and the Ecma standard subset of the specification will be subject to https://ecma-international.org/policies/by-ipr/ecma-text-copyright-policy/

PR #543, opened 2025-07-27
Use proper spelling of the word "roundtrip"

lizmat:
Something that annoyed me while going through the newly formatted test-suite  :-)


mjherzog:
@lizmat From some dictionary research (Merriam-Webster) the correct spelling for round trip as an adjective is "round-trip". We will use that in the test suite upgrade

PR #545, opened 2025-07-27
Fix case_sensitive property name in types

maennchen:
Property is called `case_sensitive` and not `is_case_sensitive` in schema:

https://github.com/package-url/purl-spec/blob/ea4bc0c02940ea51dc1468da1abf9780b0654eb6/schemas/purl-type-definition.schema.json#L35-L40


maennchen:
Same changes done in #640, closing.

PR #536, opened 2025-07-24
Add "raku" as a valid package URL type

lizmat:
This is my attempt at providing the necessary information to make "raku" a fully supported Package URL type for distributions in the [Raku Programming Language](https://raku.org).

Please note that I recently started providing support for CycloneDX SBOMs with the following modules in the Raku Programming Language:
- [PURL](https://raku.land/zef:lizmat/PURL)
- [SBOM::CycloneDX](https://raku.land/zef:lizmat/SBOM::CycloneDX)
- [SBOM::Raku](https://raku.land/zef:lizmat/SBOM::raku)

This PR adds:
- a section in PURL-TYPES.rst
- adds some tests in the test-suite-data.json

This is new to me, so I probably have made / will make mistakes.  Any comments, suggestions, fixes will be thoroughly appreciated!

This PR also removes the mention of "perl6", as Perl 6 has been renamed to the Raku Programming Language" in 2019.

PR #550, opened 2025-07-28
Add Raku Programming Language as a purl-spec type

lizmat:
- Reworked PR #536 in the new format after PR #514 has been merged, and processed some new insights as well.  Plus more tests.

### DIFF
--- a/docs/candidate-purl-types.md
+++ b/docs/candidate-purl-types.md
@@ -4,7 +4,7 @@
 - ``apache`` for Apache projects packages
 - ``atom`` for Atom packages
 - ``bower`` for Bower JavaScript packages
-- ``brew`` for Homebrew packages
+- ``brew`` for Homebrew packages - See https://github.com/package-url/purl-spec/pull/796
 - ``buildroot`` for Buildroot packages
 - ``carthage`` for Cocoapods Cocoa packages
 - ``chef`` for Chef packages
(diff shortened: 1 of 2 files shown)


types/raku-definition.json
```diff
@@ -0,0 +1,39 @@
...
+      "description": "URL of package or distribution"
+    }
+  ],
+  "examples": [
+    "pkg:raku/zef:lizmat/SBOM::CycloneDX@0.0.7",
+    "pkg:raku/cpan:TIMOTIMO/JSON::Fast@0.19",
+    "pkg:raku/zef:vrurg/Cro::RPC::JSON@0.1.6:2",
+    "pkg:raku/zef:lizmat/駱駝道@0.0.1",
+    "pkg:raku/zef:lizmat/SBOM::CycloneDX@0.07?download_url=https://360.zef.pm/S/BO/SBOM_CYCLONEDX/e9df9ac7d473c289aae4df19ea10c0289bb04851.tar.gz",
+    "pkg:raku/cpan:TIMOTIMO/JSON::Fast@0.19?download_url=https://cpan.metacpan.org/authors/id//T/TI/TIMOTIMO/Perl6/JSON-Fast-0.19.tar.gz"
```
pombredanne: I think we can fully resolve the download URLs from the PURL.

lizmat: Indeed, so pushed!

tests/types/raku-test.json
```diff
@@ -0,0 +1,277 @@
...
+  "tests": [
+
+    {
+      "description": "valid Raku dist name without api",
+      "test_group": "base",
+      "test_type": "build",
+      "input": {
+        "type": "raku",
+        "namespace": "zef:lizmat",
+        "name": "SBOM::CycloneDX",
```
pombredanne, line 12: @sjn @lizmat Just to be double sure before merging this... Raku dists are the problem target for a PURL?

PR #552, opened 2025-07-28
Make clear that any ruby gem may use a qualifier and not just the Jruby gem

darakian:
Broken out from https://github.com/package-url/purl-spec/pull/514 as its fairly minor, but worth being explicit about.

Nokogiri is another gem which makes heavy use of qualifiers https://rubygems.org/gems/nokogiri/versions

### DIFF
--- a/tests/types/gem-test.json
+++ b/tests/types/gem-test.json
@@ -2,7 +2,7 @@
   "$schema": "https://packageurl.org/schemas/purl-test.schema-1.0.json",
   "tests": [
     {
-      "description": "Java gem can use a qualifier. Rountrip an input purl to canonical.",
+      "description": "Ruby gems can use qualifiers. Rountrip an input purl to canonical.",
       "test_group": "advanced",
       "test_type": "roundtrip",
       "input": "pkg:gem/jruby-launcher@1.1.2?Platform=java",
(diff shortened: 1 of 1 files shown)


darakian:
Your ask for an enumerated list sent me down a rabbit hole. The short answer is I don't believe there is a complete list of platforms. 

The longer train of thought is that I first looked for a spec which lead to 
https://guides.rubygems.org/specification-reference/#platform=
and
https://bundler.io/man/gemfile.5.html
Neither is definitive, but the second lists a few possible platforms. Alas it also discusses some which are not in the list. Poking around a little more I came across the `--normalize-platforms` option for lock files
https://bundler.io/man/bundle-lock.1.html
Which does indicate that drift is a problem

Some more googling and I came across these discussions
https://github.com/orgs/rubygems/discussions/8645
https://github.com/orgs/rubygems/discussions/7487
Which seem to indicate that "invalid" platforms were once allowed and that there's an open conversation on how to move forward.
and eventually I started trying to read the code 
https://github.com/rubygems/rubygems/blob/master/lib/rubygems/platform.rb
Which is filled with regexes and made me think that the code is essentially `read something from the current platform and try to fit it in some known boxes`.

All in all I came to a reasonably strong belief that the platform identifier is a free string for the ruby world. One could run the experiment of trying to publish a gem with a manually configured platform of `foobarbizbaz` or something to see if the rubygems registry accepts it or not, but at least from a few hours of digging it seems like its just strings. If someone with more ruby knowledge would like to chime in I would love to be wrong here.

johnmhoran (approved):
LGTM -- thanks @darakian !

pombredanne:
@darakian Thanks! :bow:

PR #553, opened 2025-07-28
Replace `known` with `registered` for purl type lists

darakian:
Based on the comment here
https://github.com/package-url/purl-spec/pull/514#discussion_r2231757024
the word choice when listing purl types should be changed from known to registered.

When vers is broken out to its own repo we should probably do a similar change there.


pombredanne:
I am merging this... this is simple and does not require much more discussion.

PR #558, opened 2025-07-29
Improved readability of Conan spec

giterlizzi:
When Conan components are indicated in the `<tag>` like format, they are not rendered correctly because the Markdown parser mistakes them for HTML tags.


mjherzog (approved):
Looks good

PR #559, opened 2025-07-29
FIX some percent-encoding and qualifier order issue

giterlizzi:
I recently started implementing the new tests in URI-PackageURL, but after #514 I encountered some problems.

- "cocoapods", percent encoding issue
- "conan", qualifiers order issue (see #551)
- "generic", qualifiers order issue (see #551)
- "maven", qualifiers order issue (see #551)
- "mlflow", qualifiers order issue (see #551)
- "npm", percent encoding issue
- "oci", percent encoding and qualifiers order issues
- "rpm", qualifiers order issue (see #551)
- "swid", percent encoding issue

This PR is companion of #551 and fix some percent-encoding except "swid" because I have a doubts about whether `+` should be encoded as a space (`%20`) or as a plus sign (`%2B`) in namespace and name components.

In addition, I think some Conan tests should be removed or revised because they are contradictory (e.g. 'invalid conan purl only namespace’ and 'invalid conan purl only channel qualifier’) -- see last comment of #168

(edited)


matt-phylum:
Can you add a test to specification-test.json to ensure that + still decodes as +? Currently we're getting that testing through the cocoapods test and I agree that the cocoapods test shouldn't be doing that.

I'm pretty sure in the SWID tests the `+` is supposed to have been `%20`. Adobe has some documentation about their SWID tags and they don't use plus signs there. https://www.adobe.com/devnet-docs/acrobatetk/tools/AdminGuide/identify.html?highlight=swid

tests/types/npm-test.json
```diff
@@ -209,7 +209,7 @@
       "description": "valid npm purl without version and with subpath, scope unencoded",
       "test_group": "advanced",
       "test_type": "roundtrip",
-      "input": "pkg:npm/@babel/core#/googleapis/api/annotations/",
+      "input": "pkg:npm/%40babel/core#/googleapis/api/annotations/",
```
pombredanne: This is an advanced test, so the input can be non-canonical, therefore we probably do not want to change that
```suggestion
      "input": "pkg:npm/@babel/core#/googleapis/api/annotations/",
```


The same comment applies to a few other inputs change proposals.

giterlizzi: Reverted ;)

tests/types/npm-test.json
```diff
@@ -209,16 +209,16 @@
       "description": "valid npm purl without version and with subpath, scope unencoded",
       "test_group": "advanced",
       "test_type": "roundtrip",
-      "input": "pkg:npm/%40babel/core#/googleapis/api/annotations/",
-      "expected_output": "pkg:npm/%40babel/core#googleapis/api/annotations",
+      "input": "pkg:npm/@babel/core#/googleapis/api/annotations/",
+      "expected_output": "pkg:npm/@babel/core#googleapis/api/annotations",
```
jkowalleck: :x: this change is wrong.
per PURL spec, the `@` is not a valid character in a PURL - it must be encoded - as long as it is not the version separator

https://github.com/package-url/purl-spec/blob/0a884d224637ae429733b0e871d4e3e4eae7689d/PURL-SPECIFICATION.rst?plain=1#L232-L242


having an `@` at this position will cause the follwoing parse result:
- subpath: `googleapis/api/annotations`
- version: `babel/core`
- type:`npm`
- no name 
- no namespace

giterlizzi: That's right, `@babel/core` could be mistaken for the version.
(edit)

giterlizzi: I reverted the original test that included the `%20` character in the namespace for "expected_output".
https://github.com/package-url/purl-spec/blob/0a884d224637ae429733b0e871d4e3e4eae7689d/tests/types/npm-test.json#L213

I think the change should be made here too:
https://github.com/package-url/purl-spec/blob/0a884d224637ae429733b0e871d4e3e4eae7689d/tests/types/npm-test.json#L254

The alternative is set `‘expected_failure=true`, but there are some PURL implementations that support it, and it could cause problems in testing.

giterlizzi: Thanks ;)

jkowalleck: [...]  there are some PURL implementations that support it, and it could cause problems in testing.


we only care for the purl spec.
we dont care for "old"(pre PURL 1.0) implementations. they might be outdated, not up-to-date, or are waiting for the PURL spec to be finalized as 1.0

giterlizzi:
To keep the repo tidy in terms of duplicate issues and PRs, should this PR be closed in favour of #587 ?

jkowalleck:
> To keep the repo tidy in terms of duplicate issues and PRs, should this PR be closed in favour of #587 ?

#587 was merged.
are there any cases left?

giterlizzi:
I synced the JSON tests and no longer have sorting or percent-encoding errors in the qualifiers. However, there are still issues with some tests in other PURL types, but I will respond directly in the respective issues.

Thanks ;)

Issue #541, opened 2025-07-26
Move VERS spec files to their own repo


pombredanne:
As discussed in the community calls, there is a consensus to split the VERS spec for version ranges in its own repo.

Why?

- issues will be focused on either PURL or VERS
- repo tags will match exactly the version of one or the other 

Here is the list of TODOs:
- [x] Repo forked to vers-spec
- [x] Issues transferred
- [x] Discussions transferred
- [x] PRs merged
- [x] Repo synced, then unlinked from PURL
- [x] PURL content removed from VERS repo. https://github.com/package-url/vers-spec/pull/22
- [x] VERS repo README and root files updated
- [x] Close milestones in purl-spec and recreate in vers-spec
  - [x] https://github.com/package-url/purl-spec/milestone/7
  - [x] https://github.com/package-url/purl-spec/milestone/6
- [x] VERS content removed from PURL repo
- [x] VERS repo settings reviewed

mjherzog:
@pombredanne has created a new vers-spec repository as a fork of purl-spec.

This issue is to track the tasks to remove vers-spec files and data from the purl-spec repository after the new vers-spec repository is configured.

There are relatively few files in the current purl-spec repository that are related to the VERS spec. The primary file is: https://github.com/package-url/purl-spec/blob/main/VERSION-RANGE-SPEC.rst. 

As of July 28, 2025 the items with label: PURL/univers version scheme are:
- 14 Issues (10 open, 4 closed)
- 8 Pull requests (2 open, 6 closed)
We will eventually remove these issues from the purl-spec repo.

pombredanne:
We can transfer the issues, but not the PRs. I will merge the PRs, then sync up one last time the new vers repo before we can "unlink" it from the purl repo.

pombredanne:
See details at:
- https://github.com/package-url/vers-spec/issues/4

johnmhoran:
@mjherzog @pombredanne  I've marked `VERS content removed from PURL repo` as completed.  See https://github.com/package-url/purl-spec/pull/560.

mjherzog:
As documented in https://github.com/package-url/purl-spec/pull/560 we need to update VERSION-RANGE-SPEC.rst file to convert it to a redirect to https://github.com/package-url/vers-spec

mjherzog:
The last change that we need to complete this issue is to update the content of `purl-spec/VERSION-RANGE-SPEC.rst` to say:

[...]

PR #560, opened 2025-07-30
Update VERSION-RANGE-SPEC.rst #541

johnmhoran:
- Reference: https://github.com/package-url/purl-spec/issues/541


mjherzog:
My understanding now is that we need to replace the current content of VERSION-RANGE-SPEC.rst with a link to new vers-spec repo and a brief explanation of the change.

For reference:
- the CyclonexDX spec https://cyclonedx.org/schema/bom-1.6.xsd has 2 references to the CyclonexDX spec https://cyclonedx.org/schema/bom-1.6.xsd
- the CSAF spec https://docs.oasis-open.org/csaf/csaf/v2.0/os/csaf-v2.0-os.pdf has 1 reference in the description of VERS in Section 1.3 Normative References, but that reference is already a dead link because it is: https://github.com/package-url/purl-spec/blob/version-range-spec/VERSION-RANGE-SPEC.rst

johnmhoran:
Closing this in favor of 

- https://github.com/package-url/purl-spec/pull/613

PR #562, opened 2025-07-30
Add new PURL type: 'vcpkg'

### LINKED ISSUE #217: Add support for Microsoft's cross-platform C and C++ package manager, `vcpkg`

aristotelos:
- A continuation of #245  
- Closes #217
- It is also being tracked on the vcpkg side in microsoft/vcpkg#32732.

### DIFF
--- a/purl-types-index.json
+++ b/purl-types-index.json
@@ -32,5 +32,6 @@
   "qpkg",
   "rpm",
   "swid",
-  "swift"
+  "swift",
+  "vcpkg"
 ]
--- a/tests/types/vcpkg-test.json
+++ b/tests/types/vcpkg-test.json
@@ -1,0 +1,100 @@
+{
+  "$schema": "https://packageurl.org/schemas/purl-test.schema-1.0.json",
+  "tests": [
+    {
+      "description": "simple vcpkg purl",
+      "test_group": "base",
+      "test_type": "parse",
+      "input": "pkg:vcpkg/bzip2@1.0.8",
+      "expected_output": {
+        "type": "vcpkg",
+        "namespace": null,
+        "name": "bzip2",
+        "version": "1.0.8",
+        "qualifiers": null,
+        "subpath": null
+      },
+      "expected_failure": false,
+      "expected_failure_reason": null
+    },
+    {
+      "description": "vcpkg with port_version",
+      "test_group": "base",
+      "test_type": "parse",
+      "input": "pkg:vcpkg/bzip2@1.0.8?port_version=6",
+      "expected_output": {
+        "type": "vcpkg",
+        "namespace": null,
+        "name": "bzip",
+        "version": "1.0.8",
+        "qualifiers": {
+          "port_version": "6"
+        },
+        "subpath": null
+      },
+      "expected_failure": false,
+      "expected_failure_reason": null
+    },
+    {
+      "description": "vcpkg non-default git registry",
+      "test_group": "base",
+      "test_type": "parse",
+      "input": "pkg:vcpkg/ffmpeg@5.1.2?repository_url=https://github.com/azure-sdk/vcpkg&port_version=3",
+      "expected_output": {
+        "type": "vcpkg",
+        "namespace": null,
+        "name": "ffmpeg",
+        "version": "5.1.2",
+        "qualifiers": {
+          "repository_url": "https://github.com/azure-sdk/vcpkg",
+          "port_version": "3"
+        },
+        "subpath": null
+      },
+      "expected_failure": false,
+      "expected_failure_reason": null
+    },
+    {
+      "description": "vcpkg filesystem registry or overlay port",
+      "test_group": "base",
+      "test_type": "parse",
+      "input": "pkg:vcpkg/llvm@15.0.7?repository_url=file:///C:/local-registry/vcpkg&port_version=3",
+      "expected_output": {
+        "type": "vcpkg",
+        "namespace": null,
+        "name": "llvm",
+        "version": "15.0.7",
+        "qualifiers": {
+          "repository_url": "file:///C:/local-registry/vcpkg",
+          "port_version": "3"
+        },
+        "subpath": null
+      },
+      "expected_failure": false,
+      "expected_failure_reason": null
+    },
+    {
+      "description": "vcpkg with additional qualifiers",
+      "test_group": "base",
+      "test_type": "parse",
+      "input": "pkg:vcpkg/bzip2@1.0.8?port_version=6&repository_revision=41dfc53&os=Linux&os_arch=x64&build_type=Debug&linkage=dynamic",
+      "expected_output": {
+        "type": "vcpkg",
+        "namespace": null,
+        "name": "bzip2",
+        "version": "1.0.8",
+        "qualifiers": {
+          "port_version": "6",
+          "repository_revision": "41dfc53",
+          "os": "Linux",
+          "os_arch": "x64",
+          "build_type": "Debug",
+          "linkage": "dynamic"
+        },
+        "subpath": null
+      },
+      "expected_failure": false,
+      "expected_failure_reason": null
+    }
+  ]
+}
--- a/types/vcpkg-definition.json
+++ b/types/vcpkg-definition.json
@@ -1,0 +1,46 @@
+{
+  "$schema": "https://packageurl.org/schemas/purl-type-definition.schema-1.0.json",
+  "$id": "https://packageurl.org/types/vcpkg-definition.json",
+  "type": "vcpkg",
+  "type_name": "Vcpkg C/C++ packages",
+  "description": "Vcpkg C/C++ packages",
+  "repository": {
+    "use_repository": true,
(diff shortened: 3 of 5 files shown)


purl-types-index.json
mjherzog, line 1: The purl-types-index.json file is generated so it does not make sense to update it directly.

README-dev.rst
```diff
@@ -28,6 +28,6 @@ To validate that the schemas and data files are correct, run::
 To regenerate the Python utility model code from the JSON schemas, then regenerate the
 PURL type documentation from the JSON PURL type definition files, run::
 
-    make generate
-    make docs
+    make gencode
+    make gendocs
```
pombredanne: Thanks for all this! It would be best to avoid mixing code changes and the type definition contribution and have these in separate PRs for clarity and simplify the review ... no need to update that this time, but think about it for the future.

tests/types/vcpkg-test.json
```diff
@@ -0,0 +1,100 @@
...
+        "subpath": null
+      },
+      "expected_failure": false,
+      "expected_failure_reason": null
+    },
+    {
+      "description": "vcpkg with port_revision",
+      "test_group": "base",
+      "test_type": "parse",
+      "input": "pkg:vcpkg/ms-gsl@4.0.0?port_revision=4",
```
pombredanne: Where would the port_revision come from? I could not find anything in https://github.com/microsoft/vcpkg/tree/master/ports/ms-gsl

types/vcpkg-definition.json
```diff
@@ -0,0 +1,50 @@
...
+    "requirement": "required",
+    "native_name": "package-name",
+    "note": "The vcpkg package name."
+  },
+  "version_definition": {
+    "requirement": "required",
+    "native_name": "package-version",
+    "note": "The vcpg package version."
+  },
+  "subpath_definition": {
```
pombredanne: Can you elaborate why would you prohibit subpath? This is available to point inside a package, so I cannot fathom a use case where this would be prohibited.

types/vcpkg-definition.json
```diff
@@ -0,0 +1,50 @@
...
+    "requirement": "prohibited",
+    "note": "Currently reserved for future use and should be empty."
+  },
+  "qualifiers_definition": [
+    {
+      "key": "port_revision",
```
pombredanne: Is this really port_revision and not port_version?

types/vcpkg-definition.json
```diff
@@ -0,0 +1,50 @@
...
+      "description": "A string specifying the [port file revision](https://learn.microsoft.com/vcpkg/reference/vcpkg-json#port-version). If omitted, the purl refers to port file revision 0."
+    },
+    {
+      "key": "repository_revision",
+      "native_name": "repository_revision",
+      "requirement": "optional",
+      "description": "The commit hash of the vcpkg registry, potentially abbreviated. If omitted, the purl refers to the latest revision available."
+    }
+  ],
+  "note": "Additional qualifiers may be present on the purl that provide additional information about the context in which the package is being used, such as build configuration or platform information. These additional qualifiers must be tolerated during parsing and can be ignored if the parser does not expect them.",
```
pombredanne, line 39: Do you have an example of some of these?

types/vcpkg-definition.json
```diff
@@ -0,0 +1,50 @@
...
+  "type_name": "Vcpkg C/C++ packages",
+  "description": "Vcpkg C/C++ packages",
+  "repository": {
+    "use_repository": true,
+    "default_repository_url": "https://github.com/microsoft/vcpkg/",
+    "note": "The absolute URL of the vcpkg registry where the package is available (optional). If omitted, ``https://github.com/microsoft/vcpkg`` as default registry is assumed. For filesystem registries or [overlay ports](https://learn.microsoft.com/vcpkg/concepts/overlay-ports), the URI will have a `file` URI scheme."
+  },
+  "namespace_definition": {
+    "requirement": "prohibited",
+    "note": "Currently reserved for future use and should be empty."
```
pombredanne: I am not sure the note is useful... may be remove it? Reference to future use is uncommon.

types/vcpkg-definition.json
```diff
@@ -0,0 +1,50 @@
...
+  },
+  "name_definition": {
+    "requirement": "required",
+    "native_name": "package-name",
+    "note": "The vcpkg package name."
+  },
+  "version_definition": {
+    "requirement": "required",
+    "native_name": "package-version",
+    "note": "The vcpg package version."
```
pombredanne, line 23: There seems to be cases where the port-version is appended to the package-version as explained in https://github.com/azure-sdk/vcpkg
> Minimum required version. Port-version is identified with a #N suffix, for example, 1.2.3#7 names port-version 7.

Is this only the case sometimes, or often enough?
Could this mean that the port-version would be always part of the package version, and not a separate qualifier?

pombredanne (changes requested):
Thanks for all the efforts. I posted a few nit pickings for your kind consideration.

tests/types/vcpkg-test.json
aristotelos: The default is 0, see https://learn.microsoft.com/en-us/vcpkg/reference/vcpkg-json#port-version. It can be found in the port's `vcpkg.json`, see e.g. https://github.com/microsoft/vcpkg/blob/master/ports/bzip2/vcpkg.json#L4

types/vcpkg-definition.json
aristotelos: I think "port version" is maybe better, so I can update it to that. I was inspired by the [documentation](https://learn.microsoft.com/en-us/vcpkg/reference/vcpkg-json) that also states that the "port version" is the "port files revision".

types/vcpkg-definition.json
aristotelos: This is the standard notation indeed when the port-version is not 0. But as the `#` sign is not allowed in package version, isn't it better to keep it as a separate qualifier? Furthermore, for fuzzy matching of the package, the same version already counts as a match, the port-version could solve a security vulnerability but it may also not solve it. So I'd rather keep it like this because the standard `#` notation is not possible anyways.

types/vcpkg-definition.json
aristotelos: OK, will remove it then if it is not common

types/vcpkg-definition.json
aristotelos: It is not a feature of the vcpkg package manager to include just a part of a package, that was why I omitted it. But if you mean that purl is not about the capability of the vcpkg package manager but is just about being able to point to a part of a package, I'll revert that change

types/vcpkg-definition.json
aristotelos: Good point, I'll add one

tests/types/vcpkg-test.json
aristotelos: Updated the example to bzip2 to make it more realistic

types/vcpkg-definition.json
aristotelos: Reverted it for now

types/vcpkg-definition.json
aristotelos: Updated to "port_version"

types/vcpkg-definition.json
aristotelos: Added example

JamieMagee:
Hi @aristotelos, thanks for putting this together, and thanks @pombredanne for the review.

I work on supply chain security at Microsoft, and getting vcpkg into the purl spec would help us a lot, so I'd like to help land it. I've pushed a branch to my fork that builds on your commits.

Would you be OK with me taking this over and opening a fresh PR from my fork? I'm just as happy to do it the other way round. I can hand you the commits to cherry-pick so it stays your PR. Whatever works for you.

The one open item is @pombredanne's question about whether port_version should be a separate qualifier or folded into the version string. I can pick that up either way.

Branch: https://github.com/JamieMagee/purl-spec/tree/vcpkg-purl-fixes

JamieMagee:
@aristotelos I've opened #885: it's this PR rebased on current main and squashed to a single commit, with you credited as co-author. It clears the merge conflicts and the review comments here, and settles the `port_version` question (kept as a qualifier, since `#` is the PURL subpath delimiter, which also matches what [`microsoft/component-detection`][1] already emits).

[1]: https://github.com/microsoft/component-detection/blob/79782b812b982fbbc8fd2927059864e3df9ed261/src/Microsoft.ComponentDetection.Contracts/TypedComponent/VcpkgComponent.cs#L53-L63

JamieMagee:
Superseded by https://github.com/package-url/purl-spec/pull/885

johnmhoran:
Thank you @aristotelos.  As @JamieMagee noted above, this PR was superseded by https://github.com/package-url/purl-spec/pull/885, just merged.

PR #579, opened 2025-08-06
Add new PURL type: 'gitee'

### LINKED ISSUE #577: Add new PURL type: 'gitee'
## Motivation

[Gitee](https://gitee.com/) is the biggest code hosting platform in Asia with many original projects. We hope Gitee can be accepted as a type so these projects can be tracked with PURL.

## Changes

- Add a first draft for the description of the gitee type

PrinterFranklin:
- This PR will resolve #577 

- It is a new version of the closed PR #477

### DIFF
--- a/types/gitee-definition.json
+++ b/types/gitee-definition.json
@@ -1,0 +1,30 @@
+{
+  "$schema": "https://packageurl.org/schemas/purl-type-definition.schema-1.0.json",
+  "$id": "https://packageurl.org/types/gitee-definition.json",
+  "type": "gitee",
+  "type_name": "Gitee",
+  "description": "Gitee-based packages",
+  "repository": {
+    "use_repository": true,
+    "default_repository_url": "https://gitee.com"
+  },
+  "namespace_definition": {
+    "requirement": "required",
+    "case_sensitive": true,
+    "native_name": "user or organization",
+    "note": "The namespace is the user or organization. It is case sensitive."
+  },
+  "name_definition": {
+    "case_sensitive": true,
+    "native_name": "repository name",
+    "note": "The name is the repository name. It is case sensitive."
+  },
+  "version_definition": {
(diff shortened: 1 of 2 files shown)


types/gitee-definition.json
```diff
@@ -0,0 +1,30 @@
...
+  "type": "gitee",
+  "type_name": "Gitee",
+  "description": "Gitee-based packages",
+  "repository": {
+    "use_repository": true,
+    "default_repository_url": "https://gitee.com"
+  },
+  "namespace_definition": {
+    "requirement": "required",
+    "case_sensitive": false,
```
pombredanne: Is this case sensitive or not?  This flag conflicts with your note

types/gitee-definition.json
```diff
@@ -0,0 +1,30 @@
...
+    "native_name": "user or organization",
+    "note": "The namespace is the user or organization. It is case sensitive."
+  },
+  "name_definition": {
+    "case_sensitive": false,
```
pombredanne: Is this case sensitive or not?  This flag conflicts with your note

pombredanne (changes requested):
Thanks. Some feedback for you consideration! :heart:

types/gitee-definition.json
PrinterFranklin: Hi, it's case sensitive. I've fixed it. Thanks for your advice :-)

str4t0m: Sorry, I mixed it up with gitea, so you can ignore the comment.

~~Is it really case sensitive?
I am not familiar with the gitea code, but in a quick test with the generic package registry it didn't seem to care about the case. In my test I could publish/download the package no matter the case of the user/organization. Also, the name_definition/repository name case didn't seem to matter. It seems to store the package name with the casing from the first upload, but you can update or download it using different casing.~~

Issue #601, opened 2025-08-15
Extend data tracked for each package type and ecosystem with extra reference information


pombredanne:
There are a lot of data in many places about a package type and its ecosystem.
- icons like in: https://github.com/aboutcode-org/federatedcode/tree/main/federatedcode/static/pictogram-gh-pages originally from archived code at https://github.com/aboutcode-org/federatedcode/blob/main/federatedcode/static/pictogram-gh-pages.ABOUT 
- URL templates like in: https://github.com/package-url/packageurl-python/blob/main/src/packageurl/contrib/purl2url.py  or in https://github.com/librariesio/package-managers/blob/master/package-managers.json
- multiple attributes like in: https://docs.google.com/spreadsheets/d/1_CXfCRwHl33qK0Z47P-U-Old72Bwbh2jlE67SEQN2X4/edit?gid=0#gid=0
- many more details captured in: https://github.com/aboutcode-org/scancode-toolkit/ and in: https://github.com/aboutcode-org/fetchcode as well as https://github.com/CycloneDX/cdxgen/ and https://pypi.org/project/owasp-depscan/ and other tools

It would be awesome to capture that data for each PURL type definition so that we have this hard earned research and knowledge shared for everyone to reuse more easily.

@prabhu @andrew  @sjn

mjherzog:
Isn't this a job for PURLDB?

pombredanne:
@mjherzog re:

> Isn't this a job for PURLDB?

I think this is a job here. My title was misleading, this is really about each package "type" not each package. I am updating it.

For instance, documenting the pattern to query the PyPI JSON API when you know a PyPI PURL for a given package, and other interesting "meta" attributes for a package type and a whole ecosystem

mjherzog:
OK - That makes more sense. My hunch is that this will take significant design because it seems likely that this package metadata will be more divergent than the purl type data we capture today.

PR #603, opened 2025-08-16
definition improvements to make definitions ml-friendly

prabhu:
@pombredanne continuation of our slack discussion.

This quick draft is my proposal to make the type definitions more ML-friendly. This was vibe-coded by asking qwen3-coder (which scored 48.4% in logic and 90.34% in spec in [xBOMEval](https://github.com/CycloneDX/cdxgen/tree/master/contrib/xBOMEval/results)) for ideas to enhance the type definition to improve the learning and recall across all models (tiny to large).

The general principle behind any learning (human and machine) is that the key information needs to be summarised and repeated with both valid and invalid examples to reinforce learning and understanding.

Below is the response from qwen3-coder arguing the case for the enhancements as a machine.

-----------------

## Why Additional Machine-Learning Optimized Attributes Are Essential for PURL Specifications

The enhanced npm PURL JSON schema includes several attributes specifically designed to optimize understanding and reduce errors for both traditional parsers and modern AI/ML systems. Here's why each addition provides significant value:

### **1. `uri_compliance` - Explicit Encoding Rules for Deterministic Parsing**
Provides clear, unambiguous encoding requirements that eliminate guesswork for automated systems. ML models can reference these explicit rules rather than inferring encoding patterns from examples, reducing hallucination about URL-safe transformations.

### **2. `validation_rules` - Programmatic Compliance Checking**
Offers concrete validation patterns and examples that serve as training data for ML models. The explicit valid/invalid examples create clear decision boundaries, helping models learn the precise boundaries of acceptable formats.

### **3. `common_patterns` and `anti_patterns` - Positive and Negative Training Examples**
These provide essential contrastive learning signals. ML models learn not just what is correct, but crucially, what is incorrect. This dual approach significantly improves classification accuracy and reduces false positives.

### **4. `parsing_decision_tree` - Step-by-Step Logic for Complex Parsing**
Breaks down complex parsing decisions into simple, sequential steps that mirror how humans process PURLs. This structured approach helps ML models (especially smaller ones) handle complex nested conditions without losing context, reducing hallucination in multi-step reasoning.

### **5. `context` - Ecosystem Understanding for Semantic Accuracy**
Provides crucial background knowledge that helps ML models understand the "why" behind rules, not just the "what." This contextual understanding prevents models from making technically correct but semantically inappropriate suggestions (e.g., suggesting invalid npm practices that might be valid in other ecosystems).

### **6. `edge_cases` - Handling Real-World Complexity**
Documents nuanced scenarios that often cause confusion. ML models trained on these edge cases become more robust and less likely to confidently provide incorrect answers about boundary conditions, which are common sources of hallucination.

### **7. `llm_guidance` - Direct Instruction for Language Models**
Provides explicit meta-instructions that help LLMs avoid common pitfalls. The "do_not_assume" section is particularly valuable as it directly counters common hallucination patterns, while "key_facts_to_remember" reinforces critical constraints.

## **Business Case for ML-Optimized Specifications**

1. **Reduced Implementation Errors**: Clearer specifications lead to fewer bugs in parser implementations
2. **Better Tool Interoperability**: Explicit rules ensure consistent behavior across different tools and platforms
3. **Improved Developer Experience**: More intelligent autocomplete, validation, and error correction in IDEs
4. **Enhanced Security**: Better validation reduces the risk of injection attacks through malformed PURLs
5. **Future-Proofing**: Structured data is more easily consumed by emerging AI-assisted development tools

## **Recommendation**

These ML-optimized attributes transform the PURL specification from a human-readable document into a machine-actionable knowledge base. This approach benefits not just current AI systems but also traditional validation tools, documentation generators, and educational resources. The investment in structured, explicit specification pays dividends across the entire ecosystem of tools that consume PURL data.

### DIFF
--- a/types/npm-definition.json
+++ b/types/npm-definition.json
@@ -4,29 +4,205 @@
   "type": "npm",
   "type_name": "Node NPM packages",
   "description": "PURL type for npm packages.",
+  "uri_compliance": {
+    "namespace_encoding": "percent-encoded",
+    "name_encoding": "lowercase",
+    "version_encoding": "case-sensitive"
+  },
   "repository": {
     "use_repository": true,
     "default_repository_url": "https://registry.npmjs.org/",
+    "supported_repositories": [
+      "https://registry.npmjs.org/",
+      "https://npm.pkg.github.com/",
+      "https://registry.yarnpkg.com/"
+    ],
     "note": "The default repository is the npm Registry at https://registry.npmjs.org"
   },
   "namespace_definition": {
     "requirement": "optional",
     "case_sensitive": false,
+    "encoding": "percent-encoded",
     "native_name": "scope",
     "note": "The namespace is used for the scope of a scoped NPM package. The npm scope @ sign prefix is always percent encoded, as it was in the early days of npm scope."
   },
   "name_definition": {
+    "requirement": "required",
     "case_sensitive": false,
+    "encoding": "lowercase",
     "native_name": "name",
     "note": "Per the package.json spec, new package 'must not have uppercase letters in the name', therefore the name must be lowercased. The npm name used to be case sensitive in the early days for some old packages."
   },
   "version_definition": {
+    "requirement": "optional",
     "case_sensitive": true,
+    "encoding": "verbatim",
     "native_name": "version"
   },
+  "validation_rules": {
+    "namespace": {
+      "encoding_required": "%40",
+      "examples_valid": ["%40angular", "%40myorg"],
+      "examples_invalid": ["@angular", "%40Angular", "%40org name"]
+    },
+    "name": {
+      "min_length": 1,
+      "no_uppercase": true,
+      "url_safe_required": true,
+      "examples_valid": ["lodash", "react", "my-package_1"],
+      "examples_invalid": ["React", "my package", "my@package"]
+    }
+  },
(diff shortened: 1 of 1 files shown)


johnmhoran:
Is there an issue?

pombredanne:
@johnmhoran this would be https://github.com/package-url/purl-spec/issues/601 ?

types/npm-definition.json
```diff
@@ -4,29 +4,205 @@
...
+      "Including spaces without encoding",
+      "Misunderstanding scope vs name rules"
+    ]
+  },
   "examples": [
-    "pkg:npm/foobar@12.3.1",
-    "pkg:npm/%40angular/animation@12.3.1",
-    "pkg:npm/mypackage@12.4.5?vcs_url=git://host.com/path/to/repo.git%404345abcd34343"
-  ]
+    {
```
pombredanne, line 137: I guess these would be tests instead?

prabhu: Currently, it is a list. For a machine to learn, we need to explain why a given example is correct or incorrect using a description attribute. Otherwise, it will assume that only those tokens are correct.

prabhu:
i have found an alternative way to achieve this without the need for updating the types documentation. thank you for considering this proposal.

PR #613, opened 2025-08-20
Update the content of VERSION-RANGE-SPEC.rst #541

johnmhoran:
Reference: https://github.com/package-url/purl-spec/issues/541


mjherzog (approved):
Looks good.

mjherzog:
Bypassing rules because this is a simple change that can be refined later.

Issue #136, opened 2021-11-03
Letter case for npm packages


scanossmining:
Hello,

According to the specifications for the npm purl type, the package name inside the purl should be lowercased, however this would trigger some purl duplication. npmjs.com hosts case sensitive packages, some examples being:

- base64 and Base64
- dva, DVA, DvA, dVa

stevespringett:
By convention, npm is lowercase. The spec states the name SHOULD be lowercased, but it doesn't state that it MUST be lowercased. I do think we need clarity here though.

This issue was recently discovered in the official Package URL Java implementation and has subsequently been corrected. https://github.com/package-url/packageurl-java/pull/38 There may be other implementations also affected.

scanossmining:
Hi @stevespringett , thanks for answering this, it's clearer now. 
Before creating the issue, I've read the npm type specs here https://github.com/package-url/purl-spec/blob/master/PURL-TYPES.rst#npm ,where it says 

> Per the package.json spec, new package "must not have uppercase letters in the name", therefore the must be lowercased.

I guess it's a typo there and the ending was supposed to say "therefore the **package** must be lowercased" ?

pombredanne:
@scanossmining re:
> I guess it's a typo there and the ending was supposed to say "therefore the package must be lowercased" ?

Yes, good catch! PR welcomed!

iamwillbar:
From the [npm documentation](https://docs.npmjs.com/cli/v7/configuring-npm/package-json):

```
The name must be less than or equal to 214 characters. This includes the scope for scoped packages.
The names of scoped packages can begin with a dot or an underscore. This is not permitted without a scope.
New packages must not have uppercase letters in the name.
The name ends up being part of a URL, an argument on the command line, and a folder name. Therefore, the name can't contain any non-URL-safe characters.
```

The use of the word "New" on the third rule implies that this doesn't apply to existing packages. Since we want purls to be able to reference "new" and "old" packages that means we'd need to support mixed-case. The case should match the case used in the package's `package.json`.

wesleytodd:
> The use of the word "New" on the third rule implies that this doesn't apply to existing packages.

Sorry for reviving an old thread, but you are correct that there are older packages with names not following current rules.

[...]

PR #614, opened 2025-08-20
Add validation tests

TG1999:
Add tests for validating each PURL type.

- Fixes: https://github.com/aboutcode-org/purl-spec/issues/47

### DIFF
--- a/schemas/purl-test.schema.json
+++ b/schemas/purl-test.schema.json
@@ -1,7 +1,7 @@
 {
   "$schema": "http://json-schema.org/draft-07/schema#",
   "$id": "https://packageurl.org/schemas/purl-test.schema-1.0.json",
-  "title": "PURL test definition",
+  "title": "Package-URL test definition",
   "description": "Schema for Package-URL building and parsing tests with input and expected output.",
   "type": "object",
   "additionalProperties": false,
--- a/schemas/purl-types-index.schema.json
+++ b/schemas/purl-types-index.schema.json
@@ -1,8 +1,8 @@
 {
   "$schema": "http://json-schema.org/draft-07/schema#",
-  "$id": "https://purl-spec.org/schemas/purl-type-index.schema-1.0.json",
-  "title": "Package-URL types list.",
-  "description": "A list of the registered Package-URL types.",
+  "$id": "https://packageurl.org/schemas/purl-type-index.schema-1.0.json",
+  "title": "Package-URL types index",
+  "description": "An index of registered Package-URL types.",
   "type": "array",
   "additionalItems": false,
   "items": {


tests/types/mlflow-test.json
```diff
@@ -260,6 +260,104 @@
...
+        "namespace": null,
+        "name": "creditfraud",
+        "version": "3",
+        "qualifiers": {
+          "foo": "bar"
+        },
+        "subpath": null
+      },
+      "expected_messages": [
+        "Invalid qualifiers found: foo. Allowed qualifiers are: model_uuid, repository_url, run_id"
```
matt-phylum: I don't think it's right to reject unknown qualifiers. It provides little value while creating problems for extensibility and evolution.

tests/types/npm-test.json
```diff
@@ -254,6 +254,78 @@
...
+      "test_type": "validation",
+      "input": {
+        "type": "npm",
+        "namespace": null,
+        "name": "MyPackage",
+        "version": "1.0.0",
+        "qualifiers": null,
+        "subpath": null
+      },
+      "expected_messages": ["Name is not lowercased for purl type: 'npm'"]
```
matt-phylum: This is wrong. NPM does support mixed case names. PURL needs to stop adding new files saying that it doesn't.

tests/types/npm-test.json
```diff
@@ -254,6 +254,78 @@
...
+    },
+    {
+      "description": "Invalid npm package missing version",
```
matt-phylum: Why is this invalid?

schemas/purl-test.schema.json
```diff
@@ -284,6 +295,31 @@
...
+          "then": {
+            "properties": {
+              "input": {
+                "title": "Expected decoded PURL components",
+                "description": "Test input as an object with decoded PURL components for validation.",
+                "$ref": "#/definitions/purl_components"
+              },
+              "expected_messages": {
+                "title": "Expected messages",
+                "description": "Validation messages expected for this test. Can be null or a list of strings.",
```
matt-phylum: Are implementations really expected to produce exactly these text messages?

tests/types/pypi-test.json
```diff
@@ -222,6 +222,80 @@
...
+      "test_type": "validation",
+      "input": {
+        "type": "pypi",
+        "namespace": null,
+        "name": "Django-AllAuth",
+        "version": "12.23",
+        "qualifiers": null,
+        "subpath": null
+      },
+      "expected_messages": ["PyPI package name must be lowercase"]
```
matt-phylum: Is this invalid or is the implementation supposed to normalize the name as it is required to do for the previous test containing an underscore?

tests/types/mlflow-test.json
pombredanne: @matt-phylum agreed. @TG1999 we cannot reject unknown qualifiers, as a principle. Please update accordingly

tests/types/npm-test.json
pombredanne: Good point. The version is not required.

tests/types/npm-test.json
pombredanne: @matt-phylum You are correct that npm supported mixed case names in the past. But this is has long been not the case, like for the last 10+ years.

The current npm package.json spec is clear on this. Please see: https://docs.npmjs.com/cli/v11/configuring-npm/package-json

> New packages must not have uppercase letters in the name. 

See full details on that at:
- https://github.com/package-url/purl-spec/issues/136#issuecomment-3240215858

@TG1999 until this is sorted out, can you move this one validation test to a separate PR?

schemas/purl-test.schema.json
pombredanne: @matt-phylum re:
> Are implementations really expected to produce exactly these text messages?


good point. I have mixed feeling on that alright. One part of me says it would be really nice to have the same validation error messages everywhere. One part of me says it is not critical and may put too much burden on authors. What is your preference?

pombredanne (changes requested):
@TG1999 Thanks! see comments for your review.

tests/types/npm-test.json
pombredanne: @TG1999 actually, keep your validation test as-is for now, as this matches the spec for npm as it is. We need to handle separately whether to support mixed case in npms, and this is tracked in:
- https://github.com/package-url/purl-spec/issues/136#issuecomment-3240215858

pombredanne: @matt-phylum let's move that discussion to #136

matt-phylum: This test is wrong both ways. The current, incorrect spec says that NPM is not case sensitive and that the implementation must convert the name to lowercase, but this test ensures that the implementation throws a validation error instead. Are implementations supposed to allow creation of non-normalized PURLs so that they can do this validation? I think last time I checked there was maybe one implementation that supported non-normalized PURLs and could pass this test. IIRC all other implementations will immediately convert to lowercase as defined by the spec and fail this test.

tests/types/mlflow-test.json
TG1999: Done! I have updated tests for unknown qualifiers, the code for packageurl-python still gives error messages for unknown qualifiers, we may need to change that in future. And also structure validation errors in packageurl-python moving forward.

mjherzog:
@pombredanne @TG1999 We should have an issue to discuss and approve the change to purl-test-schema.json before we apply any other changes. I will draft an Issue based on the limited information in this PR.

pombredanne:
@TG1999 It would be great if you can rebase this PR?

schemas/purl-test.schema.json
```diff
@@ -104,12 +132,14 @@
...
-            "roundtrip"
+            "roundtrip",
+            "validation"
           ],
           "meta:enum": {
             "build": "A PURL building test from decoded components to a canonical PURL string.",
             "parse": "A PURL building test from decoded components to a canonical PURL string.",
-            "roundtrip": "A PURL roundtrip test, parsing then building back a PURL from a canonical string input."
+            "roundtrip": "A PURL roundtrip test, parsing then building back a PURL from a canonical string input.",
+            "validation": "A PURL validation test, checking if a PURL follows all the rules for a particular ecosystem."
```
mjherzog, line 142: The definition for the new validation test_type should be: ""A test to validate that an input PURL string complies with the rules for its PURL type."

mjherzog:
@TG1999 @pombredanne Can you provide a definition of the difference between info and warning messages?

PR #624, opened 2025-08-23
Fixes for apk

prabhu:
Fixes from gemini

*   **Corrected Metadata:** The `$id` attribute was updated from a placeholder value to the correct URI for the `apk` type definition.
*   **Required Components:** The `name_definition` and `version_definition` now correctly state that they are "required" components for a valid `apk` purl.
*   **Added Normalization Rules:** To enforce consistency, `normalization_rules` have been added to both the `namespace_definition` and `name_definition` to ensure their values are lowercased.
*   **Improved Version Description:** The note in the `version_definition` has been updated with more specific details about the expected version format, including the common `pkgver-rX` pattern.
*   **Expanded Qualifiers:** The schema now includes definitions for the `distro` and `repository_url` qualifiers, and examples have been added for all defined qualifiers to provide clearer guidance.
*   **Enhanced Examples:** The list of examples has been expanded to better illustrate the variety of valid `apk` purls, including packages from different vendors and those with version suffixes.

### DIFF
--- a/types/apk-definition.json
+++ b/types/apk-definition.json
@@ -35,32 +35,17 @@
     {
       "key": "arch",
       "requirement": "optional",
-      "description": "The arch is the qualifiers key for a package architecture.",
-      "examples": [
-        "x86",
-        "x86_64",
-        "aarch64",
-        "armhf",
-        "armv7",
-        "noarch"
-      ]
+      "description": "The arch is the qualifiers key for a package architecture."
     },
     {
       "key": "distro",
       "requirement": "optional",
-      "description": "The distribution name when using multiple distributions",
-      "examples": [
-        "alpine",
-        "postmarketos"
-      ]
+      "description": "The distribution name when using multiple distributions"
     },
     {
       "key": "repository_url",
       "requirement": "optional",
-      "description": "Base URL for the package repository",
-      "examples": [
-        "https://dl-cdn.alpinelinux.org/alpine/"
-      ]
+      "description": "Base URL for the package repository"
     }
   ],
   "examples": [


types/apk-definition.json
```diff
@@ -1,37 +1,59 @@
...
   },
   "name_definition": {
+    "requirement": "required",
     "note": "The name is the package name. It is not case sensitive and must be lowercased.",
     "native_name": "name",
-    "case_sensitive": false
+    "case_sensitive": false,
+    "normalization_rules": [
+      "It is not case sensitive and must be lowercased."
+    ]
```
matt-phylum, line 27:
```suggestion
    "case_sensitive": true
```

```/ # apk add FIGLET
ERROR: unable to select packages:
  FIGLET (no such package):
    required by: world[FIGLET]
/ # apk add figlet
(1/1) Installing figlet (2.2.5-r3)
Executing busybox-1.37.0-r18.trigger
OK: 8 MiB in 17 packages
```

types/apk-definition.json
```diff
@@ -1,37 +1,59 @@
...
   },
   "version_definition": {
-    "note": "The version is a package version as expected by apk.",
+    "requirement": "required",
```
matt-phylum, line 30:
```suggestion
```

Installed packages have versions, but I don't understand why the PURL spec should be limited to talking about only specific versions of installed packages, especially if vers is going to be a related standard. How would you use vers if you can't name the package?

types/apk-definition.json
```diff
@@ -1,37 +1,59 @@
...
     {
       "key": "arch",
+      "requirement": "optional",
       "description": "The arch is the qualifiers key for a package architecture."
+    },
+    {
+      "key": "distro",
+      "requirement": "optional",
+      "description": "The distribution name when using multiple distributions"
+    },
```
matt-phylum, line 44: This doesn't make sense. I don't know if the problem is just that it's not defined properly. The distribution goes in the namespace. What does it mean if a PURL has a distribution in the namespace and in the qualifier? If it makes sense for there two be two distributions, what if there are more than two?

PR #627, opened 2025-08-23
Fixes for bitbucket

prabhu:
Fixes from gemini

- Corrected Typos: The non-standard attribute `is_case_sensitve` has been corrected to the standard `case_sensitive` in both the namespace_definition and name_definition.
- Required Components: The name_definition and version_definition now correctly state that they are "required" components.
- Added Normalization Rules: To enforce the lowercasing requirement, normalization_rules have been added to the namespace_definition and name_definition.
- Improved Descriptions: The native_name and note for the namespace have been updated to clarify its role for both Bitbucket Cloud (user/organization) and Server (project key). The version_definition note now explicitly warns against using mutable branch names.
- Added Subpath and Qualifiers: The schema now includes a subpath_definition to support monorepos and a qualifiers_definition for the repository_url key, which is essential for self-hosted instances.
- Enhanced Examples: The list of examples has been significantly expanded to cover various use cases, including packages identified by tags, those within a monorepo, and packages from self-hosted Bitbucket servers.

### DIFF
--- a/types/bitbucket-definition.json
+++ b/types/bitbucket-definition.json
@@ -10,20 +10,42 @@
   },
   "namespace_definition": {
     "requirement": "required",
-    "note": "The namespace is the user or organization. It is not case sensitive and must be lowercased.",
-    "native_name": "user or organization",
-    "is_case_sensitve": false
+    "note": "The namespace is the user or organization (for Bitbucket Cloud), or the Project Key (for Bitbucket Server/Data Center). It is not case sensitive and must be lowercased.",
+    "native_name": "user, organization, or project key",
+    "case_sensitive": false,
+    "normalization_rules": [
+      "It is not case sensitive and must be lowercased."
+    ]
   },
   "name_definition": {
+    "requirement": "required",
     "note": "The name is the repository name. It is not case sensitive and must be lowercased.",
     "native_name": "repository name",
-    "is_case_sensitve": false
+    "case_sensitive": false,
+    "normalization_rules": [
+      "It is not case sensitive and must be lowercased."
+    ]
   },
   "version_definition": {
-    "note": "The version is a commit or tag.",
+    "requirement": "required",
+    "note": "The version is an immutable commit hash (full or short) or a tag. Branch names (e.g., 'main') should not be used as they are mutable pointers.",
     "native_name": "commit or tag"
   },
+  "subpath_definition": {
+    "requirement": "optional",
+    "note": "The subpath is used to specify a subdirectory within the repository, which is essential for monorepos."
+  },
+  "qualifiers_definition": [
+    {
+      "key": "repository_url",
+      "requirement": "optional",
+      "description": "Custom repository URL for self-hosted Bitbucket Server/Data Center instances."
+    }
+  ],
   "examples": [
-    "pkg:bitbucket/birkenfeld/pygments-main@244fd47e07d1014f0aed9c"
+    "pkg:bitbucket/birkenfeld/pygments-main@244fd47e07d1014f0aed9c",
+    "pkg:bitbucket/atlassian/atlassian-frontend@v104.1.0",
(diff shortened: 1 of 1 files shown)


types/bitbucket-definition.json
```diff
@@ -10,20 +10,42 @@
...
     "native_name": "repository name",
-    "is_case_sensitve": false
+    "case_sensitive": false,
+    "normalization_rules": [
+      "It is not case sensitive and must be lowercased."
+    ]
   },
   "version_definition": {
-    "note": "The version is a commit or tag.",
+    "requirement": "required",
```
matt-phylum, line 30: It is unnecessarily restricting the usefulness of PURL to require a version here.

types/bitbucket-definition.json
```diff
@@ -10,20 +10,42 @@
...
+    "note": "The version is an immutable commit hash (full or short) or a tag. Branch names (e.g., 'main') should not be used as they are mutable pointers.",
```
matt-phylum, line 31: This doesn't work. Tags are also mutable. Software that works with PURLs cannot reasonably knowing whether a reference is intended to be mutable.

types/bitbucket-definition.json
```diff
@@ -10,20 +10,42 @@
...
+      "key": "repository_url",
+      "requirement": "optional",
+      "description": "Custom repository URL for self-hosted Bitbucket Server/Data Center instances."
+    }
+  ],
   "examples": [
-    "pkg:bitbucket/birkenfeld/pygments-main@244fd47e07d1014f0aed9c"
+    "pkg:bitbucket/birkenfeld/pygments-main@244fd47e07d1014f0aed9c",
+    "pkg:bitbucket/atlassian/atlassian-frontend@v104.1.0",
+    "pkg:bitbucket/atlassian/atlassian-frontend@v104.1.0#packages/analytics",
```
matt-phylum, line 48: None of these examples exist? Where did these new examples come from? The Pygments one might date back to when Bitbucket still supported Mercurial.

types/bitbucket-definition.json
pombredanne: Agreed, @prabhu let's not make versions required

types/bitbucket-definition.json
```diff
@@ -10,20 +10,42 @@
...
+    ]
   },
   "name_definition": {
+    "requirement": "required",
     "note": "The name is the repository name. It is not case sensitive and must be lowercased.",
     "native_name": "repository name",
-    "is_case_sensitve": false
+    "case_sensitive": false,
+    "normalization_rules": [
+      "It is not case sensitive and must be lowercased."
```
pombredanne, line 26: This is redundant with `case_sensitive`, so we would not add this IMHO

types/bitbucket-definition.json
```diff
@@ -10,20 +10,42 @@
...
   "namespace_definition": {
     "requirement": "required",
-    "note": "The namespace is the user or organization. It is not case sensitive and must be lowercased.",
-    "native_name": "user or organization",
-    "is_case_sensitve": false
+    "note": "The namespace is the user or organization (for Bitbucket Cloud), or the Project Key (for Bitbucket Server/Data Center). It is not case sensitive and must be lowercased.",
+    "native_name": "user, organization, or project key",
+    "case_sensitive": false,
+    "normalization_rules": [
+      "It is not case sensitive and must be lowercased."
```
pombredanne, line 17: case_sensitive implies this alright

types/bitbucket-definition.json
pombredanne: Makes sense. The original wording is simple enough and does not specify too little nor too much IMHO

types/bitbucket-definition.json
pombredanne: Good point, the pygments example should go since the project long moved to GitHub from Mercurial. ... for other examples too, let's use something real

maennchen:

PR #629, opened 2025-08-23
Fixes for cargo

prabhu:
- Corrected Typo: The non-standard attribute is_case_sensitive in name_definition has been corrected to the standard case_sensitive.
- Defined Requirements: Both the name_definition and version_definition are now correctly marked as "required", reflecting that they are mandatory components for a valid cargo purl.
- Improved Descriptions: The note fields for the name and version have been updated to be more specific. The name is clarified as the "package name," and the version note now specifies that it should follow the Semantic Versioning (SemVer) specification.
- Added Qualifiers Definition: A qualifiers_definition has been added to define the optional repository_url key, which is used to specify alternative registries for packages not hosted on crates.io.

### DIFF
--- a/types/cargo-definition.json
+++ b/types/cargo-definition.json
@@ -10,17 +10,26 @@
   },
   "namespace_definition": {
     "requirement": "prohibited",
-    "note": "there is no namespace"
+    "note": "There is no namespace for Cargo packages."
   },
   "name_definition": {
+    "requirement": "required",
     "native_name": "name",
-    "is_case_sensitve": true,
-    "note": "The name is the repository name."
+    "case_sensitive": true,
+    "note": "The name is the package name. It is case-sensitive."
   },
   "version_definition": {
+    "requirement": "required",
     "native_name": "version",
-    "note": "The version is the package version."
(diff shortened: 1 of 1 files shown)


types/cargo-definition.json
```diff
@@ -10,20 +10,29 @@
...
   "name_definition": {
+    "requirement": "required",
     "native_name": "name",
-    "is_case_sensitve": true,
-    "note": "The name is the repository name."
+    "case_sensitive": true,
+    "note": "The name is the package name. It is case-sensitive."
   },
   "version_definition": {
+    "requirement": "required",
```
matt-phylum, line 22:
```suggestion
```

Why would the version be required for Cargo specifically if it is not required for other, similar package managers?

types/cargo-definition.json
```diff
@@ -10,20 +10,29 @@
...
     "native_name": "version",
-    "note": "The version is the package version."
+    "note": "The version is the package version. It should follow the Semantic Versioning (SemVer) specification."
```
matt-phylum, line 24: What does it mean for PURL implementations that the version "should" be semver? Either it is semver or it is not semver. I don't know if it matters. This seems like unnecessary information.

PR #633, opened 2025-08-23
Fixes for conda

prabhu:
- Defined Requirements: The name_definition and version_definition are now correctly marked as "required", as they are mandatory for a valid conda purl.
- Clarified Qualifiers: The requirement for all keys in the qualifiers_definition (build, channel, subdir, and type) has been explicitly set to "optional".
- Improved Descriptions: The note for the namespace_definition has been updated for better clarity.
- Expanded Examples: The list of examples has been expanded to include more variations, providing a better illustration of how conda purls can be constructed.

### DIFF
--- a/types/conda-definition.json
+++ b/types/conda-definition.json
@@ -9,36 +9,44 @@
     "default_repository_url": "https://repo.anaconda.com"
   },
   "namespace_definition": {
-    "note": "there is no namspace",
+    "note": "There is no namespace for Conda packages.",
     "requirement": "prohibited"
   },
   "name_definition": {
+    "requirement": "required",
     "native_name": "name",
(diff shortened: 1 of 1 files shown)


types/conda-definition.json
```diff
@@ -9,36 +9,44 @@
     "default_repository_url": "https://repo.anaconda.com"
   },
   "namespace_definition": {
-    "note": "there is no namspace",
+    "note": "There is no namespace for Conda packages.",
     "requirement": "prohibited"
   },
   "name_definition": {
+    "requirement": "required",
```
matt-phylum, line 16: What would it mean for a package to not have a name? What meaning could that PURL provide?

PR #634, opened 2025-08-23
Fixes for cpan

prabhu:
CPAN namespace rules are too difficult to represent in JSON Schema format. Gemini and I tried our best to convey the information from the extended [definition](https://github.com/AppThreat/purl4ml/blob/main/types/cpan-extended.json) in the current format.

- Defined Requirements: The name_definition is now correctly marked as "required", while the version_definition is marked as "optional". The requirement for all keys in the qualifiers_definition has also been explicitly set to "optional".
- Improved Namespace and Name Definitions: The namespace_definition now specifies its native_name as "author id" and includes a normalization_rules field to enforce that it must be uppercased. The note in name_definition has been updated to be more explicit about the different validation rules that apply depending on whether a namespace is present (for a distribution) or absent (for a module).
- Added Case Sensitivity: The namespace_definition now correctly includes case_sensitive: true.


giterlizzi:
Hi, this CPAN specification will be replaced by this PR #420.

prabhu:
Let's keep this open. Once 420 is merged will rebase and redo to update the models

sjn:
Seems I made a mistake in the merge; Apologies!

PR #635, opened 2025-08-23
Fixes for cran

prabhu:
- Corrected Typo: The non-standard attribute is_case_sensitve in name_definition has been corrected to the standard case_sensitive.
- Defined Requirements: The name_definition and version_definition are now correctly marked as "required", as they are mandatory for a valid cran purl.
- Improved Name Definition: The note in the name_definition has been updated to be more specific, clarifying that while the name is case-sensitive, the canonical case spelling MUST be used because CRAN enforces case-insensitive uniqueness.
- Added Qualifiers Definition: An empty qualifiers_definition has been added to the schema to explicitly state that the cran purl type does not use any qualifiers.

### DIFF
--- a/types/cran-definition.json
+++ b/types/cran-definition.json
@@ -13,14 +13,17 @@
     "note": "there is no namespace"
   },
   "name_definition": {
+    "requirement": "required",
     "native_name": "name",
-    "is_case_sensitve": true,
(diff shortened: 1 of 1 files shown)


maennchen:

PR #636, opened 2025-08-23
Fixes for deb

prabhu:
- Corrected Typos: The non-standard attribute is_case_sensitve has been corrected to the standard case_sensitive in both the namespace_definition and name_definition.
- Defined Requirements: The name_definition and version_definition are now correctly marked as "required", as they are mandatory components for a valid deb purl.
- Added Normalization Rules: To programmatically enforce the lowercasing requirement, normalization_rules have been added to both the namespace_definition and name_definition.
- Expanded Qualifiers: The qualifiers_definition has been updated to include the distro and repository_url keys. The requirement for all qualifiers has been explicitly set to "optional".

### DIFF
--- a/types/deb-definition.json
+++ b/types/deb-definition.json
@@ -10,23 +10,42 @@
   },
   "namespace_definition": {
     "native_name": "vendor",
-    "is_case_sensitve": false,
+    "case_sensitive": false,
     "note": "The namespace is the \"vendor\" name such as \"debian\" or \"ubuntu\". It is not case sensitive and must be lowercased.",
-    "requirement": "required"
+    "requirement": "required",
+    "normalization_rules": [
+      "It is not case sensitive and must be lowercased."
+    ]
   },
   "name_definition": {
+    "requirement": "required",
     "native_name": "name",
-    "is_case_sensitve": false,
-    "note": "The name is not case sensitive and must be lowercased."
+    "case_sensitive": false,
+    "note": "The name is not case sensitive and must be lowercased.",
+    "normalization_rules": [
+      "It is not case sensitive and must be lowercased."
+    ]
   },
   "version_definition": {
+    "requirement": "required",
     "native_name": "version",
     "note": "The version is the version of the binary (or source) package."
   },
   "qualifiers_definition": [
     {
       "key": "arch",
+      "requirement": "optional",
       "description": "arch is the qualifiers key for a package architecture. The special value arch=source identifies a Debian source package that usually consists of a Debian Source control file (.dsc) and corresponding upstream and Debian sources. The dpkg-query command can print the name and version of the corresponding source package of a binary package, e.g. dpkg-query -f ${source:Package} ${source:Version} -W <binary package name>"
+    },
+    {
+      "key": "distro",
+      "requirement": "optional",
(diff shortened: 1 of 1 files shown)


types/deb-definition.json
```diff
@@ -10,23 +10,42 @@
...
   "name_definition": {
+    "requirement": "required",
     "native_name": "name",
-    "is_case_sensitve": false,
-    "note": "The name is not case sensitive and must be lowercased."
+    "case_sensitive": false,
+    "note": "The name is not case sensitive and must be lowercased.",
+    "normalization_rules": [
+      "It is not case sensitive and must be lowercased."
+    ]
```
matt-phylum, line 27:
```suggestion
    "case_sensitive": true
```

The name is case sensitive and must not be lowercased.

```
root@96d7edba8206:/# apt install COWSAY
Error: Unable to locate package COWSAY
root@96d7edba8206:/# apt install cowsay
Installing:
  cowsay

Installing dependencies:
  libgdbm-compat4t64  libgdbm6t64  libperl5.40  libtext-charwidth-perl  netbase  perl  perl-modules-5.40

Suggested packages:
  filters     gdbm-l10n       perl-doc                   | libterm-readline-perl-perl  libtap-harness-archive-perl
  cowsay-off  sensible-utils  libterm-readline-gnu-perl  make

Summary:
  Upgrading: 0, Installing: 8, Removing: 0, Not Upgrading: 0
  Download size: 7596 kB
  Space needed: 53.6 MB / 68.9 GB available

Continue? [Y/n]
```

types/deb-definition.json
```diff
@@ -10,23 +10,42 @@
...
     {
       "key": "arch",
+      "requirement": "optional",
       "description": "arch is the qualifiers key for a package architecture. The special value arch=source identifies a Debian source package that usually consists of a Debian Source control file (.dsc) and corresponding upstream and Debian sources. The dpkg-query command can print the name and version of the corresponding source package of a binary package, e.g. dpkg-query -f ${source:Package} ${source:Version} -W <binary package name>"
+    },
+    {
+      "key": "distro",
+      "requirement": "optional",
+      "description": "The distribution name, codename, or suite."
+    },
```
matt-phylum, line 44: The distribution name is already in the namespace. If there is going to be a distribution in the qualifier there needs to be a clear explanation of how these are different and what it means when there is or is not a value here.

types/deb-definition.json
```diff
@@ -10,23 +10,42 @@
...
+    {
+      "key": "repository_url",
+      "requirement": "optional",
+      "description": "Base URL for the package repository."
```
matt-phylum, line 48: If apt is going to have a repository URL description then it should have more definition than this because the way packages are arranged is more complicated than in other package types. If a package is in `http://deb.debian.org/debian-security` suite `trixie-security`, what is the value for `repository_url`? Where does the `trixie-security` part go? Is that what's supposed to go in `distro`? If that's the `distro` value then either it needs very clear disambiguation or it needs to be renamed to not be `distro` so people don't put "debian" or "ubuntu" (what people typically think of as a Linux distro) there.

maennchen:

PR #637, opened 2025-08-23
Fixes for docker

prabhu:
- Defined Requirements: The name_definition is now correctly marked as "required", and the version_definition is marked as "optional".
- Added Case Sensitivity and Normalization: Both the namespace_definition and name_definition now include case_sensitive: false and a normalization_rules field to enforce that their values must be lowercased.
- Improved Descriptions: The native_name for name_definition and version_definition has been updated for better clarity.
- Added Qualifiers Definition: A qualifiers_definition has been added to define the optional keys repository_url, arch, and os, which are used for specifying custom registries and multi-platform images.

NOTE: Is it possible we are misusing the `repository_url` qualifier to represent registry. What if the repository and registry are different urls?

### DIFF
--- a/types/docker-definition.json
+++ b/types/docker-definition.json
@@ -9,15 +9,45 @@
     "default_repository_url": "https://hub.docker.com"
   },
   "namespace_definition": {
+    "requirement": "optional",
     "note": "The namespace is the registry/user/organization if present.",
-    "requirement": "optional"
+    "native_name": "registry/user/organization",
+    "case_sensitive": false,
+    "normalization_rules": [
+      "It is not case sensitive and must be lowercased."
+    ]
   },
   "name_definition": {
-    "native_name": "name"
+    "requirement": "required",
+    "native_name": "image name",
+    "note": "The name is the Docker image name.",
+    "case_sensitive": false,
+    "normalization_rules": [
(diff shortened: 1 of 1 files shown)


types/docker-definition.json
```diff
@@ -9,18 +9,48 @@
     "default_repository_url": "https://hub.docker.com"
   },
   "namespace_definition": {
+    "requirement": "optional",
     "note": "The namespace is the registry/user/organization if present.",
-    "requirement": "optional"
+    "native_name": "registry/user/organization",
```
matt-phylum, line 14: When hosted on Docker Hub, this is a user or organization, but hosted anywhere else you can't assume that level of meaning without knowing about where the image is hosted.

types/docker-definition.json
```diff
@@ -9,18 +9,48 @@
...
+    "case_sensitive": false,
+    "normalization_rules": [
+      "It is not case sensitive and must be lowercased."
+    ]
```
matt-phylum, line 18: I don't think the namespace or name is case insensitive. The spec says that the name is lowercase, and at least some implementations consider uppercase to be invalid. Uppercase is not normalized to lowercase by native OCI tools I've tried.

PR #638, opened 2025-08-24
Adds more test cases for npm

prabhu:
These additional tests were vibe-coded using [cdx1-pro](https://huggingface.co/collections/CycloneDX/cdx1-pro-688e15a3c3b593753ceefc05) trained with the new enhanced purl schema definitions(https://github.com/AppThreat/purl4ml/blob/main/types/npm-extended.json). The tests were manually reviewed, enhanced, and re-tested with the `make check` command.

## Methodology

Technically, any frontier model or LLMs tuned with the new purl4ml [repo](https://github.com/AppThreat/purl4ml) can be used to generate decent-quality additional tests for PURL types.

### Prompt

> you are an AI tasked with improving the test coverage for PURL. For each purl type, I will upload an enhanced purl schema with detailed information and then the current tests in json format. Carefully understand the enhanced schema and share the updated test json by adding additional tests both positive and negative including edge cases. acknowledge that you understand this request so that I can start sharing the json schema and tests in this chat.

### Screenshots

Screenshot 2025-08-24 at 10 29 30
Screenshot 2025-08-24 at 10 31 42

### DIFF
--- a/tests/types/npm-test.json
+++ b/tests/types/npm-test.json
@@ -254,6 +254,223 @@
       "expected_output": "pkg:npm/core@2.0.1#googleapis/api/annotations",
       "expected_failure": false,
       "expected_failure_reason": null
+    },
+    {
+      "description": "Negative Test: An npm purl with an un-encoded @ in scope should fail parsing",
+      "test_group": "advanced",
+      "test_type": "parse",
+      "input": "pkg:npm/@angular/animation@12.3.1",
+      "expected_output": null,
+      "expected_failure": true,
+      "expected_failure_reason": "Should fail to parse due to unencoded '@' in the namespace"
+    },
+    {
+      "description": "Negative Test: An npm purl with an uppercase name should fail parsing",
+      "test_group": "advanced",
+      "test_type": "parse",
+      "input": "pkg:npm/React@17.0.1",
+      "expected_output": null,
+      "expected_failure": true,
+      "expected_failure_reason": "Should fail to parse due to uppercase characters in the name"
+    },
+    {
+      "description": "Normalization Test: An npm purl with an uppercase name should be lowercased on roundtrip",
+      "test_group": "advanced",
+      "test_type": "roundtrip",
+      "input": "pkg:npm/LoDash@4.17.21",
+      "expected_output": "pkg:npm/lodash@4.17.21",
+      "expected_failure": false,
+      "expected_failure_reason": null
+    },
+    {
+      "description": "Positive Test: An npm purl with no version is valid",
+      "test_group": "advanced",
+      "test_type": "parse",
+      "input": "pkg:npm/react",
+      "expected_output": {
+        "type": "npm",
+        "namespace": null,
+        "name": "react",
+        "version": null,
+        "qualifiers": null,
+        "subpath": null
+      },
+      "expected_failure": false,
+      "expected_failure_reason": null
+    },
+    {
+      "description": "Positive Test: A scoped npm purl with no version is valid",
+      "test_group": "advanced",
+      "test_type": "parse",
+      "input": "pkg:npm/%40angular/core",
+      "expected_output": {
+        "type": "npm",
+        "namespace": "@angular",
+        "name": "core",
+        "version": null,
+        "qualifiers": null,
+        "subpath": null
+      },
+      "expected_failure": false,
+      "expected_failure_reason": null
+    },
+    {
+      "description": "Negative Test: An npm purl with an empty version should fail",
+      "test_group": "advanced",
+      "test_type": "parse",
+      "input": "pkg:npm/foobar@",
+      "expected_output": null,
+      "expected_failure": true,
+      "expected_failure_reason": "Should fail to parse due to an empty version string"
+    },
+    {
+      "description": "Negative Test: An npm purl with a name starting with a dot should fail",
+      "test_group": "advanced",
+      "test_type": "parse",
+      "input": "pkg:npm/.foobar@1.0.0",
+      "expected_output": null,
+      "expected_failure": true,
+      "expected_failure_reason": "Should fail to parse because the name starts with a dot"
+    },
+    {
+      "description": "Negative Test: An npm purl with a name starting with an underscore should fail",
+      "test_group": "advanced",
+      "test_type": "parse",
+      "input": "pkg:npm/_foobar@1.0.0",
+      "expected_output": null,
+      "expected_failure": true,
+      "expected_failure_reason": "Should fail to parse because the name starts with an underscore"
+    },
+    {
+      "description": "Edge Case Test: Version with pre-release and build metadata should be preserved",
+      "test_group": "advanced",
+      "test_type": "roundtrip",
+      "input": "pkg:npm/foobar@1.2.3-alpha.1+build.456",
+      "expected_output": "pkg:npm/foobar@1.2.3-alpha.1+build.456",
+      "expected_failure": false,
+      "expected_failure_reason": null
+    },
+    {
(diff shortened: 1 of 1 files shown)


tests/types/npm-test.json
```diff
@@ -254,6 +254,316 @@
...
       "expected_failure": false,
       "expected_failure_reason": null
+    },
+    {
+      "description": "Negative Test: An npm purl with an un-encoded @ in scope should fail parsing",
+      "test_group": "advanced",
+      "test_type": "parse",
+      "input": "pkg:npm/@angular/animation@12.3.1",
+      "expected_output": null,
+      "expected_failure": true,
```
matt-phylum, line 264: Is this true? Last I heard it was expected to parse because there was no instruction to specifically reject unencoded `@` characters in positions where they _could_ require encoding depending on the rest of the PURL.

tests/types/npm-test.json
```diff
@@ -254,6 +254,316 @@
...
+    },
+    {
+      "description": "Negative Test: An npm purl with an uppercase name should fail parsing",
+      "test_group": "advanced",
+      "test_type": "parse",
+      "input": "pkg:npm/React@17.0.1",
+      "expected_output": null,
+      "expected_failure": true,
+      "expected_failure_reason": "Should fail to parse due to uppercase characters in the name"
+    },
```
matt-phylum, line 275: This is a valid PURL and _must_ parse correctly as name "React" with an uppercase R.

tests/types/npm-test.json
```diff
@@ -254,6 +254,316 @@
...
+    {
+      "description": "Normalization Test: An npm purl with an uppercase name should be lowercased on roundtrip",
+      "test_group": "advanced",
+      "test_type": "roundtrip",
+      "input": "pkg:npm/LoDash@4.17.21",
+      "expected_output": "pkg:npm/lodash@4.17.21",
+      "expected_failure": false,
+      "expected_failure_reason": null
+    },
```
matt-phylum, line 284: This test is wrong, and also conflicts with the previous test. The correct output is `pkg:npm/LoDash`.

tests/types/npm-test.json
```diff
@@ -254,6 +254,316 @@
...
+    {
+      "description": "Negative Test: An npm purl with a name starting with an underscore should fail",
+      "test_group": "advanced",
+      "test_type": "parse",
+      "input": "pkg:npm/_foobar@1.0.0",
+      "expected_output": null,
+      "expected_failure": true,
+      "expected_failure_reason": "Should fail to parse because the name starts with an underscore"
+    },
```
matt-phylum, line 343: Why should these fail? PURL does not (and IMO should not) define any rules to forbid this.

tests/types/npm-test.json
```diff
@@ -254,6 +254,316 @@
...
+    {
+      "description": "Negative Test: An npm purl with a unicode character (emoji) in the scope should fail parsing",
+      "test_group": "advanced",
+      "test_type": "parse",
+      "input": "pkg:npm/%40foo%F0%9F%98%83/bar@1.0.0",
+      "expected_output": null,
+      "expected_failure": true,
+      "expected_failure_reason": "Should fail to parse because the scope contains characters not allowed by npm naming rules"
+    },
```
matt-phylum, line 404: Again, PURL does not define validation rules to forbid this so the tests should not test for that validation.

tests/types/npm-test.json
```diff
@@ -254,6 +254,316 @@
...
+    {
+      "description": "Normalization Test: Both scope and name with uppercase letters are lowercased",
+      "test_group": "advanced",
+      "test_type": "roundtrip",
+      "input": "pkg:npm/%40MyScope/MyPackage@1.0.0",
+      "expected_output": "pkg:npm/%40myscope/mypackage@1.0.0",
+      "expected_failure": false,
+      "expected_failure_reason": null
+    },
```
matt-phylum, line 377: This test makes the previous (incorrect) test about lowercasing the name redundant.

tests/types/npm-test.json
```diff
@@ -254,6 +254,316 @@
...
+        "name": "foo",
+        "version": "1.0.0",
+        "qualifiers": {
+          "vcs_url": "https://a.com/b/c😃d.git"
+        },
+        "subpath": null
+      },
+      "expected_failure": false,
+      "expected_failure_reason": null
+    },
```
matt-phylum, line 456: Do these tests have anything to do with NPM? They seem like they should be basic tests for the core spec.

tests/types/npm-test.json
prabhu: These are tests based only on the contents in the type definitions. The PRs are drafted to encourage comments like these, so thanks a lot.

To me (and my bot), the test looks valid because there is a note that explicitly says the name must be lowercased.

https://github.com/package-url/purl-spec/blob/96d6f8c4123d4a6a5c2bbc90cf61fca834f421d9/types/npm-definition.json#L21

tests/types/npm-test.json
prabhu: Agreed. This normalization test must be rewritten in a different manner.

tests/types/npm-test.json
prabhu: https://github.com/package-url/purl-spec/blob/96d6f8c4123d4a6a5c2bbc90cf61fca834f421d9/types/npm-definition.json#L16

tests/types/npm-test.json
prabhu: Good question. My bot includes information from additional sources.

https://github.com/npm/validate-npm-package-name/blob/main/test/index.js#L71
https://github.com/npm/validate-npm-package-name/blob/main/test/index.js#L91

tests/types/npm-test.json
prabhu: I noticed that there were no such tests. So, I pushed a commit to showcase some.

tests/types/npm-test.json
prabhu: npm does forbid, though. I agree that these shouldn't be in the purl-spec tests but must be present in SBOM tools at least.

tests/types/npm-test.json
matt-phylum: That's what the note says, but:
1. I don't think it is possible for a type spec like NPM to alter the percent encoding behavior. Percent encoding is not type-specific.
2. Whether the @ sign is supposed to be encoded or not does not specify the behavior of parsing when it is not.

tests/types/npm-test.json
matt-phylum: This has been a known issue with `pkg:npm` since 2021: #136.

Even following the incorrect spec, if the name is case insensitive, then parsing a name with mixed case should not result in an error.

tests/types/npm-test.json
matt-phylum: Including name validation rules makes PURL implementations more complicated to write and more likely to be wrong, and PURL more likely to break as ecosystems evolve. I don't think this adds any benefit because whether somebody writes `pkg:npm/_foo` (invalid according to this test) or `pkg:npm/a0103e9a-398d-4136-bced-c1cdbe95ebcd`, neither PURL can be resolved to a package today.

tests/types/npm-test.json
ppkarwasz: This is yet another example of the difference between the specification and the [instructions for parser developers](https://github.com/package-url/purl-spec/blob/main/docs/how-to-parse.md).
According to the proposed parsing algorithm, `@` can be safely used as the test input. Of course the canonical form will need to encode the character as `%40`.

Issue #609, opened 2025-08-19
Complete Schema Sections for the Ecma Standard document 


mjherzog:
For the Ecma standard documentation we need to provide a "human-readable", static representation of the schemas.

We found a FOSS tool for JSON Schema documentation:  https://github.com/coveooss/json-schema-for-humans (JSFH) that provides dynamic and static document generation from JSON Schema in HTML and Markdown formats. The tool supports JSON Schema draft-07 and is Python-based so it seems to be a good fit for our purpose. The HTML format should be very useful for future use with the packageurl.org website, but our current focus is on the static markdown output for the Ecma standard documentation.

The primary changes we need to the default JSFH output are:

- Removing the generated table of contents (easy)
- Renumbering the schema elements to Section 6 for the PURL Type schemas and Section 7 for the PURL Test schema to fit the TOC for the Ecma standard document (feasible to automate where the target numbering follows the generated numbering)
- Adjusting sub-section headings to be more human-friendly (seems feasible to automate in most cases)
- Reformatting the Property "path" to be more human-friendly (seems feasible to automate in most cases)
- Representing the allOf section of the PURL Test schema is very challenging and will likely require manual adjustment - see also https://json-schema.org/understanding-json-schema/reference/combining#allOf

mjherzog:
The drafts for Sections 6 and 7 are:

- Section 6: [purl-type-definition.schema-draft-01.md](https://github.com/user-attachments/files/21859894/purl-type-definition.schema-draft-01.md)
- Section 7: [purl-test.schema-draft-01.md](https://github.com/user-attachments/files/21860050/purl-test.schema-draft-01.md)

These files are also available on the purl-spec Gdrive at: https://drive.google.com/drive/folders/1hXwvN-K6cVv_wPeV54T5QbRUQNZDDM3B?usp=drive_link 

For both Section 6 and 7 we will add introductory text at the top of each file. 
- For Section 6 this will include the very simple schema for purl-types-index.json
- For Section 7 this will include an explanation of the complex structure of the test requirements array (7.2.1) and its allOf constraint - see also https://json-schema.org/understanding-json-schema/reference/combining#boolean-json-schema-combination

The next step is to confirm whether these drafts adequately provide the human-readable schema documentation for the Ecma Standard. The key question is the structure of the information. The formatting (fonts, spacing, etc.) can be adjusted after we agree on the structure.

mjherzog:
[...]

PR #656, opened 2025-09-05
Add, update .md files in /standard #609

johnmhoran:
Reference: https://github.com/package-url/purl-spec/issues/609#issuecomment-3259565257


mjherzog:
Bypassing rules to sync repo with latest files for Ecma standard. These may need to be updated based on feedback from the Ecma review process.

PR #663, opened 2025-09-09
Document "requirement" to PURL components

pombredanne:
As discussed in the course of the Ecma standardization, it may be best to clarify components requirement. Here we use documentation instead of modifying the core schema to clarify requirements.

- See also the PR at https://github.com/package-url/purl-spec/pull/664 that updates the schema with explicit requirement property


mjherzog:
Closing because it has been superseded by #665

PR #664, opened 2025-09-09
Add "requirement" to PURL components

pombredanne:
As discussed in the course of the Ecma standardization, it may be best to make the requirement explicit.

This PR update the core schema.

- See also the PR at https://github.com/package-url/purl-spec/pull/663 that only documents requirements.


pombredanne:
@mjherzog Thanks for the updates. :+1:

PR #665, opened 2025-09-10
Update purl-type-definition.json.schema to improve readability

mjherzog:
This PR covers 3 changes:
- Add optional_requirement and required_requirement properties to make requirement status explicit for each PURL component of a PURL type
- Updated most description text for clarity and parallel construction.


mjherzog:
@pombredanne I just noticed that we also need to update the schema id URL in the PURL type and test files. 
I backed out the changes to the schema id URL - we should do that later as a global change since the URL appears in every file in:
- /types-docs/
- /types/
- /tests/types/

pombredanne (approved):
Thanks for this!

PR #666, opened 2025-09-11
ECMA changes and corrections

stevespringett:
There are multiple changes here. They include:

- Removing `purl` from being represented as code to PURL as in the written form.
- Minor grammatical corrections
- Removed duplicate bullet point from PURL components as the same information is represented in the table with better descriptions.
- Corrected PURL type definitions schema - added explicit type to two properties.
- Corrected another occurrence of number sign to hash sign
- Updated normative references

### DIFF
--- a/docs/standard/references.md
+++ b/docs/standard/references.md
@@ -4,17 +4,17 @@ The following referenced documents are indispensable for the application of this
 For dated references, only the edition cited applies. For undated references, the latest
 edition of the referenced document (including any amendments) applies.
 
-INCITS 4-1986 [R2022], *Information Systems - Coded Character Sets - 7-Bit Standard Code for Information Interchange (7-Bit ASCII)*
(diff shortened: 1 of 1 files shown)


docs/standard/summary.md
```diff
@@ -1,30 +1,13 @@
...
-  gem, pypi, etc. Required.
-- **namespace**: some name prefix such as a Maven groupid, a Docker image owner,
-  a GitHub user or organization. Optional and type-specific.
-- **name**: the name of the package. Required.
-- **version**: the version of the package. Optional.
-- **qualifiers**: extra qualifying data for a package such as an OS,
-  architecture, a distro, etc. Optional and type-specific.
-- **subpath**: extra subpath within a package, relative to the package root.
-  Optional.
-
```
stevespringett, line 26: FYI. This information is better represented in the table below this content, which additionally has better descriptions of each of these.

PR #667, opened 2025-09-11
Improve requirement schema

pombredanne:
This PR corrects a few problems in the JSON schema:

- It supersedes #665
- The definition of the "requirement" is unified and streamlined using single common approach
- The schema id is using the correct `const` keyword for constant and a few JSON type definitions have been updated to match this corrected constraint
- The Markdown and Python script are regenerated


pombredanne:
Thanks!
Merging.

Issue #287, opened 2024-01-30
Support for vscode type


prabhu:
With an ever-growing list of malicious plugins, vscode rightly requires its type, similar to sublime and others. I'm proposing to use `vsix` as the type to represent the file extension used by these extensions since the same extension can be installed locally as well as via marketplaces such as Visual Studio and open-vsx.

Eg:

https://marketplace.visualstudio.com/items?itemName=ms-python.python

```
pkg:vsix/ms-python.python@2023.25.10292213
pkg:vsix/ms-python.python@2023.25.10292213?download_url=https%3A%2F%2Fmarketplace.visualstudio.com%2F_apis%2Fpublic%2Fgallery%2Fpublishers%2Fms-python%2Fvsextensions%2Fpython%2F2023.25.10292213%2Fvspackage
```

vscode extensions installed from [open-vsx](https://open-vsx.org)

Eg: https://open-vsx.org/extension/muhammad-sammy/csharp

```
pkg:vsix/muhammad-sammy/csharp@2.15.30?download_url=https%3A%2F%2Fgithub.com%2Fmuhammadsammy%2Ffree-vscode-csharp%2Freleases%2Fdownload%2F2.15.30%2Fcsharp-platform-neutral-2.15.30.vsix
```

pombredanne:
@prabhu This looks fine, can you draft a type definition? (PS: I am not convinced by the `vsix` name as being the most obvious thing)

prabhu:
@pombredanne, what about vsx similar to open-vsx?

prabhu:
vscode is constantly in the [news](https://unit42.paloaltonetworks.com/stately-taurus-abuses-vscode-southeast-asian-espionage/) these days. What is the process to get a type definition going?

booniepepper:
I'm taking a swing at this. @prabhu I started with what you had, but changed "download_url" to the repository https://github.com/package-url/purl-spec/pull/673

@pombredanne @johnmhoran Do you think `vsx` is a reasonable PURL type for VS Code Extensions?

mjherzog:
Closing because we now have the vscode-extension PURL type.

Issue #644, opened 2025-08-26
Expected output in pkg:generic testcase does not follow spec encoding


petergardfjall:
[This test case](https://github.com/package-url/purl-spec/blob/96d6f8c4123d4a6a5c2bbc90cf61fca834f421d9/tests/types/generic-test.json#L69) states that given input 

`pkg:generic/openssl@1.1.10g?download_url=https://openssl.org/source/openssl-1.1.0g.tar.gz&checksum=sha256:de4d501267da` 

the expected output is 

`pkg:generic/openssl@1.1.10g?checksum=sha256:de4d501267da&download_url=https://openssl.org/source/openssl-1.1.0g.tar.gz`

I think this does not follow the [canonical purl encoding](https://github.com/package-url/purl-spec/blob/main/PURL-SPECIFICATION.rst#rules-for-each-purl-component) which (correct me if I'm wrong!) should be:

`pkg:generic/openssl@1.1.10g?checksum=sha256:de4d501267da&download_url=https:%2F%2Fopenssl.org%2Fsource%2Fopenssl-1.1.0g.tar.gz`

petergardfjall:
If this is true I see similar issues with other test-cases as well:
- hex-test.json
- huggingface-test.json
- luarocks-test.json
- maven-test.json
- mlflow-test.json
- npm-test.json
- oci-test.json

petergardfjall:
No one has responded to this issue so I'm not sure if I'm right in the assessment but I've put up a number of PRs to fix the tests.
- #713 
- #714
- #715
- #716
- #717
- https://github.com/package-url/purl-spec/pull/718
- https://github.com/package-url/purl-spec/pull/719
- https://github.com/package-url/purl-spec/pull/720
- https://github.com/package-url/purl-spec/pull/721

rhalar:
I believe you are correct, this is said for the qualifier `value` field:
https://github.com/package-url/purl-spec/blob/main/docs/standard/components.md
>A value may contain any Unicode character and all characters must be encoded as described in the "Character encoding" section.

The ["Character encoding" section](https://github.com/package-url/purl-spec/blob/main/docs/standard/characters-and-encoding.md#character-encoding) does not exclude [separators](https://github.com/package-url/purl-spec/blob/main/docs/standard/characters-and-encoding.md#permitted-characters) from being encoded -  the closest definition is
>The following characters must not be percent-encoded:
> ...
> - the Separator Characters when being used as PURL separators,

which does not hold.

[...]

PR #673, opened 2025-09-17
Add VS Code extensions

booniepepper:
Related to https://github.com/package-url/purl-spec/issues/287

Clashes a little with a different approach at https://github.com/package-url/purl-spec/pull/671 since this will consider the VS Code extension marketplace as the default repository

I'm not sure that `vsx` is the right identifier, maybe `vscode` is better? The file extension for these IDE extension packages is `.vsix`.

Happy to hear & take feedback, thanks!

### DIFF
--- a/types-doc/vsx-definition.md
+++ b/types-doc/vsx-definition.md
@@ -1,0 +1,42 @@
+<!--  NOTE: Auto-generated from the JSON PURL type definition.
+Do not manually edit this file. Edit the JSON type definition instead. -->
+
+# PURL Type Definition: vsx
+
+- **Type Name:** VS Code Extension packages
+- **Description:** VS Code Extension packages
+- **Schema ID:** `https://packageurl.org/types/vsx-definition.json`
+
+## PURL Syntax
+
+The structure of a PURL for this package type is:
+
+    pkg:vsx/<namespace>/<name>@<version>?<qualifiers>#<subpath>
+
+## Repository Information
+
+- **Use Repository:** Yes
+- **Default Repository URL:** https://marketplace.visualstudio.com/vscode
+
+## Namespace definition
+
+- **Requirement:** Optional
+- **Case Sensitive:** Yes
+- **Native Label:** publisher
+
+## Name definition
+
+- **Requirement:** Required
+- **Case Sensitive:** Yes
+- **Native Label:** name
+
+## Version definition
+
+- **Requirement:** Optional
+- **Case Sensitive:** Yes
+- **Native Label:** version
+
+## Examples
+
+- `pkg:vsix/ms-python/python@2023.25.10292213`
+- `pkg:vsix/muhammad-sammy/csharp@2.15.30?repository_url=open-vsx.org`
--- a/types/vsx-definition.json
+++ b/types/vsx-definition.json
@@ -1,0 +1,30 @@
+{
+  "$schema": "https://packageurl.org/schemas/purl-type-definition.schema-1.0.json",
+  "$id": "https://packageurl.org/types/vsx-definition.json",
+  "type": "vsx",
+  "type_name": "VS Code Extension packages",
+  "description": "VS Code Extension packages",
+  "repository": {
+    "use_repository": true,
+    "default_repository_url": "https://marketplace.visualstudio.com/vscode"
+  },
+  "namespace_definition": {
+    "requirement": "optional",
+    "case_sensitive": true,
+    "native_name": "publisher"
+  },
+  "name_definition": {
+    "requirement": "required",
+    "case_sensitive": true,
+    "native_name": "name"
+  },
+  "version_definition": {
+    "requirement": "optional",
+    "case_sensitive": true,
+    "native_name": "version"
+  },
+  "examples": [
+    "pkg:vsix/ms-python/python@2023.25.10292213",
+    "pkg:vsix/muhammad-sammy/csharp@2.15.30?repository_url=open-vsx.org"
+  ]
+}
--- a/purl-types-index.json
+++ b/purl-types-index.json
@@ -30,5 +30,6 @@
   "qpkg",
   "rpm",
   "swid",
-  "swift"
+  "swift",
+  "vsx"
 ]


booniepepper:
@jkowalleck Any thoughts on this?

mixmix:
As I raised here : https://github.com/package-url/purl-spec/pull/671#discussion_r2377128615 , I'm really interested in how we keep consumers of PURLs safe from things like typosquat attacks. Some context is I work in security and have had a client ask me to review an extension giving just it's name ... and it was really hard to answer their question because I did not know which code I was meant to be auditing! (because there are two registries and they can have different code).

To that end I'm against `vsx` because I'm scared of it being a foot-gun. I think `vscode` would be a lot safer.

I think we should explore the question of how we handle referencing extensions in these two registries before merging these PRs though. Propose we continue discussion in that other PR (it's got a bit of discussion already)

booniepepper:
I have examples in the PR. I think it's easy to distinguish the source by `repository_url` and this follows the convention used in many other PURL specs.

P.S. I incorporated your changes and credited you in one of the commits included in this PR

P.P.S. I've also been working on software supply chain security for much of the past 5 years

mixmix:
Update: after some great input from community, have opted to close https://github.com/package-url/purl-spec/pull/671 in favour of this proposal ❤️

types/vsx-definition.json
```diff
@@ -0,0 +1,46 @@
...
+      "key": "platform",
+      "requirement": "optional",
+      "description": "The target platform for the extension. Common values include 'universal', 'linux-x64', 'linux-arm64', 'darwin-x64', 'darwin-arm64', 'win32-x64', 'win32-arm64', etc.",
+      "default_value": "universal",
+      "native_name": "targetPlatform"
+    }
+  ],
+  "examples": [
+    "pkg:vsx/ms-python/python@2023.25.10292213",
+    "pkg:vsx/muhammad-sammy/csharp@2.15.30?repository_url=open-vsx.org",
```
mixmix: Questions
1. should the `repositoy_url` be in the same format as the default i.e. include `https://` (but URL encoded) ? 
2. do we want to include in the spec what the valid values of this param to reduce confusion /  errors?

booniepepper: Thanks @mixmix for feedback and continuing to collaborate on this!

I'm seeing different types have taken different approaches on whether to include the protocol: https://github.com/search?q=repo%3Apackage-url%2Fpurl-spec+%22repository_url%3D%22+path%3Atypes%2F&type=code

After reviewing I think I agree that we should probably include the `https:` protocol. I'll update the PR

types/vsx-definition.json
```diff
@@ -0,0 +1,46 @@
...
+  "type": "vsx",
+  "type_name": "VS Code Extension packages",
+  "description": "VS Code Extension packages",
+  "repository": {
+    "use_repository": true,
+    "default_repository_url": "https://marketplace.visualstudio.com/vscode"
+  },
+  "namespace_definition": {
+    "requirement": "optional",
+    "case_sensitive": true,
```
rhalar: Are you certain of these? 

We wanted to add VS Marketplace to the [OSV schema](https://github.com/ossf/osv-schema) to be able to add malicious package findings to the [OSSF repository](https://github.com/ossf/malicious-packages).
https://github.com/ossf/osv-schema/pull/357

While doing some research for this we've come to the conclusion that none of the components are case-sensitive, I mentioned it in the [comment there](https://github.com/ossf/osv-schema/pull/357#discussion_r2205076798).

We've also presumed that the namespace is required, since I don't think you can have an extension hosted without a publisher, but not sure how OpenVSX behaves if that's also something that should be covered by these definitions?
There are also situations where different publishers may use the same package name, so the publisher is necessary to create a unique identifier for each package.

types/vscode-extension-definition.json
```diff
@@ -0,0 +1,46 @@
...
+    "use_repository": true,
+    "default_repository_url": "https://marketplace.visualstudio.com/vscode"
+  },
+  "namespace_definition": {
+    "requirement": "optional",
+    "case_sensitive": true,
+    "native_name": "publisher",
+    "note": "The namespace is the publisher or vendor of the extension (e.g., 'redhat', 'microsoft')"
+  },
+  "name_definition": {
```
rhalar, line 17: I'm also interested if these definitions cover Visual Studio plugins as well, since these are hosted on the Marketpace as well (https://marketplace.visualstudio.com/vs)?

We've been using an internal VSX PURL implementation for some time, and we've come across the weird situation where these plugins don't encode the `name` and `publisher` in the package manifest in the same way as VSCode extensions do.
In the manifest, these are GUIDs and non-normalized strings, e.g. if we download and scan
https://marketplace.visualstudio.com/items?itemName=SharpDevelopTeam.ILSpy

we'd get `pkg:vsx/sharpdevelop%20team/ebf12ca7-a1fd-4aee-a894-4a0c5682fc2f@9.1.0.7988`. Instead of what you'd perhaps expect, i.e. for it to reflect the `itemName` on the repoistory (`itemName=SharpDevelopTeam.ILSpy`). 
It makes it a bit difficult for tools to connects a package to a corresponding entry in the repository. I haven't come back to this since, but interested to hear if you have any insights into how you'd like to see these handled?

types/vsx-definition.json
pombredanne: @rhalar good points!

@spoenemann @amvanbaren can you chime in for OpenVSX? That would be awesome!

@mbarbero FYI [too](url)

types/vscode-extension-definition.json
pombredanne: @rhalar either this is for vsx only or for openvsx only of for both... I like the later! .... so do you have other weird examples that could become matter for the test suite?

rhalar: There is another thing that only seriously affects VS plugins, but it seems there is some kind of background mapping going on for the `Publisher` information. VS _Code_ packages seem to handle this well, the manifest contains what you'd expect.

On the repository site, these both map to `Microsoft` even though the namespace is technically different
https://marketplace.visualstudio.com/items?itemName=ms-python.python
https://marketplace.visualstudio.com/items?itemName=ms-toolsai.jupyter

The manifest will corresponds with the itemName
`... Id="python" Version="2025.10.0" Publisher="ms-python" ...`

They even acknowledge this in the `More Info` metadata section on the repository, by including a `Unique Identifier` field in such cases.


However, for VS plugins this example also maps to `Microsoft` but things behave more inconsistently
https://marketplace.visualstudio.com/items?itemName=ProBITools.MicrosoftReportProjectsforVisualStudio

The manifest:
`... Id="717ad572-c4b7-435c-c166-c2969777f718" Version="2.6.11" Publisher="Microsoft" ...`

`ProBITools` isn't mentioned anywhere, so it's even a bit worse.

types/vsx-definition.json
booniepepper: Thanks for the feedback and taking a look everyone!

I looked into it, and given the onboarding process of Open VSX _(I had to sign multiple documents with the Eclipse Foundation before I could post an extension)_ I believe a publisher is absolutely required and this will likely never change.

Also agree on case sensitivity now

types/vscode-extension-definition.json
booniepepper: > @rhalar either this is for vsx only or for openvsx only of for both... I like the later! .... so do you have other weird examples that could become matter for the test suite?

I think the initial comment by @rhalar is that there are two different **sets** of ecosystems.

1. Two VS Code (JS) extension repositories
    * https://marketplace.visualstudio.com/vscode
    * https://open-vsx.org/
2. One Visual Studio (C#/.NET) extension repository
    * https://marketplace.visualstudio.com/vs

The two IDEs are not cross-compatible, the plugins for VS Code can only work with VS Code, and the same goes for Visual Studio. The extensions are implemented in different languages and interact with different sets of APIs. From my perspective, they are just not the same ecosystem.

What they do share is a file extension `.vsix` for a zip archive.

The zip archive contents are very different. I grabbed a few Visual Studio extensions, and these contained lots of x86/arm64 `.dll` files. _(Presumably compiled and cross-compiled on Windows from C# through Visual Studio.)_ VS Code extensions follow a more standard JS structure of a `package.json` and lots of `.js` files. Both kinds of extensions contain an `extension.vsixmanifest` (xml) file, but these diverge based on a `<InstallationTarget>` tag and diverge on many other properties and structures.

Open to feedback, but I think it's worth considering Visual Studio extensions as separate from VS Code extensions. The similarities are very surface level.

This might mean, though, that it's worth reconsidering the type as `vscode` instead of `vsx`

types/vsx-definition.json
mixmix: I've been building tools around Open VSX and my current experience/ understanding confirms the above:
- package name is **case-insensitive**
- **package name includes namespace**:
  - `{namespace}.{extensionName}` (where namespace is similar to publisher, but different to publishedBy)
  - associated purl is then `pkg:vsx/{namespace}/extensionName}`

types/vscode-extension-definition.json
mixmix: _My sad-face emojis = hearing ghost stories about inconsistent data sources (not disagreement with @rhalar)_

When we merged this PR with mine (for OpenVSX) it was on the basis of "these extensions are interoperable", so I agree that the VS (Visual Studio) is out of scope for this PURL.

I think @booniepepper 's suggestion of a rename to `vscode` could be good if it minimize surprises/ errors for users and developers. _Might_ be worth a quick side-quest to check that there's a sane Visual Studio purl option (vs seems way too short = collision prone)

rhalar: I'm not an authority, but my understanding is that the format _is_ shared, the two extension types contain different files and structures but both fulfill the requirements to be considered a VSIX package.

It seems just using a different repository ID for these would suffice, and both could use the same algorithm for PURL construction through the manifest, `Id` for the name part and `Publisher` for the namespace. Both of these pieces of information are also available on the marketplace (note the `Copy ID` button in the `Resource` section, for VS extensions only, which supplies the GUID).

It just needs to be well defined since the `itemName` parameter seems to not suffice here, at least for VS extensions.

amvanbaren:
Hi! I don't know why, but I got notified about this issue. Any way I can help?

types/vscode-extension-definition.json
matt-phylum: Whether the package contains javascript files or dll files and only works in a certain environment doesn't seem relevant. This is already the case for at least `pkg:gem` (eg platform java) and `pkg:nuget` (eg C++ or powershell) and `pkg:pypi` (eg a wheel that contains only a .pyd file for PyPy).

pohlarized: Given all of these aspects that are not relevant for distinction between types, i start to wonder what aspects are relevant.
Especially for the vscode vs visual studio case, the only common things i see are
- same creator (Microsoft)
- similar names
- MS store lives on the same domain (and perhaps have a similar API?)
- extensions have same file extension / file type (`.vsix` / `zip`)

However, the editors have nothing else in common, and the extensions too. Unlike in the `pypi` case, it's not like they're different kinds of distributions for the same source, or in the `gem` case that they are made for the same language but running on a different underlying platform.

rhalar: The common thread is that they share a package format, VSIX isn't just the extension.
>The VSIX package format follows the Open Packaging Conventions (OPC) standard. The package contains binaries and supporting files, together with a [Content_Types].xml file and a .vsix manifest file.

Microsoft decided to use the same 'type' for multiple ecosystems (Visual Studio and Visual Studio Code extensions).

Maven has something kinda similar where it can also cover Jenkins plugins (https://mvnrepository.com/artifact/org.jenkins-ci.plugins.workflow/workflow-step-api) and Scala packages (https://mvnrepository.com/artifact/org.scalacheck/scalacheck).
These both follow the same format as classic JAR files, and could be encoded as `pkg:maven...` PURLs.

mixmix: It doesn't help that we don't have a clear definition of what a "Package" is, and I notice on the root README, the "type" field (which we're debating here) is defined as:

> **type**: the package "type" or package "protocol" such as maven, npm, nuget, gem, pypi, etc. Required.

---

I see two paths. 

1. **We organise around package "protocol"** 
    - the "type" is then `vsix`
    - e.g. `pkg:vsix/redhat/java` (default repository, i.e. vscode)
    - e.g. `pkg:vsix/jetbrains/resharper?repository_url=marketplace.visualstudio.com/vs`
2. **We organise around package "type"**
    - type is an overloaded word here!
    - we split out types `vscode` and `visualstudio` (for example)
    - e.g. `pkg:vscode/redhat/java`
    - e.g. `pkg:visualstudio/jetbrains/resharper`
    
I think they can both be argued as "correct". I am then interested in the usefulness, the UX/DX of each pathway. Personally as an engineer I would feel a lot less annoyed with (2). In contrast with (1), it feels a bit overloaded, like we'd be making `pkg:zip/` to me, and the `respository_url` seems like un-necessary mess (like being unclear what sort of software package till I open it potentially). 

If we're strongly dislike (2), then lets make sure we thoroughly explore the edge-cases of (1). For example notice how I had a go in (1) above showing what the example purls would look like.

trevnorris: I'm chiming in to say I support approach 2. From a practical security tooling perspective, we need PURLs to clearly identify what we're dealing with at a glance. With approach 1 (`pkg:vsix/...`), our tools would need to:

1. Parse the PURL
1. Check if there's a `repository_url` parameter
1. Infer which ecosystem we're in based on that parameter (or assume default)
1. Then understand what we're actually scanning

With approach 2, the type field immediately tells us which ecosystem and IDE we're working with. No guessing, no extra parsing logic, no assumptions about defaults.

This matters for:

* Vulnerability databases - we need to know if a CVE applies to VS Code or Visual Studio extensions
* Policy enforcement - organizations may have different policies for different IDE ecosystems
* Tool integration - scanners and SBOMs shouldn't need complex logic to figure out what they're looking at

types/vsx-definition.json
booniepepper: Added in [07a7a75](https://github.com/package-url/purl-spec/pull/673/commits/07a7a75aecd35fd18130712a175b6107eee639d5)

types/vsx-definition.json
booniepepper: updated in [07a7a75](https://github.com/package-url/purl-spec/pull/673/commits/07a7a75aecd35fd18130712a175b6107eee639d5)

types/vscode-extension-definition.json
booniepepper: Updated to use `vscode` instead of `vsx` in [07a7a75](https://github.com/package-url/purl-spec/pull/673/commits/07a7a75aecd35fd18130712a175b6107eee639d5)

booniepepper:
@amvanbaren The PR was waiting on me to update (which I have now) -- I'm not sure either, but the more eyes the merrier

types/vscode-extension-definition.json
pohlarized: I would agree that in general I'm more in favor of more distinct types, especially because these types are requires and thus more explicitly communicate what package exactly you're referring to, and "hiding" as little as possible behind optional fields like the repository URL.
Of course, it would be great to have some more "proper" definition of what makes a type for the future, because arbitrarily fine-grained types probably also don't help too much.

Also, just to touch on the same "format"; I think it's a difficult concept to use as a decider for whether something should be merged as a type or not, since a "format" can be arbitrarily general or specific.
Something like mixmix' example of `zip` would be such a very generic format, and i would argue that the definition of what makes a `vsix` package is also very generic, to the point that i'd personally say it does not make too much sense to use it as a package type just due to the format alone.

types-doc/vsx-definition.md
```diff
@@ -0,0 +1,53 @@
+<!--  NOTE: Auto-generated from the JSON PURL type definition.
```
mixmix: 🔥 TODO: delete this auto-generated file @booniepepper

mixmix (approved):
Once again, really appreciate the depth of everyones attention to detail, listening, and thoughtful dialogue. I think we've come to something which is better than any one of alone would have suggesting 🌈 🚀

types-doc/vsx-definition.md
booniepepper: I think the status quo in this project is to commit the generated docs: https://github.com/package-url/purl-spec/tree/main/types-doc

(I believe the intention is that this allows them to render directly on GitHub)

types-doc/vsx-definition.md
```diff
@@ -0,0 +1,53 @@
...
+Do not manually edit this file. Edit the JSON type definition instead. -->
+
+# PURL Type Definition: vsx
```
rhalar: I am a bit confused, are both supposed to be included, or just the `vscode` one?

types-doc/vsx-definition.md
mixmix: This generated doc cam from `types/vsx-definition.json`, which was migrated to `types/vscode-definition.json` - i.e. I believe this is an orphaned generated file that should be removed

mixmix: hmm, cannot unresolve

types-doc/vsx-definition.md
mixmix: This generated doc cam from `types/vsx-definition.json`, which was migrated to `types/vscode-definition.json` - i.e. I believe this is an orphaned generated file that should be removed

booniepepper: Oops, yes this should have been removed 🙈

mixmix: This is why peer reviewing is beautiful. I would be just as prone to make such a mistake ❤️

types/vscode-definition.json
```diff
@@ -0,0 +1,46 @@
...
+  "namespace_definition": {
+    "requirement": "required",
+    "case_sensitive": false,
+    "native_name": "publisher",
+    "note": "The namespace is the publisher or vendor of the extension (e.g., 'redhat', 'microsoft')"
+  },
+  "name_definition": {
+    "requirement": "required",
+    "case_sensitive": false,
+    "native_name": "name"
```
rhalar: I'm not sure what is in scope for these `type` definitions in general, but should it be pointed out where these values come from?
As I mentioned in other comments, in order to create a unique identity which matches you'd have to know to take `Id` and `Publisher` from the `.vsixmanifest` (and not i.e. `package.json`) of a package, or I guess the `itemName` from the Marketplace API instead of display names, which sometimes differ.

Other than that, I think this look good now.

booniepepper: Sorry I think I missed this in the other discussions

Personally when I think about "where these values come from" I've always held that the download URL<sup>[1]</sup> is more authoritative than the contents.

The point is taken, though, there are numerous contexts where an artifact or a PURL may be evaluated, not just the time of resolution, and my spec isn't giving much guidance on that front yet

---

1. _Download URL, or:whatever the relevant packaging protocol is. The artifact coordinate that a registry resolved to an artifact. I am putting on my tinfoil hat, but there are a number of ways that the contents of an artifact can be compromised before, during, or after a package is downloaded depending on what is compromised (publisher, public registry, network traffic, private registry, working environment, etc...)_

booniepepper: Added to the notes in https://github.com/package-url/purl-spec/pull/673/commits/0b9bf69a33782cd0517ec541fc44459626bac1fe

rhalar: The URL, or I presume the `itemName` parameter more specifically, would work for `VSIX` as a possible other source. Registries usually expose an API which returns exact metadata fields which can be used for the purpose, and the raw download URL is just part of that, or they at least guarantee that the URL consist of those well defined parameters which can reconstruct a unique identity. But the Marketplace is a weird outlier since they don't really have an official API, or documentation for any of the values found on the site.

The thing we care about is just that we use can construct the same identifier given just the package, URL or dependency.

types-doc/vscode-definition.md
```diff
@@ -0,0 +1,54 @@
...
+## Repository Information
+
+- **Use Repository:** Yes
+- **Default Repository URL:** https://marketplace.visualstudio.com/vscode
+
+## Namespace definition
+
+- **Requirement:** Required
+- **Native Label:** publisher
+- **Note:** `The namespace is the publisher or vendor of the extension (e.g., 'redhat', 'microsoft'). This corresponds to the Publisher value in the package's .vsixmanifest file.`
```
rhalar: It'd be more correct so say something like
`This corresponds to the 'Publisher' property of the 'Identity' element in the package's .vsixmanifest file.`

The same applies to the `name` definition

booniepepper: Thanks, that's more precise!

Added in https://github.com/package-url/purl-spec/pull/673/commits/8bb58fd1001777bbe4e67aa086f794591bb6c0ac. I also added the same wording for `version`

rhalar (approved):

booniepepper:
@pombredanne Looks like we've got all the comments addressed. Are you ready to take a look again?

types/vscode-definition.json
```diff
@@ -0,0 +1,48 @@
...
+      "key": "platform",
+      "requirement": "optional",
+      "description": "The target platform for the extension. Common values include 'universal', 'linux-x64', 'linux-arm64', 'darwin-x64', 'darwin-arm64', 'win32-x64', 'win32-arm64', etc.",
+      "default_value": "universal",
+      "native_name": "targetPlatform"
+    }
+  ],
+  "examples": [
+    "pkg:vscode/ms-python/python@2023.25.10292213",
+    "pkg:vscode/muhammad-sammy/csharp@2.15.30?repository_url=https://open-vsx.org",
```
mixmix: The encoding of `qualifiers` is currently being discussed https://github.com/package-url/purl-spec/issues/644

I'd like to propose we side-step that by changing this example to:
```suggestion
    "pkg:vscode/muhammad-sammy/csharp@2.15.30?repository_url=open-vsx.org",
```

I notice in the code I'm writing that I'm likely going to just be checking for `open-vsx.org` because I don't trust users to consistently encode correctly (regardless of what the "correct" way is 😆)

Either way, I'd love to install an open-vsx `repository_url` in `tests/types/vscode-test.json` as a clear signal + proof of functionality.

Thoughts @booniepepper ?

types/vscode-definition.json
```diff
@@ -0,0 +1,48 @@
...
+    {
+      "key": "platform",
+      "requirement": "optional",
+      "description": "The target platform for the extension. Common values include 'universal', 'linux-x64', 'linux-arm64', 'darwin-x64', 'darwin-arm64', 'win32-x64', 'win32-arm64', etc.",
+      "default_value": "universal",
+      "native_name": "targetPlatform"
+    }
+  ],
+  "examples": [
+    "pkg:vscode/ms-python/python@2023.25.10292213",
```
mixmix: I have feedback from early internal testing that it's quite annoying that the extension is identified by `extensionId` which uses a `.` (`<publisher>.<name>`) in some places, and using a`/` delimited path in others (`<publisher>/<name>`).

Should we consider an alternative which uses the unique package-name like this: 
```suggestion
    "pkg:vscode/ms-python.python@2023.25.10292213",
```

Places I've seen `/`
- package url e.g. `https://open-vsx.org/extension/ms-python/python`
- the api e.g. `https://open-vsx.org/api/my-python/python/2025.16.0`

Places I've seen `.`
- package filename e.g `ms-python.python-2025.16.0.vsix`
- the api e.g. `https://open-vsx.org/api/-/query?extensionId=my-python.python`

rhalar: The publisher is defined to be the `namespace` component, which makes sense to me, which is why it's "extracted" from the "VSCode format" into the PURL one resulting in `namespace/name`. 

Maven has a similar convention where they do `GroudID:ArtifactID` but that also converts to `groupid/artifactID` in the PURL, for the same reason.
I kind of get the point but if you are constructing PURLs and following the spec there shouldn't be any problems or inconsistencies.

mixmix: I think you're right, openvsx even calls it a namespace : https://github.com/EclipseFdn/open-vsx.org/wiki/Guidelines-on-Namespace-Requests

Will mark this resolved

mixmix: ✅

types/vscode-definition.json
mixmix: The resolution of that issue is that slashes in `qualifers` need percentage encoding

```suggestion
    "pkg:vscode/muhammad-sammy/csharp@2.15.30?repository_url=https:%2F%2Fopen-vsx.org",
```

I think in practice in my application I'm going to accept all of the following as valid `qualifiers` (because I know users will get it wrong):

`?repository_url=open-vsx.org`
`?repository_url=open-vsx.org/`
`?repository_url=https://open-vsx.org`
`?repository_url=https://open-vsx.org/`
`?repository_url=https:%2F%2Fopen-vsx.org` << probably the most spec-aligned
`?repository_url=https:%2F%2Fopen-vsx.org%2F`

matt-phylum: Your application should be percent decoding the qualifier value, in which case `https:%2F%2Fopen-vsx.org` and `https://open-vsx.org/` are the same. If for some reason you cannot percent decode, you need to look out for `https%3A//open-vsx.org/`. Whether `:` is encoded as `%3A` in qualifiers used to be unclear and IIRC half of the PURL libraries did and half of the PURL libraries didn't.

(If you are using a PURL library, it will normally take care of the decoding for you when it extracts the qualifier value from the string)

types/vscode-definition.json
pombredanne: Ack, marking this as resolved then.

mjherzog:
@booniepepper @mixmix It would probably be more accurate to set the PURL type to "vscode-extension" to be more descriptive and to leave "space" just in case there are future PURL types related to vscode. What do you think?

booniepepper:
@mjherzog I was trying to keep it succinct, but I'd be happy with that more-precise proposal

mjherzog:
@johnmhoran Please double-check the tests when you have a minute.
@booniepepper We will need to update the PURL Types grid at: https://package-url.github.io/www.packageurl.org/docs/purl/purl-spec-purl-types#registered-purl-types for the new vscode-extension type.
Do you want to display an icon there? If yes, please tell us where to find the appropriate icon.

booniepepper:
@mjherzog Yeah, I think the "blue/stable" icon from here would be best: https://code.visualstudio.com/brand

Thanks @johnmhoran - lots of testing credit goes to @mixmix from a previous PR

mjherzog:
@johnmhoran Please update Tool grid with the blue/stable icon.

PR #700, opened 2025-10-06
Fix typos in golang-definition.json

nicorikken:
Reading the golang purl definition I found some minor issues to correct and increase readability.


johnmhoran (approved):
LGTM.

Issue #528, opened 2025-07-18
Update ADOPTERS


mjherzog:
We need to:
- Convert ADOPTERS.rst to markdown format
- Add known adopters
- Document a process for future adopters update the list with a PR

mjherzog:
- Merged https://github.com/package-url/purl-spec/pull/569

pombredanne:
Some missing:
- Microsoft SBOM tool
- OSV Scanner
- OpenVEX
- ISO CSAF
- AboutCode PurlDB, ScanCode.io, VulnerableCode, DejaCode
- Anchore Syft and Grype
- BANG
- Aquasec Trivy
- Snyk Parlay
- Snyk
- Black Duck
- Mend
- Siemens
- LG
- Meta
- Cariad
- deps.dev
- Eclipse Foundation
- SW360
- crates.io
- PSF Python SBOMs
- ASF projects including Log4J, SOLR, and more
- Various Linux Foundation project to list here
- OpenSSF GUAC
- OpenRailAssociation
- cdxgen
- atom+chen
- depscan

mjherzog:
This is an impressive list.  I think that we need to implement a process to confirm each adopter by either:

- Documenting the usage from open/public code - this should cover most cases.
- Contacting the supplier of closed code to confirm the usage and their permission to list the company here.

giterlizzi:
If you're interested, my SecDB project has adopted PURL since the first specification to track affected (or fixed) versions of products and packages in advisories.

- https://secdb.nttzen.cloud/pkg
- https://secdb.nttzen.cloud/pkg/type/pypi
- https://secdb.nttzen.cloud/pkg/detail/pkg:pypi/django
- https://secdb.nttzen.cloud/security-advisory/pypi/PYSEC-2025-47 (in the next version, it will use “from_native” from URI::VersionRange (Perl) to convert the native format to VERS)

(this portal also use URI-PackageURL)

mjherzog:
@giterlizzi Thank you.  We will add SecDB to Adopters.

mjherzog:
We may want to move the Adopters information to Tools on the packageurl.org website where we can use one system to track and report adopters of PURL and VERS.

mjherzog:
We have moved the Adopters information to:
- https://www.packageurl.org/docs/getting-started/toolgrid
- https://www.packageurl.org/docs/getting-started/specgrid

and we have Issue Templates at www.packageurl.org for Add a Tool and Add a Specification

PR #701, opened 2025-10-06
Added ZEN SecDB in ADOPTERS.md

giterlizzi:
Added ZEN SecDB, which uses PURL to identify vulnerable packages for different ecosystems.

#528


mjherzog (approved):
Approved to add.
Only question is whether we want/need to apply 78 character line length to this markdown file (and all other markdown files in the repository). It is important for tracking changes to specification files, but may not be essential for more general information like this.

johnmhoran (approved):
Thanks @giterlizzi -- LGTM.

PR #702, opened 2025-10-07
Add new PURL type: 'githubactions'

### LINKED ISSUE #698: Add githubactions PURL type
In the https://github.com/actions/dependency-review-action the packages can be filtered using purls. This includes `githubactions:` type purls. There was already a pull-request on this topic in: https://github.com/package-url/purl-spec/pull/243

The current implementation in dependency-review-action is case-sensitive which leads to practical issues, so a clear supported guidance would help.

Given earlier discussions in the pull-request, it should be noted that githubactions are unique in the sense that they can refer to a workflow as well, if I recall correctly, so a file within the repository.

nicorikken:
Add a new definition for GitHub Actions, including workflows, to the package URL specification.

This implementation aligns with current practice of using the `name` component to include both the name and the repository path.

Resolves: https://github.com/package-url/purl-spec/issues/698

### DIFF
--- a/types/githubactions-definition.json
+++ b/types/githubactions-definition.json
@@ -3,7 +3,7 @@
   "$id": "https://packageurl.org/types/githubactions-definition.json",
   "type": "githubactions",
   "type_name": "GitHub Actions",
-  "description": "GitHub Actions including workflows",
+  "description": "GitHub Actions",
   "repository": {
     "use_repository": true,
     "note": "Github.com is the Actions repository."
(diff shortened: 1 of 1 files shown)


types/githubactions-definition.json
```diff
@@ -0,0 +1,32 @@
...
+  },
+  "namespace_definition": {
+    "requirement": "required",
+    "case_sensitive": false,
+    "note": "The namespace must be lowercased."
+  },
+  "name_definition": {
+    "requirement": "required",
+    "case_sensitive": true,
+    "note": "The name contains both the repository name and optionally a path within the repository to a certain action or workflow. The first section before the slash is the repository name, this is not case sensitive and should be lowercase. The rest is the path within the repository, which is case sensitive."
```
matt-phylum: This isn't accurate. The name begins at the last slash. The complicated part with the partial case sensitivity is in the namespace.

There's sort of a precedent for this. `pkg:golang` doesn't really have namespace and name and you get the same kind of thing where the PURL "namespace" is just the part of the Go name before the first slash, but the Go module system enforces case sensitivity even when GitHub (or wherever the package is hosted) allows the files to be retrieved with different case and that avoids having to write custom normalization rules. Some people don't like `pkg:golang` and want to redo it with everything being in PURL name and nothing in the PURL namespace, in which case every slash would need to be percent encoded.

nicorikken: I see, looking at Go test cases https://github.com/package-url/purl-spec/blob/505dca561f6d6f1f1f0ebb6b5c36c6aa2516d98d/tests/types/golang-test.json
So in this case most of it would be the name, would have to be url encoded or the base purl format would need to be changed. Leveraging the subpath doesn't seem fitting.

PR #723, opened 2025-10-28
FIX typo in specification-test.json

giterlizzi:
The test fails because a "comma" is missing.

```
t/99-purl-tests.t ............... 264/? Bailout called.  Further testing stopped:  t/purl/spec/specification-test.json - , or } expected while parsing object/hash, at character offset 6623 (before ""description": "Buil...") at t/99-purl-tests.t line 42.
t/99-purl-tests.t ............... Dubious, test returned 255 (wstat 65280, 0xff00)
```

### DIFF
--- a/tests/spec/specification-test.json
+++ b/tests/spec/specification-test.json
@@ -203,7 +203,7 @@
       "input": "pkg%3Amaven/org.apache.commons/io",
       "expected_output": null,
       "expected_failure": true,
-      "expected_failure_reason": "Should fail to parse a PURL from invalid purl input"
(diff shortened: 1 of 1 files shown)


johnmhoran (approved):
Good catch @giterlizzi -- tailor-made to fix the error I encountered just now when trying to use `specification-test.json` -- `json.decoder.JSONDecodeError: Expecting ',' delimiter: line 207 column 7 (char 6623)`.  :+1:

mjherzog (approved):
Looks good

Issue #16, opened 2017-11-29
Link relations?


jackfirth:
PURLs seem to imply not just one resource, but a family of related resources:

- *Package name* - some name in some package registry of some package ecosystem
- *Package version* - some specific version of a package name, defined according to some sort of versioning system
- *Package registry* - some namespace of packages within some package ecosystem, typically with the authority to handle requests for information about and installation of those packages
- *Package ecosystem* - some existing system and protocol for depending on, publishing, and installing packages, such as "all Python packages installable with `pip`"
- *Package versioning system* - some system for minting structured version identifiers that allows clients to reason about the relationships between different versions in the same system (e.g. semver, git commits, docker registry labels, etc)
- *Package variants* - some alternate form of a specific package version which clients with specific needs may opt-in to

It seems within the PURL spec's scope to define these terms and standardize ways for a package registry to say something like "these are my supported versioning systems". This can be done by defining [RFC 8288 Link Relations](https://tools.ietf.org/html/rfc8288), and indeed that's the preferred means of doing so. There are many standardized ways in existing protocols for resources to declare relations to other resources dynamically (e.g. the HTTP Link header), so you'd be able to leverage a lot more existing work while making a simpler and more flexible spec.

pombredanne:
@jackfirth would these links be something that is set in an HTTP header for instance?

jackfirth:
Yes, particularly the HTTP `Link` header. Many other standard headers also use links. For example, using links in HTTP's standard authorization headers is how OAuth works.

pombredanne:
Ok, so then these cannot be stuffed in a single string?

jackfirth:
I'm not sure what you mean. There's no rule saying a URI scheme can't define that some link relations are implied by the definition of the scheme. Link relations are more abstract than that; they don't require a particular serialization format.

pombredanne:
My point is that if this does lives in the purl string, then I am fine to add that to the spec, but that does not have much impact overall. I thought these would be made part of the a `purl` string

pombredanne:
ping: is this something that would still need something added to the spec?

pombredanne:
Closing for lack of activity. You can reopen if you want to restart this discussion!

PR #724, opened 2025-10-29
Update version references, split merged test objects #16

johnmhoran:
- Changed 'purl-test.schema-1.0.json' to 'purl-test.schema-0.1.json' in 4 files
- Split merged test objects in specification-test.json at line 207 Reference: https://github.com/package-url/packageurl.org/issues/16

@giterlizzi @mjherzog Following up on the recently-merged https://github.com/package-url/purl-spec/pull/723, when I took a look at the updated `specification-test.json` file to update some .json version references, I noticed that the point at which we just added a comma (that was lines 206/07) was actually the point at which two distinct test objects somehow were merged together.


mjherzog (approved):
These corrections look good

Issue #729, opened 2025-11-04
Fix Bazel tests


TG1999:
https://github.com/package-url/purl-spec/blob/c9f3446c8919b0948513ead17e0ba9993aafc32e/tests/types/bazel-test.json#L50

The qualifier in the PURL here does not follow "character encoding" correctly.

mjherzog:
This has been fixed.

PR #730, opened 2025-11-04
Fix bazel tests

TG1999:
Reference:
- https://github.com/package-url/purl-spec/issues/729

Related:
- #644

### DIFF
--- a/ADOPTERS.md
+++ b/ADOPTERS.md
@@ -9,14 +9,16 @@
 - [GitHub Dependency Submission API](https://docs.github.com/en/rest/dependency-graph/dependency-submission): allows third-party tools
   to submit dependency data to GitHub for inclusion in a repository's dependency graph.
 - [NTT ZEN SecDB](https://secdb.nttzen.cloud): A security intelligence platform providing unified access to vulnerabilities, advisories, and exploit data across ecosystems, leveraging PURL for consistent package identification.
-- [OWASP Dependency-Track](https://www.owasp.org/index.php/OWASP_Dependency_Track_Project):
-  Open source component analysis platform
+- [OpenVEX Specification](https://github.com/openvex/spec/blob/main/OPENVEX-SPEC.md#appendix-b-software-identifier-types-table): A machine-readable format to communicate
+  the status of vulnerabilities affecting products and projects.
 - [OSS Index](https://ossindex.sonatype.org): A free catalog of Open Source
   Components and scanning tools to help developers identify vulnerable components
 - [OSS Review Toolkit](https://oss-review-toolkit.org/): A suite of tools to
   assist with automating Open Source compliance checks.
 - [OSV Schema](https://ossf.github.io/osv-schema/) and [OSV.dev](https://osv.dev):
   Open Source Vulnerability Schema and distributed vulnerability database
+- [OWASP Dependency-Track](https://www.owasp.org/index.php/OWASP_Dependency_Track_Project):
+  Open source component analysis platform
 - [ReversingLabs Spectra Assure](https://www.reversinglabs.com/products/software-supply-chain-security):
   Software supply chain security platform for the enterprise to detect threats and exposures
 - [Scancode Toolkit](https://github.com/nexB/scancode-toolkit): Reports
--- a/README.md
+++ b/README.md
@@ -1,7 +1,7 @@
 # Context
 
 We build and release software by massively consuming and producing software
-packages such as NPMs, RPMs, Rubygems, etc.
+packages such as npm packages, RPMs, Ruby gems, etc.
 
 Each package manager, platform, type or ecosystem has its own conventions and
 protocols to identify, locate and provision software packages.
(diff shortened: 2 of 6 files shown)


johnmhoran:
@TG1999 One preliminary question I have: if the input PURL has an unencoded slash in the `qualifiers` value, doesn't that raise an exception, halt the evaluation process, and thus require a `null` `expected_output` and a `true` `expected_failure`?

@mjherzog @pombredanne @matt-phylum @jkowalleck What do you think?

matt-phylum:
If the input PURL has an unencoded slash, an exception should not be thrown. It is not part of [the spec](https://github.com/package-url/purl-spec/blob/3e27e9249d4bf6e9396ab99f4f8374787e430b20/purl-specification.md?plain=1#L502-L515), and it breaks existing PURLs where slash is not encoded, including examples from previous versions of the spec and output from code that implements an older version of the spec.

When I last updated the purl-survey output, 10 of 15 implementations passed tests that required slash to be _unencoded_ in qualifier values. PURLs canonicalized by those implementations would be broken by turning unencoded slashes in qualifiers into errors.

johnmhoran:
Thanks @matt-phylum .  Putting aside slashes for a moment (which the submitted standard makes clear MUST be percent-encoded in a  qualifiers value  however we're meant to distinguish between MUST and SHOULD and MAY), if a `roundtrip` input has no `pkg:`, `expected_output` will be `null` and `expected_failure` will be `true` -- right?  Surely a tool will not "fix" that deficiency by supplying the missing `pkg:`.

tests/types/bazel-test.json
```diff
@@ -6,7 +6,7 @@
       "test_group": "base",
       "test_type": "roundtrip",
       "input": "pkg:bazel/rules_java@7.8.0?repository_url=https://bcr.bazel.build/",
-      "expected_output": "pkg:bazel/rules_java@7.8.0",
+      "expected_output": "pkg:bazel/rules_java@7.8.0?repository_url=https:%2F%2Fbcr.bazel.build%2F",
```
jkowalleck, line 9: while fixing this for bazel explicitely,
we should have this case become a general spec case in `specification-test.json`
- encode reserved delimiters `/?&#`

PS:see https://github.com/package-url/purl-spec/pull/738

johnmhoran: If a base test input is not canonical, like this one ^, doesn't that fail, period?

jkowalleck: nope.
non-canonical inputs must be allowed.

this case is not about non-canonical input, but about invalid input.
 as a purl must not contain non-alphanumeric chars -> see <https://github.com/package-url/purl-spec/blob/main/purl-specification.md#permitted-characters>

anyway, we COULD "fix" this invalid input by properly encoding the forbidden chars. 
ala "be conservative in what you do, be liberal in what you accept from others" https://en.wikipedia.org/wiki/Robustness_principle

mjherzog: @jkowalleck This is the intended distinction between base and advanced test groups. We need a suite of base conformance tests that are pass/fail. You allow a non-canonical input but it may fail.
The advanced tests include those where you might fix an error. 
In both cases we are in active discussion about what messages to return along with the pass/fail status.
- In the base case it would be useful for the tool to return a specific message about what caused the error as possible.
- In the advanced case it would be be useful for the tool to return a specific message about what it fixed.
So in this specific test case the test_group should be advanced instead of base imho.

petergardfjall:
Anything preventing this from being merged?

It is similar in spirit to #713, #714, #715, #716, #717, #718, #719, #720, #721

shibumi (approved):

tests/types/bazel-test.json
```diff
@@ -47,7 +49,7 @@
       "test_group": "base",
       "test_type": "roundtrip",
       "input": "pkg:bazel/rules_java@7.8.0?repository_url=https://example.org/bazel-registry/",
-      "expected_output": "pkg:bazel/rules_java@7.8.0?repository_url=https://example.org/bazel-registry",
```
mjherzog, line 50: If the input is transformed to correct it then this needs to be "test_group": "advanced"

tests/types/bazel-test.json
```diff
@@ -83,7 +85,7 @@
         },
         "subpath": null
       },
-      "expected_output": "pkg:bazel/rules_java@7.8.0?repository_url=https://example.org/bazel-registry",
+      "expected_output": "pkg:bazel/rules_java@7.8.0?repository_url=https:%2F%2Fexample.org%2Fbazel-registry",
```
mjherzog, line 88: If the input is transformed to correct it then this needs to be "test_group": "advanced"

tests/types/bazel-test.json
```diff
@@ -47,7 +49,7 @@
       "test_group": "base",
```
pombredanne:
```suggestion
      "test_group": "advanced,
```

tests/types/bazel-test.json
pombredanne:
```suggestion
      "test_group": "advanced",
```

tests/types/bazel-test.json
pombredanne: Marking this resolved!

pombredanne (commented):
Looking good now... I added a few touch ups based on @mjherzog reviews.

PR #735, opened 2025-11-04
remove 'invalid cran purl without version' testcases

petergardfjall:
Addresses https://github.com/package-url/purl-spec/issues/736

The `version` component is optional for `pkg:cran` [according to spec](https://github.com/package-url/purl-spec/blob/3e27e9249d4bf6e9396ab99f4f8374787e430b20/types-doc/cran-definition.md?plain=1#L35). Therefore it makes no sense to have tests requiring parsing to fail when the `version` is missing.


petergardfjall:
Anything preventing this from being merged?

shibumi (approved):

pombredanne:
Thanks. Technically, the version is always optional for all types. Good catch. Merging!

PR #738, opened 2025-11-05
tests: more cases for qualifier spec

jkowalleck:
caused by <https://github.com/package-url/purl-spec/pull/730/files#r2493378205> and others


jkowalleck:
> My understanding is that we will not be using the `roundtrip` `test_type`, so those tests will need to be deleted.

why is this? I mean, the roundtrip exists for a very good reason: test canonicalization of string input.

mjherzog:
@jkowalleck We have been trying to figure out the use case for roundtrip - no decision has been made.
We have a discussion document on our shared gdrive at: https://docs.google.com/document/d/1dmUq9reiw8SkXCr8PDtV1dw6wdgYWzULzyhiVoq8oLs/edit?usp=drive_link to organize current topics around PURL testing and welcome your comments. I know that we want to track the resolution in Issues rather than a gdoc but Issues are not great for big topics like this.

Philippe described roundtrip as a shortcut for combining a set of parse and build tests. Could you explain more about the use case for roundtrip? Is it somehow equivalent to "validation" of a string input? 
Perhaps the problem is related to the current description of a roundtrip test type in the schema: "  "roundtrip": "A PURL roundtrip test, parsing then building back a PURL from a canonical string input." where the input should not be defined as canonical, but as just an ASCII string.

jkowalleck:
> Philippe described roundtrip as a shortcut for combining a set of parse and build tests. Could you explain more about the use case for roundtrip? Is it somehow equivalent to "validation" of a string input?

I used it as "i get some PURL string input, and want to re-usue it with all normalizations and cleaning applied." 
For example, qualifiers are sorted, qualifiers with empty values are removed, all non-alphanumeric are percent-encoded, etc ... I dont care whether the implementation needs to parse and build the purl for this, or does it in another way.

PR #739, opened 2025-11-05
Fix typos

vladh:
* We can talk about an “npm package”, but not “an NPM”, since “npm” is the package manager. Similarly, “RubyGems” is a service, but the individual built packages are “Ruby gems”.
* The link at the bottom had garbled markdown.


pombredanne:
This are minor typos, so I am merging as-is

PR #740, opened 2025-11-06
Add new PURL type: 'fusesoc'

anders-ahlberg:
This change adds a type for FuseSoC cores (see https://github.com/olofk/fusesoc).


johnmhoran:
Putting aside the fact that we have not yet defined how the various test types are to behave (parse, build, roundtrip, base, advanced and so on), the test file seems to comprise 4 groups of parse, roundtrip and build test cases that are identical but for variations in namespace, name and version values.  Why?  A few tests meant to fail (there are none here) would be more useful and informative.

anders-ahlberg:
Oh, I was just copying the format of the tests in some other type already present, taking that for an appropriate level of effort.  I'll see about adding some more.

tests/types/fusesoc-test.json
```diff
@@ -0,0 +1,318 @@
...
+      "description": "Roundtrip test for PURL type: fusesoc",
+      "test_group": "base",
+      "test_type": "roundtrip",
+      "input": "pkg:fusesoc/open-logic/open-logic/base@3.3.0",
+      "expected_output": "pkg:fusesoc/open-logic/open-logic/base@3.3.0",
+      "expected_failure": false,
+      "expected_failure_reason": null
+    },
+    {
+      "description": "Build test  for PURL type: fusesoc",
```
pombredanne: There may be an extra space there?

tests/types/fusesoc-test.json
```diff
@@ -0,0 +1,318 @@
...
+      "description": "Roundtrip test for PURL type: fusesoc",
+      "test_group": "base",
+      "test_type": "roundtrip",
+      "input": "pkg:fusesoc/fusesoc/utils/blinky@1.1.0",
+      "expected_output": "pkg:fusesoc/fusesoc/utils/blinky@1.1.0",
+      "expected_failure": false,
+      "expected_failure_reason": null
+    },
+    {
+      "description": "Build test  for PURL type: fusesoc",
```
pombredanne: Extra space in the description

types/fusesoc-definition.json
```diff
@@ -0,0 +1,33 @@
...
+  },
+  "name_definition": {
+    "requirement": "required",
+    "native_name": "name",
+    "case_sensitive": true
+  },
+  "version_definition": {
+    "requirement": "optional",
+    "native_name": "version",
+    "note": "The version is the semantic version of the fusesoc core.  We would really like this to be required, but that breaks validation."
```
pombredanne: I am not sure you need this comment.... The fact that version is optional is a spec thing when you want to refer to a package without mentioning a version like in the CVE schema, or when using that with a VERS. What about this:

```suggestion
    "note": "The version is the semantic version of the fusesoc core."
```

pombredanne (changes requested):
Thanks. Here are a few nits for your review... also I am not sure that the encoding of the namespace / matches the spec... can you double check?

PR #747, opened 2025-11-14
discussion-draft for tests.md

mjherzog:
This PR covers a rewrite of docs/tests.md to cover key definitions including:
- test groups
- test types
- error handling
- test output messages

### DIFF
--- a/docs/tests.md
+++ b/docs/tests.md
@@ -18,9 +18,11 @@ for a specific PURL type.
   that primarily cover testing the validity of individual PURL components or
   separators between a pair of PURL components.
   - component-`test.json`: These are test files for a specific PURL component.
+ See https://github.com/package-url/purl-spec/pull/738 which adds 
+ `tests/spec/qualifiers.json`.
 - `purl-spec/tests/types/`: This folder contains one JSON test file for each 
-registered PURL type. These tests primarily cover the validity of a complete 
-PURL for the corresponding PURL type.
+registered PURL type. These tests should be focused on test cases that are
+specific to a PURL type, such as those for namespace or qualifiers.
 
 Two key properties in the PURL test JSON schema are:
 
@@ -43,7 +45,8 @@ There are four PURL test types:
 decoded PURL components. See also `/docs/how-build.md`.
 - **parse**: A test to parse decoded components from a canonical PURL 
 input string. See also `/docs/how-parse.md`.
-- **roundtrip**: A test to parse an input PURL string and then rebuild it as a canonical PURL output string.
+- **roundtrip**: A test to parse an input PURL string and then rebuild it as a
+ canonical PURL output string.
 - **validation**: A test to validate a PURL input string and report severity 
 messages - info, warning or error. A validation test may optionally correct 
 errors in an output PURL string.
@@ -52,6 +55,10 @@ errors in an output PURL string.
 - *The validation test type is currently a proposed PURL test schema change. 
 See https://github.com/package-url/purl-spec/pull/614.*
 - *Do we need both roundtrip and validation test types?*
+- *Should the proposed validation test type be split into two test types for:*
+   - *validation: message output*
+   - *remediation: corrects errors and also provides messages documenting the 
+   corrections*
 - *For tests in the advanced test group we may want to distinguish between 
 validation test types that return only a validation message and a remediation
 test type which corrects an error or errors and also returns a message (see 
@@ -64,11 +71,11 @@ below) that explains the remediation.*
       "input": "pkg:PYPI/Django_package@1.11.1.dev1",\
       "expected_output": "pkg:pypi/django-package@1.11.1.dev1",\
       "expected_failure": false,\
-      "expected_failure_reason": null`
+      "expected_failure_reason": null
 
   *This test case is arguably an example of an advanced remediation test case 
   because it corrects two errors. An advanced validation test case should 
-  arguably return two error messages such as:
+  arguably return two error messages such as:*
     - *The pkg component must be lowercased.*
     - *The name component must be lowercased.*
 
@@ -79,7 +86,7 @@ below) that explains the remediation.*
 
 The standard error-handling behaviour for all test cases is based on two
 properties from the PURL test schema:
-- `expected_failure` 
+- `expected_failure`
   - description: "true if this test input is expected to fail to be 
 processed."
   - type: boolean
@@ -106,10 +113,13 @@ validation severity messages:*
    - *warning: "Warning validation message*
    - *error: "Error validation message"*
 
+  *We will need to clearly define the difference between an info message and 
+  a warning message.*
+  
   *NB - there is no message proposed for a successful test.*
 
-- *We should consider adding the message levels for all test types, not just
-the validation test type.*
+- *We should add message levels for all test types, not just the validation 
+test type.*
 
 
 


docs/tests.md
```diff
@@ -1,33 +1,127 @@
...
+better message could be:"The name component for a PyPI package is not case 
+senstive".*
+- *The PR to add a validation test type (https://github.com/package-url/purl-spec/pull/614) proposes a new `purl validation message` property with three levels of 
+validation severity messages:* 
+   - *info: "Informational validation message"*
+   - *warning: "Warning validation message*
+   - *error: "Error validation message"*
+
+  *We will need to clearly define the difference between an info message and 
+  a warning message.*
```
matt-phylum, line 117: Do any PURL implementations support info or warning messages? Do any consumers check those messages? Are there meaningful messages to generate? Implementations would need to either store these messages on the parsed PURL or return a wrapper that includes these messages.

docs/tests.md
```diff
@@ -1,33 +1,127 @@
...
+examples a tool could usually derive an error message from the description, 
+but the message would be indirect. For example from the first test case the 
+description is: "pypi names have special rules and not case sensitive". A 
+better message could be:"The name component for a PyPI package is not case 
+senstive".*
+- *The PR to add a validation test type (https://github.com/package-url/purl-spec/pull/614) proposes a new `purl validation message` property with three levels of 
+validation severity messages:* 
+   - *info: "Informational validation message"*
+   - *warning: "Warning validation message*
+   - *error: "Error validation message"*
```
matt-phylum, line 114: I don't think it is necessary or beneficial for PURL to specify error message strings to be produced by implementations. It seems potentially reasonable to have an explanation of why the test fails, but specifying the errors produced in this way prohibits or discourages implementations from handling errors in better or more customary ways. Implementations languages that throw exceptions or return typed results/outcomes should return typed errors, ie a syntactically invalid PURL and a PURL that fails package-type-specific validation should result in different types or enum values such that the consuming software doesn't need to analyze a string to determine what's going on.

It's possible to specify that there is a conversion from however the implementation specifies errors into a string, but this seems more useful for the purposes of this test case than in actual usage.

docs/tests.md
```diff
@@ -1,33 +1,127 @@
...
+
+*Discussion points:* 
+- *The current test schema does not include any specific properties for error 
+or other test result messages. Most of the test cases with an expected 
+failure are in the file `test/spec/specification-test.json`. Based on those 
+examples a tool could usually derive an error message from the description, 
+but the message would be indirect. For example from the first test case the 
+description is: "pypi names have special rules and not case sensitive". A 
+better message could be:"The name component for a PyPI package is not case 
+senstive".*
```
matt-phylum, line 109: There are special rules. PURL actually has the wrong rules (#262), but the most important part of the provided example test is checking that the underscore is converted into a hyphen.

docs/tests.md
```diff
@@ -1,33 +1,127 @@
...
+**Test types**
+
+There are four PURL test types:
+- **build**: A test to build a canonical PURL output string from an input of 
+decoded PURL components. See also `/docs/how-build.md`.
+- **parse**: A test to parse decoded components from a canonical PURL 
+input string. See also `/docs/how-parse.md`.
+- **roundtrip**: A test to parse an input PURL string and then rebuild it as a
+ canonical PURL output string.
+- **validation**: A test to validate a PURL input string and report severity
```
matt-phylum, line 50: What is the difference between a validation test and a parsing test? They both parse a PURL and they both check that the expected values are produced.

docs/tests.md
```diff
@@ -1,33 +1,127 @@
...
+*Discussion point:*
+- *Is there a better name than advanced for the second test
+group? perhaps Permissive? Or name the groups Strict and Permissive?]*
+
+**Test types**
+
+There are four PURL test types:
+- **build**: A test to build a canonical PURL output string from an input of 
+decoded PURL components. See also `/docs/how-build.md`.
+- **parse**: A test to parse decoded components from a canonical PURL
```
matt-phylum, line 46: Parse tests historically do not need their inputs to be canonical.

matt-phylum, line 46: The tests expect something that AFAIK is not specified anywhere. Implementations are expected to canonicalize during parsing and during building. If an implementation does not do this automatically (at least one implementation intentionally doesn't) then the test runner needs to do that or it will fail the tests. It's kind of obvious once you start looking at the test failures, but it would probably be good to specify that the implementation is expected to do that.

mjherzog: Added: 
"Any PURL implementation tool is expected to canonicalize a PURL string or PURL components during a parsing or building operation." at the end of the intro section.

docs/tests.md
mjherzog: The output from a parsing test is the set of decoded PURL components.
The output from a validation test (as currently proposed) is a PURL string.

docs/tests.md
mjherzog: Updated to: "The name component for a PyPI package is not case sensitive and requires changing a consecutive dash, underscore or dot character to a single dash".

docs/tests.md
mjherzog: This discussion section is not based on what current PURL implementations do. The proposed set of messages are from #614 which is on hold pending the conclusion of this discussion.
I originally used italics to call out open discussion sections, but that does not seem so clear so I switched to using a blockquote to mark the beginning and end of a discussion section.

docs/tests.md
mjherzog: Agreed and not sure why they should be. 
Updated description of parse test type to: "A test to parse decoded components from a PURL input string."
from: "A test to parse decoded components from a canonical PURL input string."

docs/tests.md
mjherzog: The idea of adding these messages was introduced with #614 which is pending the conclusion of this discussion. 
The messages are not intended to replace the `expected_failure_reason`, but to provide additional information about the detected problem.
But the current approach for `expected_failure_reason ` is pretty generic. For example the first set of test cases in `tests/spec/specification-test.json `have a pattern with a specific description but a generic failure reason, such as;
- `description`: "a scheme is always required", and `expected_failure_reason`: "Should fail to parse a PURL from invalid purl input".
Wouldn't it be more helpful to have a more specific failure reason like: "Should fail to parse a PURL with scheme component missing from the PURL input" ?

mjherzog:
@matt-phylum Thank you for your comments - see my replies above. I am still learning how to use GH PRs and Issues to manage a discussion document (instead of a gdoc with comments).

docs/tests.md
matt-phylum: The output from a building test is also a PURL string. PURL implementations generally have two operations: they can turn a PURL string into its components, and they can turn components into PURL strings. I can't think of something that round trip or validation tests could do that parsing or building tests can't already do. There's already some discussion about getting rid of round trip tests, and I wonder if this is another thing that will later be getting removed.

mjherzog: @matt-phylum What do you suggest we call the proposed validation test type? Do you recommend 2 sub-types for a parse test where one returns the components and another returns a string?

matt-phylum: I don't understand why there would be a parse test that returns a string. If the parse is correct, the components will be correct. Once the input has been parsed, there is no memory of what it was before being parsed, so a test that parsing `pkg:pypi/Django_package` results in `pkg:pypi/django-package` is no different from a test that parsing `pkg:pypi/Django_package` results in `{"package_type": "pypi", "name": "django-package"}` and that building `{"package_type": "pypi", "name": "django-package"}`, unsurprisingly, results in `pkg:pypi/django-package`.

mjherzog: The apparent difference is in the expected test output. For the current parse test type the expected output is a set of decoded PURL components. For the proposed validation test type the expected output is a PURL string. I can see how you can get the string output by combining a parse test and a build test - are you recommending that we should not have that combination as a test type and leave the combination processing to the tool/implementation?

matt-phylum: Unless there is an intention that these tests are only used by implementations that have a dedicated PURL string -> canonicalized PURL string routine that isn't implemented by parsing and building, I don't think it makes sense to have a test that parse+build for a given input string produces a specific output string. It's easier for implementers to test that both parsing and building work as expected and it avoids bugs where canonicalization is applied differently when parsing or building.

mjherzog: @matt-phylum Thank you. Your perspective is clear to me now.
Jan has separately said that: "... the roundtrip exists for a very good reason: test canonicalization of string input." 
See https://github.com/package-url/purl-spec/pull/738.
I will put this topic on the agenda for our TG2 meeting this Friday. I will post a new PR (including your input) with consolidated updates because the commentary is a bit hard to follow as PR changes in the markdown file.

docs/tests.md
```diff
@@ -1,33 +1,127 @@
...
-  components should return the test canonical `purl`
-
-- building a `purl` from the test components should return the test canonical
-  `purl`
+The Package-URL (PURL) specification provides a JSON Schema and test
+files to support language-neutral testing of PURL implementations. The 
+objectives for the PURL schema and test files are to enable tools to conform
+to the PURL specification for tool functions such as:
+- build a canonical PURL string from a set of PURL component-level data
+- parse a canonical PURL string into a set of PURL components
```
jkowalleck, line 8:
```suggestion
- parse a PURL string into a set of PURL components
```


tests include parsing valid/invalid PURLs
tests include parsing canonical/non-canonical PURLs

docs/tests.md
```diff
@@ -1,33 +1,127 @@
...
+- validate an input PURL string and optionally provide warning or info
```
jkowalleck, line 9:
```suggestion
- validate an input and optionally provide warning or info 
```

tests include inputs as string
as well as inputs as components

docs/tests.md
```diff
@@ -1,33 +1,127 @@
...
+The Package-URL (PURL) specification provides a JSON Schema and test
+files to support language-neutral testing of PURL implementations. The 
+objectives for the PURL schema and test files are to enable tools to conform
+to the PURL specification for tool functions such as:
+- build a canonical PURL string from a set of PURL component-level data
+- parse a canonical PURL string into a set of PURL components
+- validate an input PURL string and optionally provide warning or info 
+messages or correct errors in the input PURL string
+
+The current  JSON schema is available at: `purl-spec/schemas/purl-test.schema-0.1.json`.
```
jkowalleck, line 12: link the documents you reference.

```suggestion
The current  JSON schema is available at: [`purl-spec/schemas/purl-test.schema-0.1.json`](../schemas/purl-test.schema-0.1.json).
```

docs/tests.md
```diff
@@ -1,33 +1,127 @@
...
+
+The test files are available at:
+- `purl-spec/tests/spec/`: This folder contains JSON test files that are not
```
jkowalleck, line 15: link the things you reference
```suggestion
- [`purl-spec/tests/spec/`](../tests/spec/): This folder contains JSON test files that are not
```

docs/tests.md
```diff
@@ -1,33 +1,127 @@
...
+for a specific PURL type.
+  - `specification-test.json` - This file contains an array of test objects
```
jkowalleck, line 17: link the things you reference

docs/tests.md
```diff
@@ -1,33 +1,127 @@
...
+  that primarily cover testing the validity of individual PURL components or
+  separators between a pair of PURL components.
+  - component-`test.json`: These are test files for a specific PURL component.
```
jkowalleck, line 20: :+1: like the idea of a naming convention

```suggestion
  - `<component>-test.json`: These are test files for a specific PURL component.
```

docs/tests.md
```diff
@@ -1,33 +1,127 @@
...
+ See https://github.com/package-url/purl-spec/pull/738 which adds 
+ `tests/spec/qualifiers.json`.
+- `purl-spec/tests/types/`: This folder contains one JSON test file for each
```
jkowalleck, line 23: why not also add a naming convention for these, too? 
currently it looks like `<type>-test.json`

docs/tests.md
```diff
@@ -1,33 +1,127 @@
...
+
+The test files are available at:
+- `purl-spec/tests/spec/`: This folder contains JSON test files that are not
+for a specific PURL type.
+  - `specification-test.json` - This file contains an array of test objects
+  that primarily cover testing the validity of individual PURL components or
+  separators between a pair of PURL components.
+  - component-`test.json`: These are test files for a specific PURL component.
+ See https://github.com/package-url/purl-spec/pull/738 which adds 
+ `tests/spec/qualifiers.json`.
```
jkowalleck, line 22: after reading your paragraph regarding naming schema, i renamed the file in the PR

```suggestion
 `tests/spec/qualifiers-test.json`.
```

docs/tests.md
jkowalleck, line 46:
```suggestion
- **parse**: A test to parse decoded components from a PURL 
```

docs/tests.md
```diff
@@ -1,33 +1,127 @@
...
+- *Is there a better name than advanced for the second test
+group? perhaps Permissive? Or name the groups Strict and Permissive?]*
+
+**Test types**
+
+There are four PURL test types:
+- **build**: A test to build a canonical PURL output string from an input of 
+decoded PURL components. See also `/docs/how-build.md`.
+- **parse**: A test to parse decoded components from a canonical PURL 
+input string. See also `/docs/how-parse.md`.
```
jkowalleck, line 47: link the document you reference to

docs/tests.md
```diff
@@ -1,33 +1,127 @@
...
+- **roundtrip**: A test to parse an input PURL string and then rebuild it as a
```
jkowalleck, line 48: we dont care how the process works internally.
as long as the result is a validated, canonical PURL.

```suggestion
- **roundtrip**: A test to ingest a PURL string and then issue a
```

docs/tests.md
```diff
@@ -1,33 +1,127 @@
...
+
+**Test case basics**
+
+The standard error-handling behaviour for all test cases is based on two
+properties from the PURL test schema:
+- `expected_failure`
+  - description: "true if this test input is expected to fail to be 
+processed."
+  - type: boolean
+  - default: false
```
jkowalleck, line 93:
```suggestion
  - default: `false`
```

docs/tests.md
```diff
@@ -1,33 +1,127 @@
...
+
+- `expected_failure_reason`: "The reason why this test is is expected to fail 
+if expected_failure is true."
+   - default: null
```
jkowalleck, line 97:
```suggestion
   - default: `null`
```

docs/tests.md
```diff
@@ -1,33 +1,127 @@
...
+
+In the case of a test case failure the `expected_output` is null.
```
jkowalleck, line 99:
```suggestion
In the case of a test case failure the `expected_output` is `null`.
```

jkowalleck (changes requested):

mjherzog:
Closed in favor of 
- #752

PR #757, opened 2025-12-03
Add OpenVEX

schlomo:
and fix alphabetical sorting for OWASP


mjherzog:
@schlomo Thank you for the PR. Just FYI - we are working on a Package-URL website for release no later than year-end where the specs and tools will be displayed in a grid rather like CycloneDX Tool Center, but we want to keep ADOPTERS.md current in the meantime.

schlomo:
Thanks, maybe also good to link somewhere to articles like https://fossa.com/blog/understanding-purl-specification-package-url/ that I found useful to read for understanding purl

PR #777, opened 2025-12-30
Ecma format corrections 2

mjherzog:
A few more corrections - most are for known-qualifiers to format field values (single quotes) and examples (double indent)

### DIFF
--- a/docs/known-qualifiers.md
+++ b/docs/known-qualifiers.md
@@ -13,17 +13,16 @@ With this warning, the known **key** and **value** defined here are valid for
 use in all package types:
 
 - VERS allows the specification of a version range.
-  The value must adhere to the [Version Range Specification](https://github.com/package-url/vers-spec/blob/main/VERSION-RANGE-SPEC.md).
+  The **value** must adhere to the [Version Range Specification](https://github.com/package-url/vers-spec/blob/main/VERSION-RANGE-SPEC.md).
   This qualifier is mutually exclusive with the **version** component.
   For example:
 
       pkg:pypi/django?vers=vers:pypi%2F%3E%3D1.11.0%7C%21%3D1.11.1%7C%3C2.0.0
 
 - 'repository_url' is an extra URL for an alternative, non-default package
-  repository or registry. When a package does not come from the default
-  public package repository for its **type** a PURL may be qualified with
-  this extra URL. The default repository or registry of a **type** is
-  documented in the "Registered PURL types" section.
+repository or registry. When a package does not come from the 
+'default_repository_url' for its **type** a PURL may be qualified with this 
+extra URL.
 
 - 'download_url' is an extra URL for a direct package web download URL to
   optionally qualify a PURL .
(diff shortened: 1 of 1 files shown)


docs/known-qualifiers.md
```diff
@@ -5,40 +5,41 @@ keys but their usage should be limited to the bare minimum for proper package
...
 download URL, VCS URL or checksums in an API, database or web form.
 
-With this warning, the known **key** and **value**` defined here are valid for
+With this warning, the known **key** and **value** defined here are valid for
 use in all package types:
 
 - VERS allows the specification of a version range.
-  The value must adhere to the `Version Range Specification`.
-  This qualifier is mutually exclusive with the `version` component.
+  The value must adhere to the [Version Range Specification](https://github.com/package-url/vers-spec/blob/main/VERSION-RANGE-SPEC.md).
```
johnmhoran: Should "value" be in double asterisks as it is in line 12 above?

docs/known-qualifiers.md
```diff
@@ -5,40 +5,41 @@ keys but their usage should be limited to the bare minimum for proper package
...
   in Python pip or the SPDX specification. See
   https://github.com/spdx/spdx-spec/blob/cfa1b9d08903/chapters/3-package-information.md#3.7
 
-- `file_name` is an extra file name of a package archive.
+- 'file_name' is an extra file name of a package archive.
 
-- **checksum** is a qualifier for one or more checksums stored as a
-  comma-separated list. Each item in the **value** is in form of
+- 'checksum' is a qualifier for one or more checksums stored as a
+  comma-separated list. Each item in the value is in form of
```
johnmhoran: Here, too, should "value" be in double asterisks/bold?

docs/known-qualifiers.md
```diff
@@ -5,40 +5,41 @@ keys but their usage should be limited to the bare minimum for proper package
...
+      pkg:pypi/django?vers=vers:pypi%2F%3E%3D1.11.0%7C%21%3D1.11.1%7C%3C2.0.0
 
-- `repository_url` is an extra URL for an alternative, non-default package
+- 'repository_url' is an extra URL for an alternative, non-default package
   repository or registry. When a package does not come from the default
-  public package repository for its **type** a PURL  may be qualified with
+  public package repository for its **type** a PURL may be qualified with
   this extra URL. The default repository or registry of a **type** is
-  documented in the "Registered PURL  types" section.
+  documented in the "Registered PURL types" section.
```
johnmhoran: I thought Ecma forbids double quotes.  No?

docs/known-qualifiers.md
mjherzog: yes - updated

docs/known-qualifiers.md
mjherzog: Ecma does allow some use of double quotes (e.g. a reference to "lowercase" in the the case-folding clause of the standard, but it is rare. In this case the reference to "Registered PURL types" section is an ambiguous holdover from the one big RST file so I rephrased to:

- 'repository_url' is an extra URL for an alternative, non-default package repository or registry. When a package does not come from the  'default_repository_url' for its **type** a PURL may be qualified with this extra URL.

johnmhoran: Thanks @mjherzog .

mjherzog:
Merging format (editorial) changes - no normative changes

Issue #778, opened 2026-01-06
Remove refs to "Registered types section" in how-to build/parse files


johnmhoran:
Remove the references to "Registered types section" -- inherited from the OG .rst -- in:

- https://github.com/package-url/purl-spec/blob/main/docs/how-to-build.md
- https://github.com/package-url/purl-spec/blob/main/docs/how-to-parse.md

mjherzog:
Closed with merge of 
- #779

PR #779, opened 2026-01-06
Remove references inherited from OG .rst #778

johnmhoran:
Reference: https://github.com/package-url/purl-spec/issues/778


johnmhoran:
@mjherzog Please review when you have the chance.  These two corrected files are among those from `/docs` that will find their way into a packageurl.org update in the very near future.

mjherzog (approved):
Looks good.

PR #768, opened 2025-12-10
add OCaml purl implementation to Tools

hannesm:

PR #787, opened 2026-01-23
add npm `purl` package

ljharb:


mjherzog:
@ljharb Hello Jordan,
We are no longer maintaining the list of Implementations in the README.

The current data is now shown in the Software Tools section of https://package-url.github.io/www.packageurl.org/ and we will link to that from the README in the near-future

We would be happy to add your tool there as ljharb/purl. My take on the data is:
```
Description: A PURL (Package URL) parser and serializer, implementing the [TC54 PURL specification](https://tc54.org/purl/).

Base language: JavaScript
Software License: MIT
Data License: n/a
Service License: n/a
Type: Library
Standards: PURL
Platform: n/a
Home: https://github.com/ljharb/purl
Source download: n/a
Package download: https://www.npmjs.com/package/purl
Notes: 
```
Please update the data as needed.

Regarding the change for the ECMA-427 license, I cannot see any change in your PR

ljharb:
ah ok, i saw #768 still open and assumed this list was still valid. will it be removed from the readme, then?

I'm not sure what you mean by "the change for the ECMA-427 license"; the spec itself isn't in the package, so its license isn't relevant, and the package's license is MIT.

mjherzog:
Yes - the Implementations list will be replaced by a website link next week.

Re the ECMA-427 license:
- Your proposed change appears to delete the line and then add it back
- The ECMA-427 license is relevant because it is the base line for the specification (docs and schemas) which is the primary content of this repo; the MIT license here is for the software which is essentially just a few internal-use scripts. MIT is the most common license for software implementations hosted under the package-url GH org but it is a poor license for a specification

ljharb:
oh, that's probably just an added trailing newline - i assume the web interface automatically adds it, since every file should always end with a trailing newline.

ljharb:
replaced with https://github.com/package-url/www.packageurl.org/pull/60

PR #791, opened 2026-01-26
Regenerate the schema-related files #57

johnmhoran:
Reference: https://github.com/package-url/www.packageurl.org/issues/57

Note: the branch had a 56- prefix but should have been 57-*.


mjherzog:
Update to add opam-definition.md and correct some other text errors. No normative changes

PR #578, opened 2025-08-06
spec: add grammar (ABNF)

jkowalleck:
This PR defines the PackageURL grammar using ABNF.  

- fixes #535 

ABNF is a standardized grammar notation defined in [RFC 5234](https://datatracker.ietf.org/doc/html/rfc5234) and widely used across internet protocol specifications (HTTP, SIP, email, URI syntax, etc.). Unlike BNF, which exists in many incompatible variants, ABNF has a single, formal, and stable specification.

Because PackageURL is conceptually similar to URLs, using ABNF aligns the grammar with established internet standards. The IETF URL/URI specification (RFC 3986) also uses ABNF to define its syntax, so adopting the same notation keeps PackageURL consistent with the broader ecosystem of identifier formats.

ABNF is a notation for describing grammars, not a parser‑generator language like ANTLR or yacc. It is tool‑agnostic: various libraries can parse ABNF and generate validators or parsers from it, but the standard itself does not prescribe any specific generator. This makes ABNF a well‑recognized, interoperable choice for defining a grammar intended to be implemented across multiple languages and platforms.

PR #793, opened 2026-01-30
Add 15 ECMA-427 edge case tests

### LINKED ISSUE #788: Missing Test Coverage in purl-spec Fixtures
Hello,

This document identifies test cases that should be added to the [purl-spec test suite](https://github.com/package-url/purl-spec) to ensure implementations correctly handle ECMA-427 edge cases.

These gaps were discovered during [PURL.jl](https://github.com/JuliaLang/SecurityAdvisories.jl/issues/145#issuecomment-3792203806) development. Without these tests, implementations may pass the official test suite while still violating the specification.

## Summary

| Issue | ECMA-427 Section | Missing Test Type |
|-------|------------------|-------------------|
| Scheme with slashes | 5.6.1 | Parse + roundtrip |
| Plus sign in type | 5.6.2 | Parse failure |
| Colon encoding | 5.4 | Roundtrip |
| Empty qualifier values | 5.6.6 | Parse + roundtrip |
| Namespace segment encoding | 5.6.3 | Build + roundtrip |

---

## 1. Scheme with Leading Slashes (Section 5.6.1)

**Specification requirement:** PURL parsers SHALL accept URLs where the scheme is followed by one or more slash characters and ignore them.

**Problem:** No test verifies that `pkg://type/name` parses equivalently to `pkg:type/name`.

### Suggested test cases

```json
{
  "description": "scheme followed by double slash should be accepted",
  "test_group": "base",
  "test_type": "parse",
  "input": "pkg://npm/foo@1.0.0",
  "expected_output": {
    "type": "npm",
    "namespace": null,
    "name": "foo",
    "version": "1.0.0",
    "qualifiers": null,
    "subpath": null,
    "canonical_purl": "pkg:npm/foo@1.0.0"
  },
  "expected_failure": false,
  "expected_failure_reason": null
},
{
  "description": "scheme followed by triple slash should be accepted",
  "test_group": "base",
  "test_type": "parse",
  "input": "pkg:///pypi/requests@2.28.0",
  "expected_output": {
    "type": "pypi",
    "namespace": null,
    "name": "requests",
    "version": "2.28.0",
    "qualifiers": null,
    "subpath": null,
    "canonical_purl": "pkg:pypi/requests@2.28.0"
  },
  "expected_failure": false,
  "expected_failure_reason": null
},
{
  "description": "scheme with slashes roundtrips to canonical form without slashes",
  "test_group": "base",
  "test_type": "roundtrip",
  "input": "pkg://generic/name@1.0",
  "expected_output": "pkg:generic/name@1.0",
  "expected_failure": false,
  "expected_failure_reason": null
}
```

---

## 2. Plus Sign Not Allowed in Type (Section 5.6.2)

**Specification requirement:** Type shall be composed only of ASCII letters, numbers, period `.`, and dash `-`.

**Problem:** No test verifies that plus sign `+` is rejected in the type field. Implementations may incorrectly accept types like `c++` or `objective-c++`.

### Suggested test cases

```json
{
  "description": "plus sign in type should be rejected",
  "test_group": "base",
  "test_type": "parse",
  "input": "pkg:c++/stdlib@1.0",
  "expected_output": null,
  "expected_failure": true,
  "expected_failure_reason": "Should fail to parse a PURL with plus sign in type"
},
{
  "description": "plus sign in type should be rejected (build)",
  "test_group": "base",
  "test_type": "build",
  "input": {
    "type": "objective-c++",
    "namespace": null,
    "name": "foundation",
    "version": "1.0",
    "qualifiers": null,
    "subpath": null
  },
  "expected_output": null,
  "expected_failure": true,
  "expected_failure_reason": "Should fail to build a PURL with plus sign in type"
}
```

**Note:** The existing test for `n&g?inx` catches ampersand and question mark but not plus sign.

---

## 3. Colons Shall Not Be Percent-Encoded (Section 5.4)

**Specification requirement:** The colon `:` character shall not be percent-encoded, whether used as a separator or otherwise.

**Problem:** No test verifies that colons in namespace, name, or version components remain unencoded in the canonical output.

### Suggested test cases

```json
{
  "description": "colon in name should not be percent-encoded",
  "test_group": "base",
  "test_type": "build",
  "input": {
    "type": "generic",
    "namespace": null,
    "name": "std:io",
    "version": "1.0",
    "qualifiers": null,
    "subpath": null
  },
  "expected_output": "pkg:generic/std:io@1.0",
  "expected_failure": false,
  "expected_failure_reason": null
},
{
  "description": "colon in namespace should not be percent-encoded",
  "test_group": "base",
  "test_type": "build",
  "input": {
    "type": "generic",
    "namespace": "org:example",
    "name": "lib",
    "version": null,
    "qualifiers": null,
    "subpath": null
  },
  "expected_output": "pkg:generic/org:example/lib",
  "expected_failure": false,
  "expected_failure_reason": null
},
{
  "description": "colon in version should not be percent-encoded",
  "test_group": "base",
  "test_type": "roundtrip",
  "input": "pkg:generic/foo@1.0:beta",
  "expected_output": "pkg:generic/foo@1.0:beta",
  "expected_failure": false,
  "expected_failure_reason": null
},
{
  "description": "encoded colon in name should decode and stay unencoded",
  "test_group": "base",
  "test_type": "roundtrip",
  "input": "pkg:generic/std%3Aio@1.0",
  "expected_output": "pkg:generic/std:io@1.0",
  "expected_failure": false,
  "expected_failure_reason": null
}
```

---

## 4. Empty Qualifier Values Shall Be Discarded (Section 5.6.6)

**Specification requirement:** A key=value pair with an empty value is the same as no key=value pair for that key.

**Problem:** No test verifies that empty qualifier values are discarded during parsing and not serialized in canonical output.

### Suggested test cases

```json
{
  "description": "qualifier with empty value should be discarded",
  "test_group": "base",
  "test_type": "parse",
  "input": "pkg:npm/foo@1.0?empty=&valid=yes",
  "expected_output": {
    "type": "npm",
    "namespace": null,
    "name": "foo",
    "version": "1.0",
    "qualifiers": {
      "valid": "yes"
    },
    "subpath": null,
    "canonical_purl": "pkg:npm/foo@1.0?valid=yes"
  },
  "expected_failure": false,
  "expected_failure_reason": null
},
{
  "description": "multiple empty qualifiers should all be discarded",
  "test_group": "base",
  "test_type": "roundtrip",
  "input": "pkg:npm/bar@2.0?a=&b=value&c=",
  "expected_output": "pkg:npm/bar@2.0?b=value",
  "expected_failure": false,
  "expected_failure_reason": null
},
{
  "description": "all-empty qualifiers should result in no query string",
  "test_group": "base",
  "test_type": "roundtrip",
  "input": "pkg:npm/baz@3.0?empty=&also_empty=",
  "expected_output": "pkg:npm/baz@3.0",
  "expected_failure": false,
  "expected_failure_reason": null
}
```

---

## 5. Namespace Segments Encoded Individually (Section 5.6.3)

**Specification requirement:** Each namespace segment shall be a percent-encoded string, with segments separated by unencoded `/`.

**Problem:** No test verifies that namespace segments are encoded individually with slashes remaining literal. Implementations might encode the entire namespace as a single string, incorrectly encoding internal `/` characters.

### Suggested test cases

```json
{
  "description": "namespace segments should preserve literal slashes",
  "test_group": "base",
  "test_type": "build",
  "input": {
    "type": "maven",
    "namespace": "org.apache/commons",
    "name": "lang",
    "version": "3.12",
    "qualifiers": null,
    "subpath": null
  },
  "expected_output": "pkg:maven/org.apache/commons/lang@3.12",
  "expected_failure": false,
  "expected_failure_reason": null
},
{
  "description": "special characters in namespace segments should be encoded per-segment",
  "test_group": "base",
  "test_type": "build",
  "input": {
    "type": "generic",
    "namespace": "org name/sub dir",
    "name": "lib",
    "version": null,
    "qualifiers": null,
    "subpath": null
  },
  "expected_output": "pkg:generic/org%20name/sub%20dir/lib",
  "expected_failure": false,
  "expected_failure_reason": null
},
{
  "description": "namespace with encoded slash in segment should roundtrip correctly",
  "test_group": "base",
  "test_type": "roundtrip",
  "input": "pkg:generic/a%2Fb/c/name",
  "expected_output": "pkg:generic/a%2Fb/c/name",
  "expected_failure": false,
  "expected_failure_reason": null
}
```

**Note:** This tests the distinction between `/` as segment separator vs `%2F` as literal slash within a segment.

---

## Recommendations for purl-spec Maintainers

1. **Add these test cases** to `tests/spec/specification-test.json` to catch common implementation errors.

2. **Consider a "strict compliance" test group** for edge cases that trip up implementations but aren't critical for basic interoperability.

3. **Document the encoding rules explicitly** in the test file comments, especially the colon non-encoding rule which is counter-intuitive for URL implementations.

4. **Add roundtrip tests for all normalization behaviors** to catch implementations that parse correctly but serialize incorrectly.

---

## References

- [ECMA-427 Specification](https://ecma-international.org/publications-and-standards/standards/ecma-427/)
- [purl-spec GitHub Repository](https://github.com/package-url/purl-spec)
- [purl-test.schema-0.1.json](https://packageurl.org/schemas/purl-test.schema-0.1.json)

PaawanBarach:
Closes #788

Adds missing test coverage for ECMA-427 specification edge cases:

- 3 tests for scheme with slashes (Section 5.6.1)
- 2 tests for plus sign rejection in type (Section 5.6.2)
- 4 tests for colon encoding (Section 5.4)
- 3 tests for empty qualifier values (Section 5.6.6)
- 3 tests for namespace segment encoding (Section 5.6.3)

All tests follow existing format and validate successfully against the JSON schema.


tests/spec/specification-test.json
```diff
@@ -215,6 +215,199 @@
...
+      "expected_failure": false,
+      "expected_failure_reason": null
+    },
+    {
+      "description": "colon in namespace should not be percent-encoded",
+      "test_group": "base",
+      "test_type": "build",
+      "input": {
+        "type": "generic",
+        "namespace": "org:example",
```
jkowalleck, line 307: is this actually allowed? 
this is still unclear, since the grammar was not yet published, and the spec is not explicit about it.

related discussion: https://github.com/package-url/purl-spec/pull/578/changes#r3102869652

jkowalleck: as far as i understand the spec, this is correct.
/closing

jkowalleck (approved):
to my understanding of the spec, all those `input` are valid purls, and all those `expected_output` are canonical purls

tests/spec/specification-test.json
johnmhoran: @jkowalleck I agree -- I think this bullet in 5.6.3 is broad enough to include a colon among the permitted namespace characters:  `may contain any Unicode character other than '/' unless the package's type definition provides otherwise`.

tests/spec/specification-test.json
```diff
@@ -215,6 +215,199 @@
...
+      "description": "colon in version should not be percent-encoded",
+      "test_group": "base",
+      "test_type": "roundtrip",
+      "input": "pkg:generic/foo@1.0:beta",
+      "expected_output": "pkg:generic/foo@1.0:beta",
+      "expected_failure": false,
+      "expected_failure_reason": null
+    },
+    {
+      "description": "encoded colon in name should decode and stay unencoded",
```
johnmhoran, line 327: @PaawanBarach I don't think the current language of ECMA-427 permits the change of an encoded colon to unencoded as a normalization step -- 427 is silent on that topic.  

But this test raises an interesting question -- does the mandate that colons not be percent-encoded -- `the colon ':', whether used as a Separator Character or otherwise` -- also mean that `%3A` is prohibited throughout the PURL string?  I don't know but would be interested in what you, @jkowalleck and other community members think.  This might be a good time to clarify if needed.

tests/spec/specification-test.json
```diff
@@ -215,6 +215,199 @@
...
+      "test_group": "base",
+      "test_type": "roundtrip",
+      "input": "pkg:generic/std%3Aio@1.0",
+      "expected_output": "pkg:generic/std:io@1.0",
+      "expected_failure": false,
+      "expected_failure_reason": null
+    },
+    {
+      "description": "qualifier with empty value should be discarded",
```
johnmhoran, line 336: Like the colon correction above, I don't see where ECMA-427 permits the fixing of a qualifier with an empty/missing value -- that is strictly prohibited by 5.6.6: `A value shall not be an empty string: a key=value pair with an empty value is the same as if no key=value pair exists for this key.`  This test should fail.

tests/spec/specification-test.json
```diff
@@ -215,6 +215,199 @@
...
+        "qualifiers": {
+          "valid": "yes"
+        },
+        "subpath": null
+      },
+      "expected_failure": false,
+      "expected_failure_reason": null
+    },
+    {
+      "description": "multiple empty qualifiers should all be discarded",
```
johnmhoran, line 354: Same as above -- these purported normalization steps are not mentioned in the spec and thus not permitted.  This test should also fail.

tests/spec/specification-test.json
```diff
@@ -215,6 +215,199 @@
...
+      "test_group": "base",
+      "test_type": "roundtrip",
+      "input": "pkg:npm/bar@2.0?a=&b=value&c=",
+      "expected_output": "pkg:npm/bar@2.0?b=value",
+      "expected_failure": false,
+      "expected_failure_reason": null
+    },
+    {
+      "description": "all-empty qualifiers should result in no query string",
```
johnmhoran, line 363: This test also fails for the reasons stated above.

tests/spec/specification-test.json
```diff
@@ -215,6 +215,199 @@
...
+        "version": null,
+        "qualifiers": null,
+        "subpath": null
+      },
+      "expected_output": "pkg:generic/org%20name/sub%20dir/lib",
+      "expected_failure": false,
+      "expected_failure_reason": null
+    },
+    {
+      "description": "namespace with encoded slash in segment should roundtrip correctly",
```
johnmhoran, line 404: This test should fail.  5.6.3 provides that a percent-decoded segment `shall not contain any slash '/' characters`. `a%2Fb/c` violates that rule.

johnmhoran (commented):
@PaawanBarach Thank you for this set of edge cases.  Although I have included a few comments, imho overall these are great additions to the suite, not only because they test quite a few edge cases but because your descriptions and expected_failure_reason(s) are helpfully detailed.

tests/spec/specification-test.json
jkowalleck: per purl spec, a colon `:` MUST NOT be percent-encoded at any times. 

so any string including `%3A` would be no valid Purl

tests/spec/specification-test.json
jkowalleck: i think this is a good test - one that canonicalizes the PURL on roundtrip

Maybe not a `test_group=base`, but a `test_group=advanced` test.

in the spirit of https://en.wikipedia.org/wiki/Robustness_principle
>„Be conservative in what you send, be liberal in what you accept.“  


I mean, we actually have a parsing description for this case: 
https://github.com/package-url/purl-spec/blob/37e04bb9b25b80b7cca27a2d4ad2175a3c1c8e56/docs/how-to-parse.md?plain=1#L34

tests/spec/specification-test.json
jkowalleck: Spec might not, but our own docs describe this behavior here: 
https://github.com/package-url/purl-spec/blob/37e04bb9b25b80b7cca27a2d4ad2175a3c1c8e56/docs/how-to-parse.md?plain=1#L34

johnmhoran: That's old, out-of-date, from the original .rst.

tests/spec/specification-test.json
johnmhoran: Again that is from the original .rst, not relevant to the current ECMA-427.

PR #806, opened 2026-02-10
drop swift testcases that require version

petergardfjall:
Version is an optional field for `pkg:swift` as [per the spec](https://github.com/package-url/purl-spec/blob/565d7f1d97b09d525b76d3cc5b62f521342284f1/types-doc/swift-definition.md?plain=1#L35).


mjherzog (approved):
Good correction

PR #813, opened 2026-02-16
Run purl-spec validation and regeneration steps

johnmhoran:
- 1 file modified: types-doc/bazel-definition.md


johnmhoran:
Updated the `bazel` type .md `Version deinition` to reflect the change of version from `**Requirement:** Required` to `**Requirement:** Optional`.

mjherzog (approved):
This is a good correction. I do not see any bazel test cases that require **version**.

PR #818, opened 2026-02-21
apk: add more qualifiers widely used by scanners and tools

xnox:
Many apk tools are using additional qualifiers for apk
purl. Contribute the status quo as a specification.

Similar to many other purls, distro= is in use by syft/grype and many
other tools to provide a more specific distribution version for a
given namespace. This is useful to tell apart apk from different
alpine release branches.

Syft uses upstream= to denote the package origin, which is loosely a
source package name or "origin" in apk-tools native speak. The name is
unfortunate, but it is widely used. It helps to identify and match
PURLs with the vulnerability feeds - as universally vulnerability
feeds are not on per binary apk name, but on the "origin" name.


pombredanne:
@xnox Thanks!... do you mind to add an issue to support the discussion on this PR? I'd like to make sure we get some input from alpine for instance.

PR #823, opened 2026-03-04
Add new PURL type: 'git' (for generic git repositories)

### LINKED ISSUE #780: Add "generic" git PURL Type
This issue is to aggregate open issues related to:
- Adding a "generic" git PURL type
- Adding new PURL types for repositories/forges based on Git
- Definition of the syntax for the **vcs_url** qualifier

The key open question is whether there should be:

1. One git PURL type with the **vcs_url** qualifier used to specify the repository details
2. One PURL type for each Git-based repository (as requested)
3. Some combination of approaches - i.e individual PURL types for major repositories like GitHub, GitLab and others with a more generic git PURL type for other minor repositories with the possibility of promotion to an individual PURL type

darakian:
This PR adds a purl type which is intended to be used for project source code stored in a git repository. resolves #780

In its current form it mimics the swift type defintion for namespace and name definition. Version is defined as a git reference based on the conversation in #780 Ref: https://git-scm.com/book/en/v2/Git-Internals-Git-References

Happy to iterate, but wanted to get this up 👍


darakian:
@pombredanne I was going to add a human doc file as well in the `types-doc/` folder, but the headers on those all mention that they're auto generated, so.... I skipped that. Happy to write up some human prose elsewhere.

mjherzog:
@darakian The `types-doc` folder is currently reserved for the auto-generated type documentation, but we do need a place to put additional human friendly documentation at the PURL **type** level.
See also:
- #824 
- #601

It would be helpful if you draft and share a markdown file with the additional documentation that you have in mind.

types/git-definition.json
```diff
@@ -0,0 +1,33 @@
+{
+  "$schema": "https://packageurl.org/schemas/purl-type-definition.schema-1.0.json",
+  "$id": "https://packageurl.org/types/git-definition.json",
+  "type": "git",
+  "type_name": "Git",
+  "description": "Git-based source packages",
```
alilleybrinker, line 6: Nit: would "source repositories" be clearer / more accurate than "source packages"?

types/git-definition.json
```diff
@@ -0,0 +1,33 @@
...
+  "name_definition": {
+    "requirement": "required",
+    "case_sensitive": true,
+    "native_name": "repository name with owner",
+    "note": "The repository name."
+  },
+  "version_definition": {
+    "requirement": "optional",
+    "native_name": "A git reference",
+    "note": "The version is a git reference. Ideally a commit or tag."
```
alilleybrinker: Is there a mechanism to explicitly link to the documentation for git references? Just to be maximally clear what "git reference" means in this context.

alilleybrinker (commented):
A couple of minor thoughts.

types/git-definition.json
darakian: It could be. I was trying to stay with the package verbiage to align with the rest of the package url types, but I'm not opposed to swapping. @pombredanne do you have a preference one way or another?

types/git-definition.json
darakian: That's a good call out. It looks like a few of the types have url references in their type definitions, so I added https://github.com/package-url/purl-spec/pull/823/commits/8c9402c1265b2e376f6cd1b572b6ef47318aefa2

darakian:
@mjherzog I take it that there's no specific place to add those docs yet? Can do on drafting them, but where should I create the file?

types/git-definition.json
```diff
@@ -0,0 +1,33 @@
...
+    "native_name": "repository name with owner",
+    "note": "The repository name."
+  },
+  "version_definition": {
+    "requirement": "optional",
+    "native_name": "A git reference",
+    "note": "The version is a git reference (https://git-scm.com/book/en/v2/Git-Internals-Git-References). Ideally a commit or tag."
+  },
+  "examples": [
+    "pkg:git/codeberg.org/forgejo/forgejo/@a72d2c07cfca03b55371089de6aa230d8c951fa0#options/locale_readme.md",
```
sschuberth: Isn't this supposed to be _exactly_ the (canonical) Git clone URL? If so, why is the `.git` suffix being omitted in this example?

Similarly, shouldn't a trailing `/` be omitted?

darakian: Interestingly enough I can't find a primary source reference talking about the `.git` suffix. The best I was able to come across in searching is this stack overflow question where there is an assertion (without reference) that a trailing `.git` is a naming convention
https://stackoverflow.com/questions/8686691/what-does-the-git-mean-in-a-git-url
the primary git reference shows them off in examples but does not seem to explain them directly
https://git-scm.com/docs/git-clone.html#_git_urls

Testing locally (and also with the trailing `/`) I see some differences but they all come down to which url is populated in the local `.git` dir.
ex. when cloning this repo the three different ways I see
```
jon~/g/randos❯❯❯ diff test-1/.git/config test-2/.git/config
9c9
< 	url = https://github.com/package-url/purl-spec
---
> 	url = https://github.com/package-url/purl-spec/
...
jon~/g/randos❯❯❯ diff test-1/.git/config test-3/.git/config
9c9
< 	url = https://github.com/package-url/purl-spec
---
> 	url = https://github.com/package-url/purl-spec.git
```

For the purposes of identifying the actual code they all seem equivalent. I suppose there's a choice for use to deviate from what is allowed in the world of git and git tools and force a normalization or to align with the git world for ease of interoperability. Based on the conversation in #780 I believe the preference is to align with the norms of the git world.

That said, looking into the git url doc did make me think that the `namespace` and `name` definition should be altered slightly to align with the how git urls are defined upstream.
See: https://github.com/package-url/purl-spec/pull/823/changes/88e820a53b1e0b1b2bdbb58977e2ec83d752c3c8

@andrew you don't have any insight into `.git` and trailing `/`s do you?

sschuberth: > Interestingly enough I can't find a primary source reference talking about the `.git` suffix.

I believe its (conventional) use is simply hosting-platform-specific. I'm simply looking at the platform's "copy URL to clipboard" UI to see if that platform prefers to have a ".git" suffix in the URL or not.

So I agree that we can't have a general rule here, but for the scope of the examples, which name concrete platforms, IMO we should adhere to the preference of that platform.

andrew: I believe the `.git` suffix comes from the original git-http-backend CGI script served repositories over HTTP by mapping the URL path directly to the filesystem. 

As far as I know all the major forges support both with and without `.git` when cloning

matt-phylum: It is commonly removed by forges, not Git itself, meaning it's not possible to tell for any given Git URL whether the .git is optional or not.

Forgejo: https://codeberg.org/forgejo/forgejo/src/commit/df79ccf7d8b69f63b7cb66d340e26ce1e3e79c89/routers/web/repo/githttp.go#L62
GitLab: https://gitlab.com/gitlab-org/gitlab/-/blob/358f46317b3613756ee6c471379d4cbdc33d9197/config/routes/git_http.rb#L58-68

darakian: @sschuberth would you like to propose an example to add/edit? It seems to me that we essentially have an equivalence class that the consumer of a git purl will need to be aware of on consumption and having an example with a trailing `.git` probably is a good idea.

Many thanks for the context @andrew and @matt-phylum 👍

sschuberth: > [@sschuberth](https://github.com/sschuberth) would you like to propose an example to add/edit?

Well, any Git clone URL of a GitHub project would do, I guess, so for example:

    pkg://git/github.com/oss-review-toolkit/ort.git@86c6b09b93db996689735d1eaabaa86a1051a319#cli/build.gradle.kts

sschuberth: However, if we give such an example, we should probably make clear that this is not the "canonical form" of a GitHub PURL, and people should prefer `pkg://github/...` instead.

matt-phylum: `pkg:git/projects.blender.org/blender/blender.git` and `pkg:git/projects.blender.org/blender/blender` are equivalent, but PURL implementations cannot be expected to know.

matt-phylum:
Git supports multiple different protocols, which HTTP over TLS now being the most common. Can a Git PURL be constructed for:
- A filesystem path, which may point to a network resource on a company LAN?
- HTTP without TLS?
- HTTPS on a different port?
- SSH?
- SSH on a different port?
- The Git protocol (eg git://git.kernel.org/pub/scm/bluetooth/bluez.git)?

types/git-definition.json
darakian: @sschuberth how do you feel about the example (stolen from @matt-phylum) added in https://github.com/package-url/purl-spec/pull/823/commits/8b1d9f0cd4392096017fec0479ddd44c8f759b2f ?

As for the advice about preferring the github type over the more primitive git type; totally agree. That's probably something to add to the human readable doc that @mjherzog mentioned. Maybe that's a follow up PR to add that though? I'm not sure there's a paved path for that doc yet (please correct me if I'm wrong).

darakian:
@matt-phylum transports were mostly discussed synchronously in the purl community meetings (though I did mention it https://github.com/package-url/purl-spec/issues/780#issuecomment-3922834423 ), but transports are out of scope. The goal with this type is to describe content rather than access method.

types/git-definition.json
sschuberth: > [@sschuberth](https://github.com/sschuberth) how do you feel about the example (stolen from [@matt-phylum](https://github.com/matt-phylum)) added in [8b1d9f0](https://github.com/package-url/purl-spec/commit/8b1d9f0cd4392096017fec0479ddd44c8f759b2f) ?

Fine with me.

> As for the advice about preferring the github type over the more primitive git type; totally agree. [...] Maybe that's a follow up PR to add that though?

Fine with me as well to do that as a follow-up; we just should not forget about giving that advice.

types/git-definition.json
```diff
@@ -0,0 +1,34 @@
...
+    "native_name": "repository name with owner",
+    "note": "The path on the host to the git repository. See: https://git-scm.com/docs/git-clone.html#_git_urls"
+  },
+  "version_definition": {
+    "requirement": "optional",
+    "native_name": "A git reference",
+    "note": "The version is a git reference (https://git-scm.com/book/en/v2/Git-Internals-Git-References). Ideally a commit or tag."
+  },
+  "examples": [
+    "pkg:git/codeberg.org/forgejo/forgejo/@a72d2c07cfca03b55371089de6aa230d8c951fa0#options/locale_readme.md",
```
alilleybrinker: This example uses the `#`, followed by a path to a file in the repository, but it doesn't look like this behavior is specified in this document.

Also, we likely ought to specify / explain in whatever documentation is made for this type that the path part must be stripped if you're going to `git clone` a repository.

darakian: I'm not opposed to expanding on the the `#` path behavior, but it is also shared with the `github` and `bitbucket` types and I think comes from the `file_name` qualifier here
https://github.com/package-url/purl-spec/blob/main/docs/common-qualifiers.md
I'll be honest, I built from the example of the github type which has that same behavior here
https://github.com/package-url/purl-spec/blob/main/types/github-definition.json#L30

The `#` character is not explicitly mentioned in the common qualifiers doc, so perhaps this is a spec level clarification rather than just for this type?

alilleybrinker: Hm, good references. I suppose this goes to the purl folks for where information should be communicated. I'm in favor of being as explicit as possible (even if that means that the specs for other types need to change to be more detailed).

mjherzog: @alilleybrinker In any case we need to start with documenting details outside of the `<purl-type>.definition.json` files that are based on the Schema included in ECMA-427 (Clause 6). Making schema level changes based on the most complex cases is problematic and we need many good examples before proposing schema changes.

There are 3 other places to put this information:
- Test cases for a PURL **type**
- PURL **type** background documentation
- Specific PURL **type** information in the How to documentation

We have flexibility for adding new documentation and we now have an efficient process for publishing documentation at www.packageurl.org.

darakian: Alright, with a little help I found the reference @alilleybrinker. I was looking for a file specifier and the term I should have been looking for was `subpath`. So, I believe we're good on this being documented as normal construction
https://packageurl.org/docs/purl/specification#subpath

darakian:
I let this get away from me, but added some tests. Was there anything else we wanted prior to merging? Was the human doc something to do async?

mjherzog:
@darakian We need the PR to include:
- Update to `purl-types-index.json`
- Generation of the type documentation

We are still working on better documentation for this but the basic steps are in `CONTRIBUTING.md` (formerly in README-dev.md)

For the bonus PURL **type** documentation, please add your file to the new `docs/drafts` folder. I suggest the naming:
`docs/drafts/types/git-documentation.md` for now. 

I will ask @johnmhoran to review the test cases.

darakian:
@mjherzog Ok. I've updated the PR with both the rendered `types-doc` and a human doc following your naming suggestion. The human doc is a little bare bones, but I tried to give a basic intro to the uninitiated as well as some human text for the different parameters.

docs/drafts/types/git-documentation.md
```diff
@@ -0,0 +1,15 @@
+# Git purl type
+The git purl type is a primitive type intended to be used as a base case for encoding information about a code base which exists under git version control.
+
+## Definitions
+namespace: The namespace for this purl type is defined as the url path to the git host. 
+eg. The `host` in git terminology.
+https://git-scm.com/docs/git-clone.html#_git_urls
```
booniepepper: It's a little different from repository_url in other PURLs, but I really do think it's justified for this protocol

types/git-definition.json
booniepepper: My two cents would be to use the word "repository" here in the description, but for the purposes of PURL, a git repository can be considered a "package" in other contexts.

booniepepper (approved):

docs/drafts/types/git-documentation.md
darakian: `repository_url` is probably redundant for this type. Perhaps its worth explicitly forbidding it though

types/git-definition.json
darakian: That's two votes for the term `repository` I guess. I'll again defer to @pombredanne here 😄

darakian:
@mjherzog and @pombredanne. I think this is good to go aside from the two questioned raised today. 

I'm neutral on the question about terminology choice  (`repository` vs `package`) in the type description and could be changed later too if desired. As for the `repository_url` question; how do you two feel about forbidding its use for this type given that the location is encoded into the name and namespace? This could also be done later if you want to punt on that for the moment 😄

booniepepper:
Hope I didn't delay this with my feedback. I think both could be added later and neither should be blocking. I think I can sort of see a (weak) argument for supporting repository_url to indicate a mirror (opposed to a fork)

Anyway, @mjherzog and @pombredanne I think this is good to go too

darakian:
Following up on the topic of a logo. I suggest we use the `Orange logomark for light backgrounds`
Orange git logo
from
https://git-scm.com/community/logos
the usage policy for this logo is available here
https://git-scm.com/about/trademark
and I believe this use case is covered by section 2.2 `Use of the Marks without written permission`

docs/drafts/types/git-documentation.md
mjherzog: I don't think that we have a way to prohibit a mechanism to prohibit use of a **qualifiers key** like `repository_url`  and I doubt that we want one. The best approach to discourage that usage would be an 'advanced' test with an "expected_message" about why it is redundant. That approach should be available soon with the v0.02 PURL test schema - see proposal at: https://github.com/package-url/purl-spec/blob/main/docs/drafts/test-overview-schema-v0.01.md

darakian: That's fair and I think the idea of using `repository_url` as an indicator of a mirror location does imply a valid use case (niche as it may be).

types/git-definition.json
mjherzog: I recommend that we stay with "source packages" for now because the PURL will usually refer to a "package" within a repository.
This description is editorial only so changing it in the future would be a minor change.

darakian: Works for me 👍

types/git-definition.json
mjherzog: Just confirming that '#' is the separator character for the **subpath** PURL component - it has nothing to do with **qualifiers**

types/git-definition.json
mjherzog: @darakian Do you have time to update the new `types/git-documentation.md` file with this commentary and advice?
This seems like the kind of background information that would be good to provide in the supplemental **type** documentation.

darakian: @mjherzog ya, I can do that. Are there any other markdown files I can follow? Also, is `https://github.com/package-url/purl-spec/tree/main/docs/types/definitions` the correct full path? I think everything in there right now has been auto generated.

mjherzog:
@darakian I have pushed some changes to the  `docs/' folder structure which includes the new `docs/types/notes/` folder and with a draft template for the notes content. Please update your PR to move/rename:
- `docs/drafts/types/git-documentation.md`
to
- `docs//types/notes/git-notes.md`
The template is a first draft for suggested content. My thinking is:
- The notes need to supplement, not repeat, notes that are already in the notes in the JSON file.
- A big point is to summarize the discussion and resolution of important questions/issues that surfaced during discussion of the new PURL **type** and spare future generations the slog of going through the comment threads. This will be an important reference for people who come along later and raise questions about the rational for the definition.
- We can reasonably standardize basic updates for the history of when a PURL **type** was registered or changed in a normative way. This is also for posterity to spare people from needing to research this individually.

I do not have any examples atm so you are the pioneer here....thank you for your service

darakian:
Merged in your changes and added a new doc in https://github.com/package-url/purl-spec/pull/823/changes/880d837757fb21c72592417c32d3096a80499c01 which should capture the main points.

mjherzog:
@darakian I made some editorial changes to the `git-notes.md` file

mjherzog:
@darakian
I made two sets of changes:
- updated `git-notes.md` for consistent style with other docs
- corrected all instances of the "pkg:git/codeberg.org/forgejo/forgejo..." PURL string in test cases to remove the trailing slash before the version. For the 'base' test cases the trailing slash would mean a failing test case input.

I have one last item for you. There seems to be a duplicate set of test cases:
- lines 13-53
- lines 54-94

The only differences I can see here are the descriptions. Am I missing something?

darakian:
Many thanks for the updates and indeed those are the same tests. When I was making this set of tests I was mocking the github type and it looks like the same duplication is happening there too in the same lines (albeit with shorter hashes in lines 54-94).
https://github.com/package-url/purl-spec/blob/main/tests/types/github-test.json

I suppose we just drop the duplicates?

mjherzog:
@darakian Thank you for confirming. I will delete the extra set at lines 13-53 and merge the PR.

darakian:
sounds good 👍

alilleybrinker:
Well done @darakian! Great to see this merged!

PR #827, opened 2026-03-05
Update CPAN Spec

giterlizzi:
This PR proposes updates to the CPAN PURL type definition to improve adoption in vulnerability and packaging workflows.

Requiring a `namespace` is too restrictive when only the CPAN distribution name/version is available (e.g., CVE attribution, CSAF, vulnerability management, downstream packaging such as NixOS, etc.).

**Changes**
- `namespace` is now optional (not deprecated). It may be reserved for future semantics.
- Added `author` qualifier for the CPAN author/publisher ID (CPANID).
- Added `distpath` qualifier for a repository-relative distribution path (e.g., `authors/id/...`).
- Clarified name is the distribution name; parsers MUST reject name containing `::` (module naming).

**Examples**
- pkg:cpan/perl@5.42
- pkg:cpan/DBI@1.646
- pkg:cpan/SBOM-CycloneDX
- pkg:cpan/URI-PackageURL?author=GDT
- pkg:cpan/libwww-perl@6.76?author=OALDERS
- pkg:cpan/DateTime@1.55?author=DROLSKY&repository_url=backpan.perl.org
- pkg:cpan/Term-Gnuplot@0.90380906?distpath=authors/id/I/IL/ILYAZ/modules/Term-Gnuplot-0.90380906.zip


tests/types/cpan-test.json
```diff
@@ -5,22 +5,24 @@
       "description": "CPAN distribution name are case sensitive. Roundtrip an input purl to canonical.",
       "test_group": "advanced",
       "test_type": "roundtrip",
-      "input": "pkg:cpan/DROLSKY/DateTime@1.55",
-      "expected_output": "pkg:cpan/DROLSKY/DateTime@1.55",
+      "input": "pkg:cpan/DateTime@1.55?author=DROLSKY",
+      "expected_output": "pkg:cpan/DateTime@1.55?author=DROLSKY",
```
sjn: I'm wondering if it would make sense to keep the original tests unchanged (since they still test for a valid use), and instead _add_ separate tests for the new syntax?

That way we have tests to ensure that backward compatibility to previous versions of this spec still works.

types/cpan-definition.json
```diff
@@ -9,9 +9,8 @@
     "default_repository_url": "https://www.cpan.org/"
   },
   "namespace_definition": {
-    "requirement": "required",
-    "native_name": "CPAN ID of the author/publisher",
-    "note": "It MUST be written uppercase and is required"
+    "requirement": "optional",
+    "note": "There is no author namespace. Use 'author' qualifier instead"
```
sjn: If it's optional, then the note should state this fact, and under what circumstances it is ok to use it. We can still discourage it (i think that's not necessary), or describe the situations where it makes sense to use it.

Also, the "native_name" is still correct, even if it is optional, right?

types/cpan-definition.json
```diff
@@ -43,14 +52,18 @@
     {
       "key": "ext",
       "requirement": "optional",
-      "description": "file extension",
+      "description": "distribution file extension",
       "default_value": "tar.gz"
     }
   ],
-  "note":"The previous CPAN PURL type specification allowed module names (e.g. URI::PackageURL) to be used as PURL 'name' while also omitting the PURL 'namespace'. The parser MUST emit an error when a module is specified as a PURL 'name' or detect '::' characters.",
+  "note":"The previous CPAN PURL type specification allowed the CPAN ID of the author/publisher to be used as PURL 'namespace'. To specify the author/publisher CPAN ID, use the 'author' qualifier instead. A parser MUST reject PURL name values that contain the '::' separator (module name).",
```
sjn: I'm thinking that we can remove the text that refers to "previous CPAN PURL type", and instead focus on just describe how *this* type specification is used.

Since the CPAN ID author is optional, then then let's describe the different situations where it makes sense to use them.

types/cpan-definition.json
```diff
@@ -43,14 +52,18 @@
...
-    "pkg:cpan/GDT/URI-PackageURL",
-    "pkg:cpan/OALDERS/libwww-perl@6.76",
-    "pkg:cpan/DROLSKY/DateTime@1.55?repository_url=backpan.perl.org"
+    "pkg:cpan/perl@5.42",
+    "pkg:cpan/DBI@1.646",
+    "pkg:cpan/SBOM-CycloneDX",
+    "pkg:cpan/URI-PackageURL?author=GDT",
+    "pkg:cpan/libwww-perl@6.76?author=OALDERS",
+    "pkg:cpan/DateTime@1.55?author=DROLSKY&repository_url=backpan.perl.org",
+    "pkg:cpan/Term-Gnuplot@0.90380906?distpath=authors/id/I/IL/ILYAZ/modules/Term-Gnuplot-0.90380906.zip"
```
sjn: The `pkg:cpan/DROLSKY/DateTime@1.55?repository_url=backpan.perl.org` example is still a valid one - even if we discourage it, there may be existing PURls references out there in existing documents  & metadata. Do we wish to allow these to continue to work? I think we should.

sjn (commented):
Some comments about the optionality of the CPAN ID (author) namespace; We may want to consider keeping around tests and comments describing it's use (and recommended alternative when this is the wrong option).

stigtsp (approved):
Looks good to me. Thanks for fixing this

stigtsp:
Anything left to address on this PR? With this change, we will be able to add package URLs to new and historical CVE records for example.

types/cpan-definition.json
```diff
@@ -25,6 +24,16 @@
...
   "qualifiers_definition": [
+    {
+      "key": "author",
+      "requirement": "optional",
+      "description": "CPAN ID of the author/publisher"
+    },
+    {
+      "key": "distpath",
+      "requirement": "optional",
+      "description": "Repository-relative path (not a URL) to the distribution archive or directory, typically under 'authors/id/...'"
```
mjherzog, line 36: Could **subpath** be a better solution for this?

mjherzog:
@stigtsp Three things seem open:
- Resolution of the comments from @sjn 
- The PR is missing the updated version of types-doc/cpan-definition.md
- The PR is still in WIP status

types/cpan-definition.json
giterlizzi: Subpath are usually used when you need to refer to a specific file contained within the package.

tests/types/cpan-test.json
giterlizzi: Re-added most of the previous tests.

types/cpan-definition.json
giterlizzi: I've updated the CPAN type and "namespace" notes with a brief explanation of how to use namespaces and the "author" qualifier.

If it's too complicated, we can simplify it.

giterlizzi:
> The PR is missing the updated version of types-doc/cpan-definition.md

I remember that GH Action converts to MD format, which is why I didn't include it in the PR

mjherzog:
> > The PR is missing the updated version of types-doc/cpan-definition.md
> 
> I remember that GH Action converts to MD format, which is why I didn't include it in the PR

The GH Action is not automatically running atm - we need to make adjustments for security such as pinning the GH Actions

mjherzog (approved):
The changes look good. Since the proposal is to make namespace optional instead of required there does not seem to be a backward-compatibility issue.

stigtsp:
Just checking in on this: Any blockers? We're waiting for this change to support package URLs in our CVEs.

tests/types/cpan-test.json
```diff
@@ -1,5 +1,5 @@
 {
-  "$schema": "https://packageurl.org/schemas/purl-test.schema-0.1.json",
+  "$schema": "https://packageurl.org/schemas/purl-test.schema-1.0.json",
```
mjherzog: Please remove this change.  The current test schema is purl-test.schema-0.1.json.

tests/types/cpan-test.json
```diff
@@ -1,5 +1,5 @@
...
   "tests": [
     {
       "description": "CPAN distribution name are case sensitive. Roundtrip an input purl to canonical.",
```
mjherzog, line 5: This test case seems to be a duplicate of the following test case:
"CPAN distribution name are case sensitive. Roundtrip a canonical input to canonical output.",
except that the test_group is 'advanced' in this case. Is this supposed to be an example with a non-canonical input PURL?

mjherzog (commented):
@stigtsp We are pretty close for these changes.
See my comments:
- The schema version change must be reverted
- Removing the first test case is optional
And we have a conflict with cpan-definition.md because the markdown files were moved to `docs/types/definitions` . We have recently reinstated the workflow to generate the markdown doc files so you can remove that file from the PR to resolve the conflict (sorry for the "churn" on what we requested for the docs workflow)

tests/types/cpan-test.json
johnmhoran: Hmm.  How is the input on line 8 non-canonical?

tests/types/cpan-test.json
giterlizzi: Reverted

types/cpan-definition.json
```diff
@@ -43,14 +53,18 @@
...
-    "pkg:cpan/GDT/URI-PackageURL",
-    "pkg:cpan/OALDERS/libwww-perl@6.76",
-    "pkg:cpan/DROLSKY/DateTime@1.55?repository_url=backpan.perl.org"
+    "pkg:cpan/perl@5.42",
+    "pkg:cpan/DBI@1.646",
+    "pkg:cpan/SBOM-CycloneDX",
+    "pkg:cpan/URI-PackageURL?author=GDT",
+    "pkg:cpan/libwww-perl@6.76?author=OALDERS",
+    "pkg:cpan/DateTime@1.55?author=DROLSKY&repository_url=backpan.perl.org",
+    "pkg:cpan/Term-Gnuplot@0.90380906?distpath=authors/id/I/IL/ILYAZ/modules/Term-Gnuplot-0.90380906.zip"
```
johnmhoran: A slash '/' in a qualifiers value must be percent-encoded.

johnmhoran (changes requested):
One example has a few slashes that must be percent-encoded.

types/cpan-definition.json
giterlizzi: Fixed

johnmhoran (approved):
Thanks @giterlizzi -- LGTM.  :+1:

giterlizzi:
Thanks !

PR #855, opened 2026-04-08
Add new doc/drafts folder

mjherzog:
Adding new folder for drafts of new documentation files and major changes (rewrite) of existing documentation. 
The goal is to make document review easier and more efficient than it would be using a PR.


johnmhoran (approved):
Looks good @mjherzog and will be very handy.  :+1:

Issue #871, opened 2026-05-21
PURL type documentation utility does not document the case_sensitive property


mjherzog:
The etc/scripts/generate_index_and_docs.py utility does not publish the `case_sensitive` property information
Examples are:
- 'golang' - Notes for **namespace** and **name** say: "The namespace must be lowercased." but the `case_sensitive` property is missing for both in the generated documentation.
- 'npm' - Note for **name** says: ""Per the package.json spec, new package 'must not have uppercase letters in the name', therefore the name must be lowercased. The npm name used to be case sensitive in the early days for some old packages." but the `case_sensitive` property is missing for both **namespace** and **name** in the generated documentation.

Lines 92-93 of `generate_index_and_docs.py` say:
```
        if case_sensitive := component.get("case_sensitive"):
            lines.append(f"- **Case Sensitive:** {get_yes_no(case_sensitive)}")
```
but this does not seem to be working.

After fixing this but we need to rerun the documentation generation for all PURL **types**.

mjherzog:
Explanation from @johnmhoran 
This checks for true/false values in the *-definition.json 'case_sensitive' values using Python's so-called walrus operator, which checks for truthiness.  Basically, true ==> display Case Sensitive: Yes, otherwise display is not triggered.

```
def get_yes_no(value): """Return a human-readable yes/no from a boolean value""" return "Yes" if value else "No"

. . .

        if case_sensitive := component.get("case_sensitive"):
            lines.append(f"- **Case Sensitive:** {get_yes_no(case_sensitive)}")
```

PR #872, opened 2026-05-22
Update case_sensitive properties for golang, npm

johnmhoran:
Changed the respective `case_sensitive` values to `true` in `*-definition.json` files and reran the documentation commands.

Reference: https://github.com/package-url/purl-spec/issues/871

### DIFF
--- a/types-doc/golang-definition.md
+++ b/types-doc/golang-definition.md
@@ -21,11 +21,13 @@ The structure of a PURL for this package type is:
 ## Namespace definition
 
 - **Requirement:** Required
+- **Case Sensitive:** Yes
 - **Note:** `The namespace must be lowercased.`
 
 ## Name definition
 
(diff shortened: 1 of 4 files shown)


mjherzog (commented):
This looks good except that we need to update the **name**  note for npm because it describes the old definition of this property?
Current: "Per the package.json spec, new package 'must not have uppercase letters in the name', therefore the name must be lowercased. The npm name used to be case sensitive in the early days for some old packages.`
Possible replacement: "The package.json spec changed in 2015 to require that a new package 'must not have uppercase letters in the name', but old packages with mixed case names were "grandfathered in".

johnmhoran:
Updated as requested @mjherzog , ready for your eyes.

mjherzog (approved):
This is ready to merge.

PR #875, opened 2026-06-04
Update pkg-iana-registration.txt

mjherzog:
Updated to prepare for registration following the template from https://www.rfc-editor.org/info/rfc7595/#section-7.4
- changed Contact and Controller to aboutcode.org
- Removed the scheme syntax and scheme semantics according to latest guidance - these are covered by ECMA-427
- Added URL for more applications/protocols listed on the packageurl.org website
- Updated References / Scheme specification to ECMA-427

### DIFF
--- a/pkg-iana-registration.txt
+++ b/pkg-iana-registration.txt
@@ -2,18 +2,12 @@ Scheme name: pkg
 
 Status: Provisional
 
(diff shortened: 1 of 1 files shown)


mjherzog:
I made the changes directly to:
-  #141 
to avoid confusion

PR #877, opened 2026-06-09
Add PURL type: 'nix'

### LINKED ISSUE #876: Add PURL type: nix

chinyeungli:
This PR replaces https://github.com/package-url/purl-spec/pull/314 which was based on the old rST definition file.

There has been discussion about the use of commit_hash, since commit hashes are not comparable and may introduce additional complexity when performing vulnerability matching, even though a commit hash is specific enough to uniquely identify the nixpkgs revision of a package.

As the purpose of purl is to provide
> a simple, consistent, and flexible approach to identifying software packages with precision and clarity.

I believe this approach fits that goal, and that vulnerability identifiers, perhaps, should be handled by the security scanner rather than the PURL itself.

Feedback and discussion are welcome.


chinyeungli:
Discussion can take place at https://github.com/package-url/purl-spec/issues/876

docs/candidate-purl-types.md
```diff
@@ -28,7 +28,7 @@
 - ``melpa`` for Emacs packages
 - ``meteor`` for Meteor JavaScript packages
 - ``nim`` for Nim packages
-- ``nix`` for Nixos packages - See https://github.com/package-url/purl-spec/pull/314 
+- ``nix`` for Nixos packages - See https://github.com/package-url/purl-spec/pull/877
```
lucperkins:
```suggestion
- ``nix`` for Nix packages - See https://github.com/package-url/purl-spec/pull/877
```

docs/candidate-purl-types.md
```diff
@@ -28,7 +28,7 @@
 - ``melpa`` for Emacs packages
 - ``meteor`` for Meteor JavaScript packages
 - ``nim`` for Nim packages
-- ``nix`` for Nixos packages - See https://github.com/package-url/purl-spec/pull/314
```
navayla777-dotcom, line 31: chinyeungli:876-adding_nix_type

docs/candidate-purl-types.md
```diff
@@ -42,5 +42,5 @@
 - ``vagrant`` for Vagrant boxes
 - ``vcpkg`` for Microsoft vcpkg - See https://github.com/package-url/purl-spec/issues/217
 - ``vim`` for Vim scripts packages
-- ``winget`` for Windows Package Manager - See https://github.com/package-url/purl-spec/issues/255
```
navayla777-dotcom, line 45: chinyeungli:876-adding_nix_type

navayla777-dotcom (approved):
#880 Dock type
#876 type: nix

navayla777-dotcom (commented):
#880 dock

mjherzog:
@chinyeungli I have reviewed the PR in some detail and it looks good to me, but I have to admit that I am not completely clear about how the nixpkgs ecosystem works.  We seem to have reasonable input from community members.

One request is that you add failing test cases to cover **type** specific conditions. In this case **namespace** seems to be the component with test-relevant **type** specific conditions.
- missing namespace [required]
- mixed-case namespace [expected lowercase when case_sensitive=false

The other PURL components (of course) have specific definitions for the component content but not in a way that lends itself to **type**-specific test cases - e.g. validating values for the 'system' or 'output' **qualifiers keys**.

RossComputerGuy:
> One request is that you add failing test cases to cover **type** specific conditions. In this case **namespace** seems to be the component with test-relevant **type** specific conditions.
> 
>     * missing namespace [required]

I'm a heavy user of Nix and I don't think a missing namespace is even possible. If it is missing, then something is malformed and bad. This is because you need Nix expressions to come from somewhere. The central place to get the Nix expressions is nixpkgs, the main repository for Nix packages (hence the name). There's also other repos people can get expressions from like dotfile repos.

In terms of the PURL spec, I am not sure how it could be handled where Nix packages have come from since there's many ways to get them like Git (GitHub, GitLab, local Git path), local file path (without Git). `nixpkgs` is an easy one since that is pretty much https://github.com/NixOS/nixpkgs. But with as many possible locations there are, maybe the spec needs a [Flake reference](https://zero-to-nix.com/concepts/flakes/#references).

mjherzog:
@RossComputerGuy The point for missing namespace test is not that whether a missing namespace is likely but rather to check for that condition because an PURL for the 'nix' **type** could come from anywhere.

chinyeungli:
@RossComputerGuy Thanks for the feedback!

What do you think about introducing an optional `flake_ref` qualifier?
For instance:
```
pkg:nix/dotfiles/my-project?flake_ref=github:myuser/dotfiles
```

This keeps the namespace and name as clean logical identifiers, while the qualifier handles the precise location tracking as it captures exactly where the remote sources are located. 
(It also maps to a native Nix installation command: `nix build github:myuser/dotfiles#my-project`.)

What do you think of this approach? Suggestions are very welcome along with some concrete samples.

----

>One request is that you add failing test cases to cover type specific conditions

@mjherzog will do

RossComputerGuy:
> @RossComputerGuy The point for missing namespace test is not that whether a missing namespace is likely but rather to check for that condition because an PURL for the 'nix' **type** could come from anywhere.

Gotcha

> What do you think about introducing an optional `flake_ref` qualifier?

Maybe call it `flakeref`, `ref`, or `flake`? Underscores seem funky in URI's.

chinyeungli:
> Maybe call it flakeref, ref, or flake? Underscores seem funky in URI's.

"flakeref" sounds more specific
```
pkg:nix/dotfiles/my-project?flakeref=github:myuser/dotfiles
```

PR #885, opened 2026-06-12
Add new PURL type: vcpkg

### LINKED ISSUE #217: Add support for Microsoft's cross-platform C and C++ package manager, `vcpkg`

JamieMagee:
Registers a `vcpkg` PURL type for Microsoft's C/C++ package manager. Continues #562 (and #245); closes #217.

## Adds
- `types/vcpkg-definition.json`: type definition
- `tests/types/vcpkg-test.json`: parse tests
- `types-doc/vcpkg-definition.md`: generated docs
- `vcpkg` in `purl-types-index.json`

## Shape
`pkg:vcpkg/<name>@<version>?<qualifiers>`

- namespace: prohibited (vcpkg has no namespace; prefixes like `boost-` are part of the name)
- name: required, the port name
- version: optional, the upstream version
- qualifiers: `port_version` (port-version, default 0), `repository_revision` (registry baseline SHA), `triplet` (target config), `repository_url` (non-default registry)

## Design notes
- `port_version` is a qualifier, not part of the version: `#` is the PURL subpath delimiter and vcpkg's `version-string` scheme disallows it. Matches what `microsoft/component-detection` emits, e.g. `pkg:vcpkg/nlohmann-json@3.10.4?port_version=5`. Resolves @pombredanne's question on #562.
- `triplet` is one optional qualifier (e.g. `x64-windows-static`), not separate os/arch/linkage keys. It's build/target context and doesn't affect package identity.
- `repository_revision` is the 40-char registry commit SHA (the `baseline` in `vcpkg-configuration.json`).

`make check` passes; docs regenerated with `make gendocs`. Related: https://github.com/microsoft/vcpkg/issues/44201.

### DIFF
--- a/tests/types/vcpkg-test.json
+++ b/tests/types/vcpkg-test.json
@@ -1,0 +1,251 @@
+{
+  "$schema": "https://packageurl.org/schemas/purl-test.schema-0.1.json",
+  "tests": [
+    {
+      "description": "simple vcpkg purl",
+      "test_group": "base",
+      "test_type": "parse",
+      "input": "pkg:vcpkg/bzip2@1.0.8",
+      "expected_output": {
+        "type": "vcpkg",
+        "namespace": null,
+        "name": "bzip2",
+        "version": "1.0.8",
+        "qualifiers": null,
+        "subpath": null
+      },
+      "expected_failure": false,
+      "expected_failure_reason": null
+    },
+    {
+      "description": "vcpkg with port_version",
+      "test_group": "base",
+      "test_type": "parse",
+      "input": "pkg:vcpkg/bzip2@1.0.8?port_version=6",
+      "expected_output": {
+        "type": "vcpkg",
+        "namespace": null,
+        "name": "bzip2",
+        "version": "1.0.8",
+        "qualifiers": {
+          "port_version": "6"
+        },
+        "subpath": null
+      },
+      "expected_failure": false,
+      "expected_failure_reason": null
+    },
+    {
+      "description": "vcpkg non-default git registry",
+      "test_group": "base",
+      "test_type": "parse",
+      "input": "pkg:vcpkg/ffmpeg@5.1.2?repository_url=https:%2F%2Fgithub.com%2Fazure-sdk%2Fvcpkg&port_version=3",
+      "expected_output": {
+        "type": "vcpkg",
+        "namespace": null,
+        "name": "ffmpeg",
+        "version": "5.1.2",
+        "qualifiers": {
+          "repository_url": "https://github.com/azure-sdk/vcpkg",
+          "port_version": "3"
+        },
+        "subpath": null
+      },
+      "expected_failure": false,
+      "expected_failure_reason": null
+    },
+    {
+      "description": "vcpkg filesystem registry or overlay port",
+      "test_group": "base",
+      "test_type": "parse",
+      "input": "pkg:vcpkg/llvm@15.0.7?repository_url=file:%2F%2F%2FC:%2Flocal-registry%2Fvcpkg&port_version=3",
+      "expected_output": {
+        "type": "vcpkg",
+        "namespace": null,
+        "name": "llvm",
+        "version": "15.0.7",
+        "qualifiers": {
+          "repository_url": "file:///C:/local-registry/vcpkg",
+          "port_version": "3"
+        },
+        "subpath": null
+      },
+      "expected_failure": false,
+      "expected_failure_reason": null
+    },
+    {
+      "description": "vcpkg with additional qualifiers",
+      "test_group": "base",
+      "test_type": "parse",
+      "input": "pkg:vcpkg/bzip2@1.0.8?port_version=6&repository_revision=84a143e4caf6b70db57f28d04c41df4a85c480fa&triplet=x64-linux",
+      "expected_output": {
+        "type": "vcpkg",
+        "namespace": null,
+        "name": "bzip2",
+        "version": "1.0.8",
+        "qualifiers": {
+          "port_version": "6",
+          "repository_revision": "84a143e4caf6b70db57f28d04c41df4a85c480fa",
+          "triplet": "x64-linux"
+        },
+        "subpath": null
+      },
+      "expected_failure": false,
+      "expected_failure_reason": null
+    },
+    {
+      "description": "vcpkg namespace is prohibited; boost-asio is a single port name, not a namespace and name",
+      "test_group": "base",
+      "test_type": "parse",
+      "input": "pkg:vcpkg/boost/asio@1.84.0",
+      "expected_output": null,
+      "expected_failure": true,
+      "expected_failure_reason": "Should fail to parse a vcpkg purl with a namespace because vcpkg namespaces are prohibited"
+    },
+    {
+      "description": "vcpkg purl without a name",
+      "test_group": "base",
+      "test_type": "parse",
+      "input": "pkg:vcpkg/@1.0.8",
+      "expected_output": null,
+      "expected_failure": true,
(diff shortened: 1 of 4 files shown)


JamieMagee:
Added some additional tests, specifically failing tests, as requested. Also, vcpkg logo can be found here: https://github.com/vcpkg/vcpkg.github.io/blob/main/assets/mark/mark.svg

types/vcpkg-definition.json
```diff
@@ -0,0 +1,52 @@
...
+      "key": "triplet",
+      "native_name": "triplet",
+      "requirement": "optional",
+      "description": "The vcpkg [triplet](https://learn.microsoft.com/vcpkg/concepts/triplets) describing the target build configuration (architecture, operating system, and CRT/library linkage), for example `x64-windows` or `arm64-osx`. It describes the build target and does not affect package identity."
+    }
+  ],
+  "note": "A purl may carry extra qualifiers that describe the context in which the package is used, such as build configuration or platform. Parsers must tolerate these and may ignore any they do not expect. The `features` qualifier, a comma-separated list of enabled vcpkg feature names, records which optional features were enabled at build time; it is informational and not currently normative.",
+  "examples": [
+    "pkg:vcpkg/bzip2@1.0.8?port_version=6",
+    "pkg:vcpkg/ffmpeg@5.1.2?repository_url=https://github.com/azure-sdk/vcpkg",
```
mjherzog: I realize that this may not be clearly explained in the current doc, but a URL in a **qualifiers** value needs to be  percent-encoded so this example should have:
 ?repository_url=https:%2F%2github.com%2azure-sdk%2vcpkg"

types/vcpkg-definition.json
```diff
@@ -0,0 +1,52 @@
...
+    "pkg:vcpkg/llvm@15.0.7?repository_url=file:///C:/local-registry/vcpkg",
```
mjherzog: In the note under repository you wrote: "The absolute URL of the vcpkg registry where the package is available. If omitted, the default registry `https://github.com/microsoft/vcpkg` is assumed. For filesystem registries or [overlay ports](https://learn.microsoft.com/vcpkg/concepts/overlay-ports), the URL uses the `file` scheme."

I assume that this example is for that use case. 

We encourage the use of standard **qualifiers keys** but some PURL **types** may merit a special **key**.
Following: https://learn.microsoft.com/en-us/vcpkg/concepts/registries, perhaps this case merits a 'vcpkg'-specific **qualifiers** **key** like 'filesystem_registry'.

tests/types/vcpkg-test.json
```diff
@@ -0,0 +1,251 @@
...
+        "subpath": null
+      },
+      "expected_failure": false,
+      "expected_failure_reason": null
+    },
+    {
+      "description": "vcpkg filesystem registry or overlay port",
+      "test_group": "base",
+      "test_type": "parse",
+      "input": "pkg:vcpkg/llvm@15.0.7?repository_url=file:///C:/local-registry/vcpkg&port_version=3",
```
mjherzog: See my related comment for vcpkg-definition.json.

mjherzog (changes requested):
@JamieMagee We are quite close to ready to register the 'vcpkg' PURL **type**, but I have some feedback.

tests/types/vcpkg-test.json
```diff
@@ -0,0 +1,251 @@
...
+        "subpath": null
+      },
+      "expected_failure": false,
+      "expected_failure_reason": null
+    },
+    {
+      "description": "vcpkg non-default git registry",
+      "test_group": "base",
+      "test_type": "parse",
+      "input": "pkg:vcpkg/ffmpeg@5.1.2?repository_url=https://github.com/azure-sdk/vcpkg&port_version=3",
```
johnmhoran: As @mjherzog has commented elsewhere, this input is also an example of a qualifiers value that contains unencoded slashes.  A slash not being used as a defined PURL separator must be percent-encoded, including in a qualifiers value.  The slash is not used as a Separator Character in a qualifiers value.  See

- [ECMA-427 5.6.6 Qualifiers](https://ecma-tc54.github.io/ECMA-427/#sec-purl-specification-rules-qualifiers) -- last bullet: "A value may contain any Unicode character and all characters shall be encoded as described in the Character encoding clause."
- [ECMA-472 5.4 Character encoding](https://ecma-tc54.github.io/ECMA-427/#sec-purl-specification-character-encoding) -- "The following characters shall not be percent-encoded: . . .  the Separator Characters when being used as PURL separators . . . ."

tests/types/vcpkg-test.json
johnmhoran: The qualifiers value's unencoded slashes here as well.

johnmhoran (commented):
Thank you for this @JamieMagee .  My only comments below concern unencoded slashes in qualifiers values -- not an uncommon sight.  ;-)  I realize we have many many test case examples with slashes used in qualifiers values without being percent encoded -- a vestige, I think, of the pre-ECMA-427 evolution of the spec.  BTW, great to see some test cases expected to fail -- an effective way of defining edge cases and bringing specificity to the broader language elsewehere.

types/vcpkg-definition.json
JamieMagee: Thank you. Done.

tests/types/vcpkg-test.json
JamieMagee: Thank you. Done.

tests/types/vcpkg-test.json
JamieMagee: Thank you. Done.

JamieMagee: The slashes here are now encoded. On the dedicated filesystem key, I've replied on your other thread.

types/vcpkg-definition.json
JamieMagee: Thanks. I'd prefer to keep `repository_url` for all registry kinds rather than add a separate `filesystem_registry` key. The `file://` scheme already marks a filesystem registry, `repository_url` is the standard qualifier that existing tooling reads, and a dedicated key would split the registry location across two keys depending on scheme. The `file://` example is meant to cover that case. Happy to revisit if you'd still prefer a dedicated key.

mjherzog: We can just use `repository_url` and adjust in the future if needed. And thank uyou for providing an example of this case.

mjherzog (approved):
@JamieMagee We are good to go,
@johnmhoran  Please take one final look and merge assuming no issues.

tests/types/vcpkg-test.json
```diff
@@ -0,0 +1,251 @@
...
+        "subpath": "6"
+      },
+      "expected_failure": false,
+      "expected_failure_reason": null
+    },
+    {
+      "description": "vcpkg purl with a features qualifier listing enabled features",
+      "test_group": "advanced",
+      "test_type": "parse",
+      "input": "pkg:vcpkg/curl@8.6.0?features=ssl,http2",
```
johnmhoran: @JamieMagee Apologies for not picking this up before -- I just noticed the comma in the qualifiers value, haven't seen that before, but as I read ECMA-427 5.4 and 5.6.6, that comma must be percent-encoded.  Please let me know if you read the language differently.

johnmhoran (changes requested):

tests/types/vcpkg-test.json
JamieMagee: Good catch, thanks. Done!

johnmhoran (approved):
Thank you @JamieMagee -- LGTM.  :+1:

PR #889, opened 2026-06-19
Add new PURL type: 'terraform'

ghouscht:
New PR because I accidentially closed #498 by deleting my fork - sorry about that.

Original PR text:

> See https://github.com/package-url/purl-spec/issues/369 for details.
>
> Please let me know what you think. This proposal is for now solely based on my ideas and I'm open for feedback. It might even make sense to define two distinct types terraform-provider and terraform-module for more clarity.


rafikk:
Hi, is there any movement on this? We're interested in producing SBOMs for terraform configuration and would love to have a standardized terraform package type.

types/terraform-definition.json
```diff
@@ -0,0 +1,30 @@
...
+    "note": "The HashiCorp registry is the default. Use the standard repository_url qualifier to point to another registry."
+  },
+  "namespace_definition": {
+    "requirement": "required",
+    "note": "The namespace is the provider or module namespace, which is usually the organization or vendor name.",
+    "case_sensitive": true
+  },
+  "name_definition": {
+    "case_sensitive": true,
+    "native_name": "Name is the name of the terraform provider or module."
```
mjherzog, line 19: Shouldn't this just be module name with provider defined in **namespace**?

types/terraform-definition.json
```diff
@@ -0,0 +1,30 @@
...
+  },
+  "version_definition": {
+    "case_sensitive": true,
+    "native_name": "Version is the package version and is required."
```
mjherzog, line 23: Version is always optional to support dependency or vulnerability use cases where you may want to express a range of versions with the 'vers' **qualifiers key** instead of a single version.

types/terraform-definition.json
```diff
@@ -0,0 +1,30 @@
...
+  "type_name": "Terraform providers and modules",
+  "description": "Terraform providers and modules",
+  "repository": {
+    "use_repository": true,
+    "default_repository_url": "https://registry.terraform.io",
+    "note": "The HashiCorp registry is the default. Use the standard repository_url qualifier to point to another registry."
+  },
+  "namespace_definition": {
+    "requirement": "required",
+    "note": "The namespace is the provider or module namespace, which is usually the organization or vendor name.",
```
mjherzog, line 14: Shouldn't this be provider name - usually an organization or vendor, but there are also individual persons.

types/terraform-definition.json
```diff
@@ -0,0 +1,30 @@
+{
+  "$schema": "https://packageurl.org/schemas/purl-type-definition.schema-1.0.json",
+  "$id": "https://packageurl.org/types/terraform-definition.json",
+  "type": "terraform",
+  "type_name": "Terraform providers and modules",
+  "description": "Terraform providers and modules",
```
mjherzog, line 6: Since **name** is always required this description is misleading. The package would appear to be just a terraform module.

types/terraform-definition.json
```diff
@@ -0,0 +1,30 @@
...
+    "default_repository_url": "https://registry.terraform.io",
+    "note": "The HashiCorp registry is the default. Use the standard repository_url qualifier to point to another registry."
+  },
+  "namespace_definition": {
+    "requirement": "required",
+    "note": "The namespace is the provider or module namespace, which is usually the organization or vendor name.",
+    "case_sensitive": true
+  },
+  "name_definition": {
+    "case_sensitive": true,
```
mjherzog, line 18: From browsing the registry module names seem to be all lowercase which is the canonical form for `"case_sensitive": false`. I could not find relevant documentation for this question at https://developer.hashicorp.com/terraform

types/terraform-definition.json
```diff
@@ -0,0 +1,30 @@
...
+    "native_name": "Name is the name of the terraform provider or module."
+  },
+  "version_definition": {
+    "case_sensitive": true,
+    "native_name": "Version is the package version and is required."
+  },
+  "examples": [
+    "pkg:terraform/hashicorp/aws@6.0.0",
+    "pkg:terraform/terraform-aws-modules/vpc@6.0.1",
+    "pkg:terraform/terraform-aws-modules/vpc@6.0.1?registry_url=https://registry.example.com"
```
mjherzog, line 28: `registry_url` must be encoded

mjherzog:
@ghouscht
- See my specific comments in terraform-definition.json
- You need to remove the change for PURL-TYPES.rst - this file was deprecated a year ago and is only retained now as a placeholder. (I tried to delete it from the PR for you, but I think that ended up as a change to delete the file which we do not want to do yet.)
- We normally prefer/expect some input from the relevant package ecosystem/community. Is there anyone from Hashicorp who has reviewed your proposed PURL **type**?
- We do not absolutely require a **type** level test file for a new PURL **type**, but it is good to have some test cases especially to show common use cases of incorrect PURL strings.

PR #890, opened 2026-06-25
Fix typo and incorrect regex in chrome-extension type definition

Phaeilo:
- Remove duplicate https:// in default_repository_url
- Tighten name regex from [a-z]{32} to [a-p]{32}; Chrome extension IDs are base16-encoded using letters a-p instead of hex digits


mjherzog:
@Phaeilo Thank you for finding the URL typo!
For the regex change - is there a documentation reference that we can include (for posterity)?

Phaeilo:
Did not find any good documentation, but maybe this file from the chromium codebase is sufficient? https://chromium.googlesource.com/chromium/src/+/refs/tags/128.0.6551.1/components/crx_file/id_util.cc

mjherzog:
@Phaeilo That is pretty good evidence
@mixmix Do you agree with this correction?

mjherzog (approved):
The explanation for the regex change is sufficient

johnmhoran (approved):
LGTM -- thanks @Phaeilo .  :+1:

Issue #157, opened 2022-03-17
OCI type: is version required?


MarkLodato:
The text currently says:

> The `version` is the [...] and is required to uniquely identify the artifact.

Does that mean that the field is REQUIRED, and the "to uniquely identify the artifact" is explaining why it's required? Or is it only required if you want to uniquely identify the artifact?

To clarify, I suggest either:

* The `version` is REQUIRED and is the [...]. This is needed to uniquely identify the artifact.
* The `version` is the [...]. It SHOULD be specified to uniquely identify the artifact.

If it is actually the first version, I strongly suggest making it optional. `oci` is the only type where `version` is required, and the [spec](https://github.com/package-url/purl-spec/blob/master/PURL-SPECIFICATION.rst) even says "version: the version of the package. Optional". In some cases it is necessary to describe a package with a floating label where the hash is not yet resolved.

itaysk:
the spec already says it's optional for all types:

> version: the version of the package. Optional.

johnmhoran:
Hi @MarkLodato .  I don't know what the language of the `oci` type definition was when you opened this issue, but right now, the definition [does not include the term "required"](https://github.com/package-url/purl-spec/blob/main/PURL-TYPES.rst#oci):

```
The version is the sha256:hex_encoded_lowercase_digest of the artifact and is used to uniquely identify the artifact.
```

Ah, looking at blame, I see the language was updated via [PR 432](https://github.com/package-url/purl-spec/pull/432) by @p-rog, merged on 2025-03-27.  If that PR fully addresses your concern, can you please close this issue?

MarkLodato:
Yes, PR 432 addressed the issue. Thank you.

PR #898, opened 2026-07-07
Register new PURL type: spack

### LINKED ISSUE #895: Register new PURL type: spack
[spack](https://spack.readthedocs.io/en/latest/) is a package management tool designed to support multiple versions and configurations of software on a wide variety of platforms and environments. It was designed for large supercomputing centers, where many users and application teams share common installations of software on clusters with exotic architectures, using libraries that do not have a standard ABI. spack is also the backing for the [Extreme-scale Scientific Software Stack](https://e4s.io/).

spack offers strong mechanisms for examining the supply chain for every build of each package. For example, the SHA256 hash value for upstream tarballs of each package are verified by the package maintainer as part of the standard package maintainence process. spack can also utilize pre-built binaries from known caches with cryptographic signatures to provide extended provenance while speeding up builds for large software stacks.

Like other package managers such as conan and vcpk, spack makes available a wide variety of software packages with a simple, unified installer across many operating systems, compilers, and runtimes. Each package installation is _uniquely_ identified by a SHA256/512 hash tracking configure and build information such as the package's version, compiler options, language versions, manually-applied patches, and more. Importantly, the same information for each of the package's dependencies is also encoded into the hash.

**Why is this new PURL type needed?**

Recently, Software Bill of Materials (SBOM) support was added to provide a uniform description of all the information encoded in the hash. With this, a PURL will allow security tools to easily map a unique installation (via its hash) to its SBOM data.

**What are the PURL component level definitions?**

`pkg:spack/<package>?sha=<SHA>` where `SHA` is the value returned by `spack find -l package`.

Example,

`pkg:spack/zlib?sha=u463uoum2lujbc2ug4fpbersp4jev2yl` maps to

```
spack spec zlib
[+]  zlib@1.3.2+optimize+pic+shared build_system=makefile platform=linux os=ubuntu22.04 target=skylake %c,cxx=gcc@11.4.0
[+]      ^compiler-wrapper@1.1.0 build_system=generic platform=linux os=ubuntu22.04 target=skylake 
[e]      ^gcc@11.4.0+binutils+bootstrap~graphite+libsanitizer~nvptx~piclibs~profiled~strip build_system=autotools build_type=RelWithDebInfo languages:='c,c++' platform=linux os=ubuntu22.04 target=x86_64 
[+]      ^gcc-runtime@11.4.0 build_system=generic platform=linux os=ubuntu22.04 target=skylake 
[e]      ^glibc@2.35 build_system=autotools platform=linux os=ubuntu22.04 target=x86_64 
[+]      ^gmake@4.4.1~guile build_system=generic platform=linux os=ubuntu22.04 target=skylake %c=gcc@11.4.0
```

**What input do you have from the relevant package ecosystem/community?**

The spack SBOM team are onboard.

**Are there any open questions or concerns about this new PURL type**

No.

hainest:
closes #895

Implementation for #895 

**TODO**
- [ ] Thumbnail image for [grid](https://www.packageurl.org/docs/purl/purl-types#registered-purl-types)
  - Make PR in https://github.com/package-url/www.packageurl.org
  - [image+text](https://spack.readthedocs.io/en/latest/_static/spack-logo-text.svg)
  - [just image](https://spack.io/assets/images/spack-logo.svg)
- [x] Add tests
- [x] Verify [VERS](https://www.packageurl.org/docs/vers/introduction) compatibility
  - Overall, spack versions are conformant. The only exception would be that spack versions allow using git refs. Those aren't comparable, so specifying a constraint in that case wouldn't be valid.


docs/types/definitions/spack-definition.md
```diff
@@ -0,0 +1,46 @@
...
+- **Note:** `The spack namespace as given by `spack find -N``
+
+## Name definition
+
+- **Requirement:** Required
+- **Note:** `The package's name`
+
+## Version definition
+
+- **Requirement:** Optional
```
becker33, line 34: I think the namespace and version should be required -- we obviously have the information if we have a Spack spec, and it's hard to claim we know what spec it is without it.

Then the purl can be given by `spack find --format "pkg:spack/{namespace}/{name}@{version}?sha={hash}"`

hainest: >I think the namespace and version should be required -- we obviously have the information if we have a Spack spec, and it's hard to claim we know what spec it is without it.

The SHA is required. Modulo perfectly reproducible builds, my perception is that this uniquely identifies a build. I don't have any objections to requiring a namespace and version. My only hesitation would be that some folks might not know that there is a default namespace, so using a PURL to identify a package would then have a possible sharp edge (maybe not too sharp). Certainly for _generating_ the PURLs inside spack, this is a non-issue.

matt-phylum: I don't think it's possible for the version or hash to be required in PURL. That would prohibit talking about the package in general without referring to a specific version. See #157

hainest: @matt-phylum Michael noted that a qualifier key _can_ be required, but should be rare (https://github.com/package-url/purl-spec/issues/895#issuecomment-5027866262). From 5.6.3 of the standard,

>The namespace is optional, unless required by the package's type definition.

so we can enforce a namespace. The standard does not make the same exception for a version- it is always stated as optional, so we should follow that.

I will revisit the standard and see if I can get any more clarity.

mjherzog: The only exception for 'required' **qualifiers keys** to date is for the Julia UUID and there was a long and deep discussion in the Julia community for this. It would be better to require a namespace instead if that is reasonable for spack.

hainest: The current structure is

1. namespace is **required**
2. version is **optional**
3. sha is **optional**

This makes it consistent with my reading of the standard and with other purls that I've seen in test cases for osv-scanner.

@mjherzog @matt-phylum

I didn't find a mechanism for running the tests in the Makefile or the CI workflows. Am I missing something?

If this is accepted, is the next step to make a PR in the [website](https://github.com/package-url/www.packageurl.org) repo to add the logo?

hainest: @mjherzog @matt-phylum If/when you have time, could you take a quick look? Thanks!

docs/types/definitions/spack-definition.md
```diff
@@ -0,0 +1,46 @@
...
+- **Note:** `The spack version using the standard spec syntax`
+
+## Qualifiers Definition
+
+| Key  | Requirement | Native name | Default Value | Description |
+|------|-------------|-------------|---------------|-------------|
+| sha | Optional |  |  | The hash of the installation as given by 'spack find -l' |
```
matt-phylum: This is unfortunate, but I guess if this is how Spack works it maybe makes more sense to have a special Spack sha qualifier like this than to use the usual PURL checksum qualifier. Using the standard qualifier would require Spack PURL implementations to convert back and forth between Spack's hash format and PURL's hex format, if that's even possible (is Spack truncating the hash?).

types/spack-definition.json
```diff
@@ -0,0 +1,35 @@
...
+    "requirement": "required",
+    "note": "The spack namespace as given by `spack find -N`"
+  },
+  "name_definition": {
+    "requirement": "required",
+    "note": "The package's name"
+  },
+  "version_definition": {
+    "requirement": "optional",
+    "note": "The spack version using the standard spec syntax"
```
matt-phylum: If this note is required, can it direct the reader to the relevant specifications?

docs/types/definitions/spack-definition.md
matt-phylum: I don't really know anything about Spack.

docs/types/definitions/spack-definition.md
hainest: Spack's hash encodes the specific configuration used to build the package. To avoid confusion between Spack's hash and a PURL checksum, I'll rename it to `daghash`.

types/spack-definition.json
hainest: Updated.

PR #906, opened 2026-07-10
Correct unencoded slashes in URLs in Examples

### LINKED ISSUE #905: Correct unencoded URLs in PURL **type** examples
The following PURL **type** definitions have unencoded slashes in Example URLs. This likely means that there may also be test cases for these PURL **types** where slash unencoding is not correctly handled.

**repository_url**
- bazel
- hex
- huggingface
- julia
- luarocks 
- maven
- mlflow
- npm
- oci

**download_url & vcs_url**
- generic

mjherzog:
Corrected 10 PURL **type** definition files to encode slashes in URLs in **qualifiers**

### DIFF
--- a/types/bazel-definition.json
+++ b/types/bazel-definition.json
@@ -38,7 +38,7 @@
   "examples": [
     "pkg:bazel/rules_java@7.8.0",
     "pkg:bazel/curl@8.8.0.bcr.1",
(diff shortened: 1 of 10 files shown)


johnmhoran (approved):
This looks great @mjherzog -- :+1:

PR #908, opened 2026-07-10
Add `vcs_url` to the vcpkg PURL type

JamieMagee:
vcpkg pins a port's exact recipe with the `vcs_url` git-tree URL, not `repository_revision`, which is a registry baseline. Adds `vcs_url` to the examples, note, and tests, and clarifies the `repository_revision` description.

Follow-up based on the discussion in https://github.com/microsoft/vcpkg-tool/pull/2049


mjherzog (approved):
Looks good to go.

johnmhoran (approved):
Thank you @JamieMagee -- LGTM.  :+1:

PR #915, opened 2026-07-14
Run remove_duplicate_tests.py on tests/types/ and tests/spec/ #832

johnmhoran:
Reference: https://github.com/package-url/purl-spec/issues/832


mjherzog (approved):
The changes looks good.

PR #926, opened 2026-07-16
833 document current purl spec ci workflows

johnmhoran:


etc/scripts/README.md
```diff
@@ -0,0 +1,18 @@
...
+| [etc/scripts/purl_type_definition.py](../../etc/scripts/purl_type_definition.py) | Generated by Makefile `gencode` command. |
+| [etc/scripts/purl_types_index.py](../../etc/scripts/purl_types_index.py) | Generated by Makefile `gencode` command. |
+| [etc/scripts/pyproject.toml](../../etc/scripts/pyproject.toml) | Configuration for Python ruff tool (linter and code formatter). Appears to use ruff for basic formatting (linting, isort etc.). |
+| [etc/scripts/remove_duplicate_tests.py](../../etc/scripts/remove_duplicate_tests.py) | Identifies duplicate test cases (i.e., 100% key-value match between test objects in a single `*-test.json` file, regardless of sort order). The script round-trips each file with `json.loads()` and `json.dumps()`, which might coincidentally correct some errors such as removing blank lines between adjacent items, but the process is not looking for, identifying and fixing syntactic or other errors in the test objects. We might want to incorporate a variation of this process into the `generate-index-and-docs.yml` workflow to check test file content. |
```
mjherzog: the process is not looking for, identifying _**and**_ fixing syntactic or other errors in the test objects.
should probably be:
the process is not looking for, identifying **_or_** fixing syntactic or other errors in the test objects.

etc/scripts/README.md
```diff
@@ -0,0 +1,18 @@
...
+| [etc/scripts/archive/git-split-file.sh](../../etc/scripts/archive/git-split-file.sh) | One-time use for initial mega-PR to create schema and test files. Archived. |
+| [etc/scripts/format_json.py](../../etc/scripts/format_json.py) | Basic formatting (e.g., indents, line endings). Called by the Makefile `formatjson` command (which is also called by the `format` command). Not used by any other purl-spec files. Targets `schemas/`, `types/` and `tests/`. |
+| [etc/scripts/generate_index_and_docs.py](../../etc/scripts/generate_index_and_docs.py) | Called by the Makefile `gendocs` command (which is also run by `.github/workflows/generate-index-and-docs.yml`). Not used by any other purl-spec files. Generates `/types/\*-definition.json`, `/docs/types/definitions/\*-definition.md` and `/purl-types-index.json` files. |
```
mjherzog: The markdown formatting seems off for two cases:
- `/types/\*-definition.json`, 
- `/docs/types/definitions/\*-definition.md`
The backslash formatting for the asterisk does not seem to be working when I preview in GH.
Perhaps it is affected by the preceding forward slash.
Some sites suggest using back-ticks of back slash instead but that might not work inside a string that is already in codestyle.
An alternative may be to replace the asterisk with <purl-type-name> or similar

etc/scripts/README.md
```diff
@@ -0,0 +1,18 @@
+# purl-spec workflow files
+
+This file (currently being drafted) summarizes the names, roles and functions of the purl-spec workflows, including but not limited to the process of submitting a PR proposing a new PURL `type`.  The majority of the files below are located in [etc/scripts/](https://github.com/package-url/purl-spec/blob/main/etc/scripts/).
```
mjherzog: We can probably remove (currently being drafted)

etc/scripts/README.md
johnmhoran: Good eye, done.

etc/scripts/README.md
johnmhoran: @mjherzog   Another good catch -- these are remnants of the .xlsx-to-.md conversion I ran when I first created this table.  My proposed update:

Generates `/types/\*`-definition.json`, `/docs/types/definitions/\*-definition.md` and `/purl-types-index.json` files.

should become

Generates `/types/*-definition.json`, `/docs/types/definitions/*-definition.md` and `/purl-types-index.json` files.

etc/scripts/README.md
johnmhoran: @mjherzog   Done -- PR has been updated for your review.

mjherzog (approved):
Looks good with the 3 resolved changes

PR #932, opened 2026-07-28
Update 001-PURL_type-software_without_registry.md

mjherzog:
Changed proposed PURL **type** name from 'sid' to 'software-id' with explanation added at the end of the document. Also updated format to:
- Replace bold with single quotes for the 'software-id' PURL **type**
- Consistently bold PURL **type** 
- Replaced some bold with italics

These are format standards from Ecma documentation which we are applying across PURL documenation

### DIFF
--- a/docs/decisions/001-PURL_type-software_without_registry.md
+++ b/docs/decisions/001-PURL_type-software_without_registry.md
@@ -1,34 +1,33 @@
 ---
 status: proposed
-date: 2026-03-24
+date: 2026-07-28
 ---
 
-# Proposal: Add new PURL type **sid** (**S**oftware **ID**entification)
+# Proposal: Add new PURL type 'software-id'
 
 ## Background
 
-The current list of PURL types covers a wide range of package ecosystems. 
+The current list of PURL **types** covers a wide range of package ecosystems. 
 However, there remains a significant gap in the ability to identify 
-**non-ecosystem** software components, particularly **commercial**, 
-**proprietary**, and **standalone open source** software that is not 
+*non-ecosystem* software components, particularly *commercial*, 
+*proprietary*, and *standalone open source* software that is not 
 distributed through a package manager or authoritatively available on a 
-repository with a supported PURL type, such as GitHub.
+repository with a supported PURL **type**, such as GitHub.
 
 ## Proposal
 
-Introduce a new PURL type: **sid**, an acronym for **Software IDentification**, 
-pronounced /sɪd/.
+Introduce a new PURL **type**: 'software-id'.
 
-This PURL type is intended to cover:
-- **Standalone open source projects** (e.g., the Linux Kernel)
-- **Commercial and proprietary software** (e.g., Acme Database Server)
-- **Internally developed applications** that do not belong to a recognized 
+This PURL **type** is intended to cover:
+- *Standalone open source projects* (e.g., the Linux Kernel)
+- *Commercial and proprietary software* (e.g., Acme Database Server)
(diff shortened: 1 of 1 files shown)


mjherzog:
@johnmhoran Please review for typos, consistent formatting etc.

docs/decisions/001-PURL_type-software_without_registry.md
```diff
@@ -1,34 +1,33 @@
...
+- *Commercial and proprietary software* (e.g., Acme Database Server)
+- *Internally developed applications* that do not belong to a recognized 
 package ecosystem
-- **Binary-only or installable software** where no manifest-based or 
+- *Binary-only or installable software* where no manifest-based or 
 registry-driven packaging exists
 
 ```
-pkg:sid/<authority>/<component>@<version>?arch=<arch>&edition=<edition>&target=<target>&locale=<locale>
+pkg:software-id/<authority>/<component>@<version>?arch=<arch>&edition=<edition>&target=<target>&locale=<locale>
```
johnmhoran, line 30: @mjherzog I was going to say that the `qualifiers` are meant to be sorted alphabetically by key, but after checking the [1st Edition ECMA-427 section 5](https://ecma-tc54.github.io/ECMA-427//#sec-purl-specification), I see no such requirement there.  I do see that instruction in the current how-to-build.md -- perhaps this becomes an optional normalization for the "recommended" test group rather than part of the "required" test group....

docs/decisions/001-PURL_type-software_without_registry.md
```diff
@@ -152,11 +153,17 @@ currently rely on CPE naming conventions.
...
+'software-id' PURLs align with CPE parts `a` (application) and `o` (operating 
 system). Hardware identification (CPE part `h`) is out of scope. Established 
 standards such as GS1 GMN and GTIN already serve that purpose.
 
 **NB**: This document replaces the core content of: https://github.com/package-url/purl-spec/issues/516.
-See the new Discussion item [Design for PURL type for software without a registry](https://github.com/package-url/purl-spec/discussions/841) for a recap of the commentary on issue 
-516 organized by topic. You can comment on each topic there using the "threaded"
-comments feature of Discussions.
\ No newline at end of file
+See the new Discussion item [Design for PURL type for software without a registry](https://github.com/package-url/purl-spec/discussions/841) for a recap of the commentary on issue 516 organized by topic. You can comment on each topic
```
johnmhoran: @mjherzog In the GH diff this line 161 looks far too long for the 78-character (right?) length limit.  Not sure if line length should be treated like indenting bullet carryover lines, i.e., no need to comment.  Please advise.  ;-)

johnmhoran (changes requested):
Looks good @mjherzog .  I have one minor observation and a comment/requested change.

docs/decisions/001-PURL_type-software_without_registry.md
mjherzog: adjusted

docs/decisions/001-PURL_type-software_without_registry.md
mjherzog: We may add this sorting to the 2nd edition standard but that is tbd. In this table the sort order for the "qualifiers keys"" seems to be the CPE field rather than the **key** so I think that I will leave this as-is. I had thought about swapping the columns to put the _'software-id' PURL equivalent_ first, but the CPE order seems to make sense here.

pombredanne (approved):
This looks fine. I am merging now. There are more issues though and I will comment separately as this is NOT from this PR.

PR #933, opened 2026-07-29
update must to shall for type definitions

mjherzog:
Changed "must" or "MUST" to shall in all PURL **type** definitions to conform to Ecma standard terminology.


docs/types/definitions/npm-definition.md
```diff
@@ -31,7 +31,7 @@ The structure of a PURL for this package type is:
 - **Requirement:** Required
 - **Case Sensitive:** Yes
 - **Native Label:** name
-- **Note:** `The package.json spec changed in 2015 to require that a new package 'must not have uppercase letters in the name', but old packages with mixed case names were "grandfathered in".`
+- **Note:** `The package.json spec changed in 2015 to require that a new package must not have uppercase letters in the name, but old packages with mixed case names were "grandfathered in".`
```
johnmhoran, line 34: Why is "must" not changed to "shall" here?  When it was in a quoted passage, as it had been, that result followed, but with the single quotes removed . . . ?

johnmhoran (changes requested):
Looks good except for one "must" that was not changed to "shall", not clear why.  See below.

docs/types/definitions/npm-definition.md
mjherzog: Not changed because this is a reference to package.json spec not PURL

johnmhoran (approved):
LGTM.

PR #938, opened 2026-08-05
Fix erroneous test schema filenames

johnmhoran:
- change 'purl-test.schema-1.0.json' to 'purl-test.schema-0.1.json'


mjherzog (approved):
Looks good

Issue #942, opened 2026-08-09
Clarification regarding qualifier key letter case


yan-foto:
(Please excuse if this is actually a duplicate issue. I couldn't find a relevant issue)

The [specification for qualifiers](https://ecma-tc54.github.io/ECMA-427/#sec-purl-specification-rules-qualifiers) states that:

> The key shall be composed only of lowercase ASCII letters and numbers, period '.', dash '-' and underscore '_'.

If I understand it correctly, "shall" in this context is a strict requirement, so that uppercase qualification keys make the respective PURL invalid. If this is the case, then some of the tests, e.g. for `gem` type are incorrect:

https://github.com/package-url/purl-spec/blob/a87369fabd705b7a81c57dd23da84bfe38a55b77/tests/types/gem-test.json#L4-L12

Violating a requirement (denoted by **shall**) must throw an error instead of silently fixing it, right?

mjherzog:
@yan-foto You are correct about the interpretation of "shall", but the purpose of 'recommended' test cases is to demonstrate how to normalize data to pass 'required' test cases (required for conformance). 
See also the quite new documentation at: https://package-url.github.io/www.packageurl.org/docs/purl/tests-folder for a longer explanation of the recent changes.

yan-foto:
Thanks @mjherzog for the clarification and the pointer to the new docs. This solves the issue for me. 

So basically for `recommended` , I should ignore validation errors and only check if my normalization is capable of producing a valid/canonical PURL?

P.S.
I also noted that I have  `advanced` instead of `recommended` since I'm on `v1.0.1` tag. Is there a tag that contains `v0.2` tests? I am vendoring the tests into my own project and I prefer to choose a specific tag instead of the `main` branch.

mjherzog:
@yan-foto - Yes to your question about 'recommended' test cases.
We agree about using a version instead of `main`. We will tag version 1.1.0 as soon as we complete review of the doc changes. No later than the end of this week.

yan-foto:
Sorry for circling back to the `required` vs `recommended`: I've found an example, where the test type is `required` *and* the qualifier keys are *not* in lowercase:

https://github.com/package-url/purl-spec/blob/f319c8ebe478a9c410d47aa56ed9a42720422562/tests/types/rpm-test.json#L13-L31

[...]

PR #946, opened 2026-08-11
Fix error in RPM test suite.

yan-foto:
One of the tests (with type `required`) has invalid qualifier keys (not in lowercase), but did not expect a failure. This was fixed by updating the test to a failing `required` test as discussed in issue #942 (original repository).


mjherzog (approved):
Looks good

PR #949, opened 2026-08-11
Fix swid example

### LINKED ISSUE #948: Possible error in `swid` example and tests
I'm testing my implementation against the test suites. I noted the first example in `swid` type definition:

https://github.com/package-url/purl-spec/blob/1133ae4c7392e88215d13310b990b53dd85da23a/types/swid-definition.json#L57-L61

If I'm not mistaken, the plus sign (`+`) in ` Enterprise+Server` must be percent-encoded, as it is [not permitted](https://ecma-tc54.github.io/ECMA-427/#sec-purl-specification-permitted-characters) in canonical PURL:

> A canonical PURL is composed of these permitted ASCII characters:
> 
> - the Alphanumeric Characters: A to Z, a to z, 0 to 9
> - the Punctuation Characters: .-_~ (period '.', dash '-', underscore '_' and tilde '~')
> - the Percent Character: % (percent sign '%')
> - the Separator Characters :/@?=&# (colon ':', slash '/', at sign '@', question mark '?', equal sign '=', ampersand '&' and hash sign '#')

It seems that this issue was [fixed](https://github.com/package-url/purl-spec/commit/cc66a1b43ff3dbfdf7616e7e12919504d9cbcf43) in other examples in the same file, but not for the first line.

As a result, the ([automatically generater](https://github.com/package-url/purl-spec/commit/18b23c6ea51f7301410bb7cbcfd64ac980535641)?) test suits also include invalid PURLs as input.

I'd be glad to submit a PR, if you confirm this issue.

yan-foto:
First example PURL for swid contains 'Enterprise+Server' as package name, which is not percent-encoded. This commit fixes #948.

### DIFF
--- a/types/swid-definition.json
+++ b/types/swid-definition.json
@@ -55,7 +55,7 @@
   ],
   "note": "Use of known qualifiers key/value pairs such as download_url can be used to specify where the package was retrieved from.",
   "examples": [
-    "pkg:swid/Acme/example.com/Enterprise+Server@1.0.0?tag_id=75b8c285-fa7b-485b-b199-4745e3004d0d",
+    "pkg:swid/Acme/example.com/Enterprise%2BServer@1.0.0?tag_id=75b8c285-fa7b-485b-b199-4745e3004d0d",
     "pkg:swid/Fedora@29?tag_id=org.fedoraproject.Fedora-29",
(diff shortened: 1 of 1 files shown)


mjherzog (approved):
Looks good - thank you

PR #954, opened 2026-08-12
Update Makefile and purl-test.py to use test schema 0.2 #950

### LINKED ISSUE #950: Update etc/scripts/purl_test.py for PURL test schema 0.2 changes
The merge of PR #949 failed the `generate-index-and-docs` step apparently because `etc/scripts/purl_test.py` was not updated for the terminology changes. 

From an initial look at the script it seems that we need to update the terminology in `class PurlTest(BaseModel)` for:
- `test_group`
- `test_type`
- `expected_failure_reason`

johnmhoran:
Reference: https://github.com/package-url/purl-spec/issues/950


mjherzog (approved):
Looks good.

Issue #969, opened 2026-08-16
Register new PURL type: smartsetup


dejo-ro:
We propose to introduce the `smartsetup` PURL type for the [TMS Smart Setup](https://doc.tmssoftware.com/smartsetup/) ecosystem / package manager. Smart Setup is the leading package manager in the Delphi/RAD Studio ecosystem, used for distributing both commercial and free packages from TMS itself and increasingly for any third-party creators of packages in the Delphi space.

**1. Why is this new PURL type needed?**
In short: regulatory push, mandatory SBOM generation and especially automated vulnerability monitoring. Next to Embarcadero's GetIt package manager (which does not currently employ standardized standard-based package URLs), Smart Setup is the major package manager for the Delphi ecosystem for both proprietary and open packages.
This is not only a design proposal. Smart Setup already generates a CycloneDX SBOM per installed product (`<productid>.cdx.json`), and these SBOMs already identify components with `pkg:smartsetup/…` purls. So this is about formalizing what is already happening in practice.

**2. What are the PURL component-level definitions?**
  - Form: pkg:smartsetup/<product-id>@<version>
  - type: smartsetup
  - namespace: prohibited — the vendor/owner segment is already part of the dotted product ID;
  splitting it would create a second naming convention Smart Setup doesn't use.
  - name: required — the full dotted product ID (native_name: "product id"); case-insensitive, 
  lowercased canonical form.
  - version: optional — the native version understood by Smart Setup (e.g. server-published version);
  preserved as-is, no imposed version grammar.
  - repository: use_repository: true, default_repository_url =
  https://www.smartsetup.org/ (the central Smart Setup registry that routes to the underlying
  sub-registries); note that other servers and Git-backed sources are also supported.
  - qualifiers: optional repository_url to point at a non-default server/source repository.
  - subpath: no Smart Setup-specific semantics proposed.
  - Examples: pkg:smartsetup/tms.biz.aurelius@5.26, pkg:smartsetup/tms.webcore@2.9.9.3,
  pkg:smartsetup/sglienke.spring4d, pkg:smartsetup/tms.biz.bcl.

**3. What input do you have from the relevant ecosystem/community?**
* The PURL standard is already being recognized by tooling (https://en.delphipraxis.net/topic/15582-ann-june-updates-for-pascal-analyzer-pascal-expert-and-pascal-browser/)
[...]

PR #970, opened 2026-08-16
Register smartsetup type

dejo-ro:
Cf. Issue https://github.com/package-url/purl-spec/issues/969

### DIFF
--- a/docs/types/definitions/smartsetup-definition.md
+++ b/docs/types/definitions/smartsetup-definition.md
@@ -1,0 +1,56 @@
+<!--  NOTE: Auto-generated from the JSON PURL type definition.
+Do not manually edit this file. Edit the JSON type definition instead. -->
+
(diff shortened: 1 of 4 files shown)


Maenken:
Hi @pombredanne @johnmhoran @mjherzog – friendly ping on this one.

Quick context in case it helps with prioritisation: this is not a name reservation. Smart Setup already emits pkg:smartsetup/... PURLs in the CycloneDX SBOMs it generates, and our own SBOM generator for RAD Studio (Maenken Systems) uses the same type. So this PR documents a scheme that is already in production use.

The main driver on our side is CRA compliance (Regulation (EU) 2024/2847, Annex I Part II) and BSI TR-03183-2, which expects a PURL per component where one exists.

Happy to adjust the definition to whatever the 2nd-edition schema work requires, if that is what is currently holding up type registrations. Thanks for your time!

PR #971, opened 2026-08-17
purl-spec: subpath-segments amount 1 or more

### LINKED ISSUE #963: purl-spec: clarify subpath-segments amount
based on https://github.com/package-url/purl-spec/pull/578/changes#r3775976167

there is a mis-alignment in the current purl standard: 
https://github.com/package-url/purl-spec/blob/9ca95bdd22fecb937bc4b43ad07d40b07f3da771/docs/specification/standard/specification.md?plain=1#L238-L240

So it brings together "when not empty" and "zero or more segments".
This is kind of confusing, the one thing talks about non-empty while the other allowed zero items...

in other parts of the PURL we allow one or many items - like in namespace.
https://github.com/package-url/purl-spec/blob/9ca95bdd22fecb937bc4b43ad07d40b07f3da771/docs/specification/standard/specification.md?plain=1#L185-L186

Lets streamline this **subpath** with other similar phrases we already use.

so proposed patch
```patch
- The **subpath** contains zero or more segments, separated by slash '/'
+ If present, the **subpath** may contain one or more segments, separated by
+ a single unencoded slash '/' character.
```

jkowalleck:
- closes #963

### DIFF
--- a/docs/specification/standard/specification.md
+++ b/docs/specification/standard/specification.md
@@ -237,8 +237,8 @@ The rules for each component are:
 ### Subpath
 - The **subpath** string is prefixed by a '#' separator when not empty.
 - The '#' is not part of the **subpath**.
-- If present, the **subpath** may contain one or more segments, separated by
-  a single unencoded slash '/' character.
+- If present, the **subpath** may contain one or more segments, each
+  separated by a single unencoded slash '/' character.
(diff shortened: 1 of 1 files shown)


johnmhoran (approved):
Good catch @jkowalleck -- thank you.  :+1:

PR #974, opened 2026-08-19
spec: clarify types may only add restrictions on components

### LINKED ISSUE #973: spec: clarity on component's type-driven restrictions
Based on https://github.com/package-url/purl-spec/pull/578#discussion_r3779749040,
we should revisit the spec a where it comes component restrictions due to type definitions.

currently it reads:
"unless the package's **type** definition provides otherwise."

this "provides otherwise" could mean, that a type could narrow an existing rule, or even widen it.
If a type could widen a rule, the rule would have no effect anymore.

It shall be clear about that this means narrowing a rule only - and never widen a rule.

# proposed change 
- as of https://github.com/package-url/purl-spec/pull/578#discussion_r3779749040
  ```patch
  - unless the package's **type** definition provides otherwise
  + unless the package's **type** definition prohibits specific characters
  ```
- cleaner more directional (as some AI claimed after researching the topic)
  ```patch
  - unless the package's **type** definition provides otherwise
  + unless the package's **type** definition further restricts the allowed characters
  ```

jkowalleck:
- fixes https://github.com/package-url/purl-spec/issues/973

### DIFF
--- a/docs/specification/standard/specification.md
+++ b/docs/specification/standard/specification.md
@@ -191,7 +191,7 @@ The rules for each component are:
     - shall not contain any slash '/' characters
     - shall not be empty
     - may contain any Unicode character other than '/' unless the package's
-      **type** definition restricts the allowed characters
+      **type** definition further restricts the allowed characters
 - A URL host or Authority shall not be used as a **namespace**. Use instead a
   **repository_url** qualifier. Note however, that for some types, the
   **namespace** may look like a host.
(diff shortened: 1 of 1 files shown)


johnmhoran (approved):
@jkowalleck -- In light of your initial instincts and @mjherzog 's comments, I'm convinced there's no ambiguity.  Thank you, Jan.  :+1:

mjherzog (commented):
This is an improvement

mjherzog (approved):
This is an improvement

PR #976, opened 2026-08-21
Update $id in purl-type-definition.schema-1.1.json

### LINKED ISSUE #956: Correct $id pattern for PURL type definition files
The current $id identifier is defined in `purl-type-definition.schema-1.0.json` as:
```
    "$id": {
      "title": "PURL type definition id",
      "description": "The unique identifier URI for this PURL type definition.",
      "type": "string",
      "pattern": "^https:\\/\\/packageurl\\.org/types/[a-z0-9-]+-definition\\.json$"
    },
```
This results in the following pattern for the PURL **type** definition files:
-  "$id": "https://packageurl.org/types/alpm-definition.json"
-  "$id": "https://packageurl.org/types/apk-definition.json"
- "$id": "https://packageurl.org/types/bazel-definition.json"
- and so on

Theses are invalid URLs because there is no `https://packageurl.org/types/` folder.
This is apparently not an operational problem because this is the pattern for all PURL **type** definition files and we have had no bug reports, but it seems to be an error that we should correct.

We could add a folder at www.packageurl.org, but it would need to be: https://packageurl.org/purl/types/.
And we would need an automated way to keep these files in synch with additions or updates of the primary PURL **type** definition files located at: https://github.com/package-url/purl-spec/tree/main/types/ .

### LINKED ISSUE #987: Update $id in existing PURL type definition files.
With #976 we changed "pattern" for "$id" in from:
"https:\/\/packageurl\.org/types/[a-z0-9-]+-definition\.json$" in `purl-type-definition.schema-1.0.json` 
to:
"https:\/\/packageurl\.org/purl-types/[a-z0-9-]+-definition\.json$" in `purl-type-definition.schema-1.`.json`.

After we finalize `purl-type-definition.schema-1.1.json` for ECMA-427 2nd Edition we will need to update the `$id` field in all existing PURL type definition files to the new value - see 
- #992

mjherzog:
Changed "pattern" for "$id" from:
 "https:\\/\\/packageurl\\.org/types/[a-z0-9-]+-definition\\.json$"
to:
 "https:\\/\\/packageurl\\.org/purl-types/[a-z0-9-]+-definition\\.json$"

to avoid conflict with VERS **type** definitions.

### DIFF
--- a/schemas/purl-type-definition.schema-1.1.json
+++ b/schemas/purl-type-definition.schema-1.1.json
@@ -84,7 +84,7 @@
       "title": "PURL type definition id",
       "description": "The unique identifier URI for this PURL type definition.",
       "type": "string",
-      "pattern": "^https:\\/\\/packageurl\\.org/types/[a-z0-9-]+-definition\\.json$"
+      "pattern": "^https:\\/\\/packageurl\\.org/purl-types/[a-z0-9-]+-definition\\.json$"
     },
     "type": {
       "title": "PURL type",


johnmhoran (approved):
Thanks @mjherzog  -- the path is now consistent with the corresponding file's path in vers-spec.  :+1:

PR #977, opened 2026-08-24
Add definition for recommended_requirement and add it to qualifers

### LINKED ISSUE #710: Schema change / purl-type-definition: Add recommended qualifier keys
This issue proposes updating the **qualifiers** subschema of the PURL **type** definition schema to add "recommended' **keys** for a PURL type. The current v1.0 schema provides the choice of 'optional' or 'required' for a **key* . 

The purpose of the change is to document the **key** or **keys** that "should" be included for a particular PURL **type**.  The distinction here between 'required' and 'recommended' is that a **qualifier key** should only be 'required' when it is essential for the identification of a package across all use cases - e.g. for Linux distro PURL **types** you will normally want a **qualifier key** like 'arch' for an identity or location use case, but not for a vulnerability mgt. use case where you want to find vulnerabilities across architectures.

This change will support test cases for 'recommended' ('should", not "shall') behavior that provide a warning-style test message for a PURL that does not include the 'recommended' **keys** for its **type**

**NB**: Rewrite on August 24, 2026 to make it distinct from #690

mjherzog:
Add recommended to requirements for qualifiers
- Added new `recommended_requirement` to definitions
- Added to requirement for **qualifiers** only

### DIFF
--- a/schemas/purl-type-definition.schema-1.1.json
+++ b/schemas/purl-type-definition.schema-1.1.json
@@ -12,6 +12,12 @@
       "type": "string",
       "const": "optional"
     },
+    "recommended_requirement": {
+      "title": "Component recommended_requirement requirement",
+      "description": "States that this PURL component is recommended for a PURL type.",
+      "type": "string",
+      "const": "optional"
+    },
     "required_requirement": {
       "title": "Component required requirement",
       "description": "States that this PURL component is required for a PURL type.",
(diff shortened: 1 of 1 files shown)


johnmhoran (approved):
Interesting parallel use of the `test_group` values.  LGTM.

mjherzog:
> Interesting parallel use of the `test_group` values. LGTM.

Yes - that is pretty much the point

pombredanne (approved):
All good.

PR #983, opened 2026-08-25
Update standard doc file names to match ECMA-427

mjherzog:
Updated file names in `docs/specification/standard/` to match ECMA-427 structure
None of these files are currently published to the website


johnmhoran (approved):
Good idea @mjherzog LGTM.

PR #989, opened 2026-08-27
Create Clause 5.7 for PURL type definitions

### LINKED ISSUE #286: Clarify whether the `type` should be required to be a "known" type or whether it can be an arbitrary field
I've been playing around with [GitHub's Dependency Submission API](https://docs.github.com/en/rest/dependency-graph/dependency-submission?apiVersion=2022-11-28#create-a-snapshot-of-dependencies-for-a-repository), which consumes pURL(s).

However, as noted [here](https://github.com/renovatebot/renovate/discussions/24823#discussioncomment-8132131) and in https://github.com/anchore/syft/issues/1622, it appears that GitHub's support for pURL `type` fields has some validation, looking for "known" pURLs, rather than allowing arbitrary `type`s.

For instance, if we use [packageurl-go](github.com/package-url/packageurl-go), the pURL `pkg:mix/req@~%3E%200.3` parses correctly:

```go
package main

import (
	"fmt"
	"log"

	"github.com/package-url/packageurl-go"
)

func main() {
	s := "pkg:mix/req@~%3E%200.3"
	fmt.Printf("s: %v\n", s)
	p, err := packageurl.FromString(s)
	if err != nil {
		log.Fatal(err)
	}
	fmt.Printf("p.Name: %v\n", p.Name)
	fmt.Printf("p.Version: %v\n", p.Version)
}
```

Clarification on this issue would allow raising this to GitHub as a defect, if it is one.

mjherzog:
Added `docs/specification/standard/Clause-5-Package-URL-Specification.md`
Updated `docs/specification/README.md` to clarify that the files in `docs/specification/standard/` are the drafts for ECMA-427 2nd Edition

### DIFF
--- a/docs/specification/README.md
+++ b/docs/specification/README.md
@@ -12,12 +12,16 @@ There are two levels of PURL specification documentation:
   of the PURL Specification in other software or databases.
 
 ## ECMA-427 documentation
-ECMA-427 is the official documentation for the 1st edition of the Package-URL
-(PURL) Specification Standard. The source for this documentation is located
+ECMA-427 is the official documentation for the the Package-URL (PURL) Specification Standard. The source for this documentation is located
 at: https://github.com/Ecma-TC54/ECMA-427/blob/main/spec.html. The text from
 this HTML file is processed with several Ecma tools to produce a PDF file that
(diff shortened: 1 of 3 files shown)


mjherzog:
The primary change to add section 5.7 to ECMA-427 Clause 5 was approved at the TC54/TG2 meeting today.

PR #991, opened 2026-08-27
Update Clause-5-Package-URL-Specification.md

### LINKED ISSUE #435: Add wording that a PURL type SHOULD be locatable
We should identify any PURL **type** that CANNOT be located  A PURL of a given **type** SHOULD be locatable. Unlocatable PURLs are mostly harmless and should be rare. 

We could add this wording at the top:
- https://github.com/package-url/purl-spec/blob/main/PURL-TYPES.rst
- and also in the spec we should define for each type if it is locatable or not.

mjherzog:
- Replaced list item: "A PURL is a valid URL because it is a locator even though it has no Authority URL component: each type has a default repository location when defined" with a new section at the bottom of 5.1
- Also corrected commas in Permitted Character item list

The original statement is not correct because there is not always a default repository location for a PURL **type**.

### DIFF
--- a/docs/specification/standard/Clause-5-Package-URL-Specification.md
+++ b/docs/specification/standard/Clause-5-Package-URL-Specification.md
@@ -67,7 +67,7 @@ to a **type**.
   scheme. They may be used to reference URLs in separate attributes outside of
   a PURL or in a PURL qualifier.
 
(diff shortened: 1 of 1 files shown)


mjherzog:
PR approved at the TG2 meeting today

PR #993, opened 2026-08-28
Update normaliz* to normalis* #990

### LINKED ISSUE #990: Update normalization_rules to normalisation_rules?
There is an inconsistency in ECMA-427 1st Edition for the term normalisation/normalization because we use "normalisation" and "normalise" in the text (e.g. Clause 2 - Conformance) but we have `normalization_rules` as a property name in the PURL **type** definition schema for the **name**, **version** and **subpath** components.

Should we update the PURL **type** properties to 'normalisation_rules` for consistency?

johnmhoran:
Reference: https://github.com/package-url/purl-spec/issues/990


mjherzog (approved):
The changes are comprehensive and look good

PR #998, opened 2026-09-04
Upgrade datamodel-code-generator from 0.31.2 to 0.64.0 #997

### LINKED ISSUE #997: Address datamodel-code-generator vulnerabilities
We have nine vulnerabilities reported for `datamodel-code-generator` at:
https://github.com/package-url/purl-spec/security/dependabot

The recommendation is to upgrade to a version `~> 0.61.0` based on the dependabot information.
We should cross-check the dependabot report against info in VCIO.

johnmhoran:
Reference: https://github.com/package-url/purl-spec/issues/997

### DIFF
--- a/etc/scripts/purl_test.py
+++ b/etc/scripts/purl_test.py
@@ -20,23 +20,23 @@
 #
 # Visit https://github.com/package-url/purl-spec and https://packageurl.org for support
 
-
 from __future__ import annotations
-from typing import Any, Literal, Optional
+
 from pydantic import BaseModel, ConfigDict, Field
+from typing import Any, Literal
 
 
 class PackageUrlTestDefinition(BaseModel):
     model_config = ConfigDict(
         extra="forbid",
     )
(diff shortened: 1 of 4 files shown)


johnmhoran:
@mjherzog As discussed, I've upgraded `datamodel-code-generator` from `0.31.2` to `0.64.0` (in `requirements.txt`).  I also ran `make conf`, activated the venv, then ran `make check`, `make gendocs` and `make gencode`.  All succeeded, and `make gencode` updated 3 .py files:

etc/scripts/purl_test.py
etc/scripts/purl_type_definition.py
etc/scripts/purl_types_index.py 

Some of the changes are substantial, but I'm not able to evaluate whether any is problematic.  Let's discuss tomorrow when you have time -- perhaps someone more experienced with Python than I am could take a look before we proceed further.

mjherzog:
I think that we should proceed since our key utilities work.

PR #1005, opened 2026-09-14
Update Clause-6-Package-URL-Type-Definition-Schema.md

mjherzog:
Applied two major changes from purl-types-definition.schema-1.1.json`
- Added `registered_values` for the **namespace** component (an array with 3 properties)
- Added `recommended_requirement` as "Component recommended requirement" for the **qualifiers** component.  `recommended_requirement` is defined in the JSON schema under schema-level `definitions` but for Clause 6 the PURL schema-level definitions are defined at the PURL component level.


johnmhoran (approved):
LGTM.

PR #1010, opened 2026-09-17
use angle brackets in ABNF prose for rule names

jkowalleck:
https://en.wikipedia.org/wiki/Augmented_Backus%E2%80%93Naur_form
> Angle brackets (`<`, `>`) are not required around rule names (as they are in BNF). However, they may be used to delimit a rule name when used in prose to discern a rule name.

So i aplied this - should make it more clear, when writing/mentioning an ABNF rule - while the purl components are stylized in bold.


mjherzog (approved):
The changes are an improvement for readability.

PR #1012, opened 2026-09-17
Update Clause 5 to remove "canonical"

mjherzog:
- Replaced "canonical" with "valid" (1 instances)
- Replaced "The canonical form is lowercase." with "The form is lowercase." (1 instance)
- Replaced "should be stripped in the canonical form." with "should be stripped." 2 instances)
These changes align with similar changes in Clause 6 and the PURL type definition schema


johnmhoran (approved):
LGTM @mjherzog

PR #1014, opened 2026-09-19
Remove backticks from component field

jaimergp:
Comes from observation made at https://github.com/package-url/purl-spec/pull/1013/files#r4053091972.

This looks like an oversight, since every other "note" field in the schema is not backticked and can use Markdown.


mjherzog:
@jaimergp Thank you for reporting this problem but the solution needs to be in the utility that generates the markdown documentation files from the original definition files in `purl-spec/types`.
See:
- #1015

jaimergp:
Yes, @mjherzog, I edited that file (see the end of the diff) and then regenerated the documentation. I also added now a check for Python versions because the script requires 3.10 for `.write_text(..., newline="\n")` and some systems still ship an earlier version (e.g. macOS 26.6 carries Pytrhon 3.9).

mjherzog:
@jaimergp My mistake to not read to the bottom of the file list. This looks good.

mjherzog (approved):
Thank you for addressing the core issue in the generate_index_and_docs script

johnmhoran (approved):
This looks great @jaimergp , including the 3.10 check.  Thank you.  :+1:

